PR #13, opened 2026-04-27
refactor: rename OP_PHASE_* bytecode opcodes to OP_ARRAY_*

bachase:
## Summary

The Subspace category in `docs/opcodes.json` was conflating two distinct runtime semantics:

1. **k-mutators** — `OP_EXPAND`, `OP_EXPAND_T`, `OP_EXPAND_T_DAG`, `OP_EXPAND_ROT`. These genuinely change the active dimension $k$.
2. **Non-Clifford at fixed k** — `OP_PHASE_T`, `OP_PHASE_T_DAG`, `OP_PHASE_ROT`. These touch one virtual axis of the state vector, $k$ unchanged. The runtime work shape is identical to `OP_ARRAY_S` (`exec_phase_t` and `exec_array_s` apply `diag(1, *)` via the same waterfall) — only "non-Clifford" sets them apart.

Lumping (2) under "Subspace" reads the category along the wrong axis (Clifford vs non-Clifford) instead of along runtime cost shape (frame-only / fixed-k state vector / k-mutating). This PR splits them: the renamed and recategorized ops join the Array block; Subspace now contains only the genuine k-mutators.

It also resolves a pre-existing inconsistency: `OP_ARRAY_U2` / `OP_ARRAY_U4` were named `ARRAY` but filed under Subspace because they fuse non-Clifford rotations. Under the cleaner split they land where their name says they should.

### Renames

| Before | After |
|---|---|
| `OP_PHASE_T` | `OP_ARRAY_T` |
| `OP_PHASE_T_DAG` | `OP_ARRAY_T_DAG` |
| `OP_PHASE_ROT` | `OP_ARRAY_ROT` |
| `make_phase_t` / `make_phase_t_dag` / `make_phase_rot` | `make_array_t` / `make_array_t_dag` / `make_array_rot` |
| `exec_phase_t` / `exec_phase_t_dag` / `exec_phase_rot` | `exec_array_t` / `exec_array_t_dag` / `exec_array_rot` |
| Dispatch labels `L_OP_PHASE_*` | `L_OP_ARRAY_*` |

### Recategorization in `docs/opcodes.json`

Subspace → Array: `OP_ARRAY_T`, `OP_ARRAY_T_DAG`, `OP_ARRAY_ROT`, `OP_ARRAY_U2`, `OP_ARRAY_U4`.
Subspace stays: `OP_EXPAND`, `OP_EXPAND_T`, `OP_EXPAND_T_DAG`, `OP_EXPAND_ROT`.

### Intentionally unchanged

The HIR layer keeps `HeisenbergOp::make_phase_rotation` and `OpType::PHASE_ROTATION`. At HIR level (before Pauli localization) this is a phase rotation on a Pauli product — a distinct concept from the post-localization runtime opcode, and the existing name is correct there.

### Files changed

- `src/clifft/backend/backend.{h,cc}` — `Opcode` enum (entries reordered to group all `ARRAY_*` together), factory declarations and definitions.
- `src/clifft/svm/svm_kernels.inl` — kernel function names, computed-goto labels, dispatch table (reordered to keep designated-initializer indices monotonic — see note below), generic `case` switch.
- `src/clifft/optimizer/{expand_t,single_axis_fusion,tile_axis_fusion}_pass.{cc,h}` — opcode references in pattern matchers.
- `src/clifft/util/introspection.{cc,h}` — opcode-name strings, "is array op" predicate.
- `src/python/bindings.cc` — `Opcode` enum bindings, pass docstrings.
- `docs/opcodes.json` — keys, categories, cross-references in `detail` text.
- `docs/macros.py` — `ExpandTPass` / `ExpandRotPass` / `SingleAxisFusionPass` `detail` strings.
- `playground/src/components/GuidedTour.tsx` — VM Bytecode tour bullet (`OP_PHASE_T` → `OP_ARRAY_T`).
- Tests (`tests/*.cc`, `tests/python/test_introspection.py`) — opcode references.

### Note on the dispatch table

Reordering the enum to group `ARRAY_*` together broke GCC's monotonic-index requirement for the C99 designated-initializer dispatch table in `svm_kernels.inl`. Reordering the table block to match the new enum order resolves it (the table still uses `static_cast<uint8_t>(Opcode::...)` keys, so it's robust to any future enum tweaks as long as the source-order matches the enum-value order).

## Test plan

- [x] `uv run pytest tests/python` — 595 passed.
- [x] Standalone CMake build (`cmake -S . -B build-tests -G "Unix Makefiles"`) + `./build-tests/tests/clifft_tests` — 650 cases, 113,390 assertions, all pass.
- [x] `uv run mkdocs build` — clean (Array category in `docs/reference/instructions.md` now contains the renamed ops as expected).
- [x] No stray `OP_PHASE_*` / `make_phase_t*` / `make_phase_rot(` / `exec_phase_*` / `L_OP_PHASE_*` references remain (`grep -rn` clean across `src/`, `tests/`, `docs/`, `playground/src/`).
- [ ] Spot-check the Playground guided tour and `instructions.md` rendering after this lands.

## Sequencing

PR #12 (docs–paper alignment) will rebase on top of this so its docs already reference the new names.

### DIFF
--- a/docs/macros.py
+++ b/docs/macros.py
@@ -141,7 +141,7 @@ def define_env(env: Any) -> None:
             "python_name": "ExpandTPass",
             "summary": "Fuses EXPAND + T-phase into a single array pass.",
             "detail": (
-                "Fuses contiguous OP_EXPAND + OP_PHASE_T (or T_DAG) pairs into "
+                "Fuses contiguous OP_EXPAND + OP_ARRAY_T (or T_DAG) pairs into "
                 "single OP_EXPAND_T (or OP_EXPAND_T_DAG) instructions. The separate "
                 "instructions cause two array passes; the fused instruction performs "
                 "both in one loop: arr[i+half] = arr[i] * exp(+/-i*pi/4)."
@@ -154,7 +154,7 @@ def define_env(env: Any) -> None:
             "python_name": "ExpandRotPass",
             "summary": "Fuses EXPAND + continuous rotation into a single array pass.",
             "detail": (
-                "Fuses contiguous OP_EXPAND + OP_PHASE_ROT pairs into single "
+                "Fuses contiguous OP_EXPAND + OP_ARRAY_ROT pairs into single "
                 "OP_EXPAND_ROT instructions, eliminating the two-pass penalty of "
                 "separate expand and phase-rotate operations."
             ),
@@ -195,8 +195,8 @@ def define_env(env: Any) -> None:
             "python_name": "SingleAxisFusionPass",
             "summary": "Fuses single-axis operation chains into precomputed 2x2 unitaries.",
             "detail": (
-                "Fuses consecutive single-axis operations (ARRAY_H, ARRAY_S, PHASE_T, "
-                "PHASE_ROT, etc.) on the same virtual axis into a single OP_ARRAY_U2 "
+                "Fuses consecutive single-axis operations (ARRAY_H, ARRAY_S, ARRAY_T, "
+                "ARRAY_ROT, etc.) on the same virtual axis into a single OP_ARRAY_U2 "
                 "instruction. Pre-computes 2x2 unitary matrices for all 4 possible "
                 "incoming Pauli frame states (I, X, Z, Y). A run is fused if it "
                 "contains at least 3 array-touching operations, or at least 2 when "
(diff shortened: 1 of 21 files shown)


bachase (commented):
👍

PR #19, opened 2026-04-29
chore(util): drop redundant comments in util headers

bachase:
## Summary

First subfolder in a methodical comment-cleanup sweep across `src/clifft/`. Goal: remove silly/uninformative comments while keeping anything load-bearing (non-obvious *why*, design constraints, paper-relevant terminology).

What changed in `util/`:

- **bitmask.h** — dropped \`///\` comments that just restated the function name (\`is_zero\`, \`popcount\`, \`bit_get/set/xor/swap\`). Kept the file-level header (load-bearing why: compile-time-constant extents enabling auto-vectorization without intrinsics), the \`lowest_bit\` sentinel note (trimmed to \"Returns N as a sentinel when no bit is set\"), the \`clear_lowest_bit\` \`x &= x - 1\` analogy, and the implicit-ctor justification for the \`NOLINT\`.
- **config.h** — dropped the redundant second line of the file header and the restatement \"Maximum number of qubits supported at compile-time\". Kept the build-flag instructions, the BitMask/Pauli-frame consequence, and the \"Instruction struct stays 32 bytes regardless\" why.
- **introspection.h** — dropped pure-restatement comments above \`op_type_to_str\`, \`format_hir_op\`, \`opcode_to_str\`, and \`format_instruction\`. Kept the file header (\"used by both Python and Wasm bindings\" — explains why it lives in util) and the \`format_pauli_mask\` example.

No content changes to \`introspection.cc\`, \`constants.h\`, \`version.h.in\`, or \`tests/test_bitmask.cc\` — those were already in good shape (or the comments were genuinely useful annotations of test-case intent).

Net: 3 files, +4/-16.

## Test plan

- [x] \`clifft_tests \"BitMask:*\"\` — all 18 cases / 57 assertions pass
- [x] Library + tests compile cleanly

### DIFF
--- a/src/clifft/util/bitmask.h
+++ b/src/clifft/util/bitmask.h
@@ -89,7 +89,6 @@ struct BitMask {
 
     // -- Queries --
 
-    /// True if all bits are zero.
     [[nodiscard]] constexpr bool is_zero() const {
         for (size_t i = 0; i < kWords; ++i) {
             if (w[i] != 0)
@@ -98,7 +97,6 @@ struct BitMask {
         return true;
     }
 
-    /// Population count across all words.
     [[nodiscard]] constexpr int popcount() const {
         int c = 0;
         for (size_t i = 0; i < kWords; ++i)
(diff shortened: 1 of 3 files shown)


bachase (commented):
👍

PR #21, opened 2026-04-29
docs: trim restatement comments and stale notes in circuit/

bachase:
## Summary

Second component in the methodical comment-cleanup pass (after #19 for `util/`). Same philosophy: drop pure restatements of well-named identifiers, keep WHYs.

- **Stale fix in `parser.h`**: header listed `R q -> M q; CX rec[-1] q` (and the RX equivalent) under "Parser transformations" — but R/RX are kept as first-class gates (matches `circuit.h:11`, `parser.cc`, and `test_parser.cc:281` "R is kept as a first-class gate, not decomposed"). Dropped.
- **Clarified in `circuit.h`**: the `num_measurements` field said "M, MX, MY, MPP, MR, MRX" but is also incremented for MRY and MPAD. Replaced with the principle: visible measurements are entries in the measurement record; resets without measurement (R, RX, RY) don't increment.
- **Dropped "AG pivot" framing** in `test_stim_contract.cc` — Aaronson-Gottesman is no longer the attribution used in the paper. (Other AG references still live in `frontend/`, `backend/`, `tests/test_hir.cc` — those will be addressed in their own subfolder passes.)
- **`target.h` / `parser.cc`**: drop `///` comments that just paraphrase the function name (e.g. `is_rec()`, `is_inverted()`, `inverted()`, `is_gate_char()`, `// Strip comments.`). Trim verbose ones (`value()`, `pauli()`, `pauli_char()`).

`gate_data.h` and `test_parser.cc` left unchanged — their comments are pulling weight (Stim-name disambiguation; concrete test expectations).

Net: 5 files, +11/-30.

## Test plan

- [x] `cmake --build build-tests` succeeds
- [x] `./build-tests/tests/clifft_tests "[parser]","[gate_data]","[stim][contract]"` — 733 assertions / 149 cases pass
- [x] `./build-tests/tests/clifft_tests` (full C++ suite) — 113390 assertions / 650 cases pass
- [ ] CI green

### DIFF
--- a/src/clifft/circuit/circuit.h
+++ b/src/clifft/circuit/circuit.h
@@ -43,16 +43,14 @@ struct AstNode {
 
 // A parsed circuit ready for compilation.
 struct Circuit {
-    // Flat list of operations in execution order.
     std::vector<AstNode> nodes;
 
     uint32_t num_qubits = 0;
 
-    // Number of visible measurements (M, MX, MY, MPP, MR, MRX).
-    // R and RX are resets without visible measurements.
+    // Number of visible measurements (entries in the measurement record).
+    // Resets without measurement (R, RX, RY) don't increment this.
     uint32_t num_measurements = 0;
 
-    // Number of DETECTOR declarations.
     uint32_t num_detectors = 0;
 
     // Number of observables (max observable index + 1).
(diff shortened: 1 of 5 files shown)


bachase (commented):
👍 after that second commit

PR #28, opened 2026-04-29
chore: clean up python bindings comments and docstrings

bachase:
## Summary
A five-commit cleanup pass over `src/python/bindings.cc` and `src/python/clifft/__init__.py`. No behavior changes — comments, docstrings, and module-level prose only.

### Comment cleanup
- **c5b5194**: Drop ~13 internal C++ comments that restate the docstring of the very next `m.def(...)` (`// Front-end:`, `// Back-end:`, `// Sample:`, `// Execute:`, etc.) plus a few one-line restatements (`// Register ParseError`, `// Version info`, `// Import the Python SampleResult class`, `// Helper lambda ...`) and a decorative `===` separator. Keeps WHY-bearing comments: `vec_to_numpy` capsule ownership, the sentinel-based enum tripwire, the `BorrowedPass` non-owning wrapper, and the gate-category navigation banners.

### Content alignment
- **7bc4a13**: The package docstring (shown by `help(clifft)`) still described Clifft as "a multi-level AOT compiler and Schrodinger Virtual Machine for quantum circuits (Clifford + T and beyond)" — older internal framing. Re-anchored on the current README/pyproject tagline ("a fast exact simulator for near-Clifford quantum circuits") with a brief pipeline summary. Also drops a stale `(matching the previous default behavior)` parenthetical from the C-extension `compile()` docstring.

### Accuracy fixes
- **d820ded**: Two docstrings disagreed with the code:
  - `default_hir_pass_manager` said "the standard optimization passes" — switched to "the default passes" for parity with the bytecode version.
  - `sample_survivors` and `sample_k_survivors` exposed a `keep_records` argument that wasn't documented. Added a sentence explaining that survivor metadata is always populated, but per-shot record arrays are only filled when `keep_records=True`.
- **c4081ab**: Simplified "256-bit OS hardware entropy" to "hardware entropy" in the `sample()` / `sample_survivors()` seed-default docstrings. The size and source are implementation details; "hardware entropy" is the user-relevant fact.
- **3a3e01a**: Per review feedback, dropped the enumerated default-pass lists from both `default_hir_pass_manager` and `default_bytecode_pass_manager` docstrings — those would silently desync as the default set evolves. Both now simply say "the default passes".

Verified by rendering `help()` output for each touched function — formatting is clean.

## Test plan
- [x] C++ tests: 650/650 pass
- [x] Python tests: 595/595 pass (after editable rebuild)
- [x] `pre-commit run` clean on the changed files
- [x] `help(clifft.compile)`, `help(clifft.sample)`, `help(clifft.default_bytecode_pass_manager)`, etc. render correctly

### DIFF
--- a/src/python/bindings.cc
+++ b/src/python/bindings.cc
@@ -814,9 +814,9 @@ NB_MODULE(_clifft_core, m) {
         nb::arg("bytecode_passes") = nb::none(),
         "Compile a quantum circuit string to executable bytecode.\n\n"
         "Runs the full pipeline: parse -> trace -> [HIR optimize] ->\n"
-        "lower -> [bytecode optimize].  Pass manager arguments are\n"
+        "lower -> [bytecode optimize]. Pass manager arguments are\n"
         "optional; when None the corresponding optimization stage is\n"
-        "skipped (matching the previous default behavior).\n"
+        "skipped.\n"
         "\n"
         "When normalize_syndromes=True, a noiseless reference shot is\n"
         "executed internally to extract expected detector and observable\n"
--- a/src/python/bindings.cc
+++ b/src/python/bindings.cc
@@ -501,9 +501,7 @@ NB_MODULE(_clifft_core, m) {
 
     m.def(
         "default_hir_pass_manager", []() { return clifft::default_hir_pass_manager(); },
-        nb::rv_policy::move,
-        "Return an HirPassManager pre-loaded with the default passes:\n"
-        "PeepholeFusionPass, StatevectorSqueezePass.");
+        nb::rv_policy::move, "Return an HirPassManager pre-loaded with the default passes.");
 
     nb::class_<clifft::BytecodePass>(m, "BytecodePass",
                                      "Abstract base class for bytecode optimization passes.\n\n"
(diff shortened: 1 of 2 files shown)


src/python/bindings.cc
```diff
@@ -571,7 +568,7 @@ NB_MODULE(_clifft_core, m) {
         nb::rv_policy::move,
         "Return a BytecodePassManager pre-loaded with the default passes:\n"
         "NoiseBlockPass, MultiGatePass, ExpandTPass, ExpandRotPass, SwapMeasPass,\n"
-        "SingleAxisFusionPass.");
+        "TileAxisFusionPass, SingleAxisFusionPass.");
```
bachase: This will get stale quickly -- just say the default passes. Users can read the docs.

src/python/bindings.cc
```diff
@@ -506,7 +502,8 @@ NB_MODULE(_clifft_core, m) {
     m.def(
         "default_hir_pass_manager", []() { return clifft::default_hir_pass_manager(); },
         nb::rv_policy::move,
-        "Return an HirPassManager pre-loaded with the standard optimization passes.");
+        "Return an HirPassManager pre-loaded with the default passes:\n"
```
bachase: This will get stale quickly -- just say the default passes. Users can read the docs.

src/python/bindings.cc
bachase: Good call — fixed in 3a3e01a, both docstrings now just say 'the default passes'.

src/python/bindings.cc
bachase: Good call — fixed in 3a3e01a, both docstrings now just say 'the default passes'.

Issue #45, opened 2026-05-01
Support arbitrary circuit qubit counts at runtime


bachase:
## Problem or motivation

Clifft currently has a compile-time maximum qubit count controlled by `CLIFFT_MAX_QUBITS`. This keeps fixed-width Pauli masks fast and predictable, but it creates a poor user experience for circuits whose physical qubit count exceeds the build setting: users must know the limit ahead of time, rebuild with a larger value, and potentially trade away small-circuit performance for larger-circuit compatibility.

The simulator already distinguishes total physical qubits from active rank, and many target workloads have large physical qubit counts while keeping the active rank modest. Those circuits should not require a custom rebuild just because their qubit indices exceed the default compile-time cap.

## Proposed solution

Support choosing the required total-qubit capacity at runtime, based on the parsed circuit, while preserving Clifft's existing performance goals and the VM's active-rank-centered execution model.

The expected user-facing behavior is that a normal build can compile and run circuits with arbitrary physical qubit counts, subject to practical memory and runtime limits, without requiring `CLIFFT_MAX_QUBITS` to be configured in advance.

## Alternatives considered

- Keep the current compile-time setting and document rebuild instructions. This is simple and fast, but it pushes internal implementation constraints onto users.
- Increase the default compile-time limit. This reduces how often users hit the cap, but does not remove the underlying limitation and may penalize common smaller workloads.
- Add runtime-sized storage for full-qubit Pauli/frame data. This improves usability, but needs care to minimize allocation overhead and preserve locality on hot paths.

PR #49, opened 2026-05-01
feat(util): add runtime-width Pauli mask views and arena

bachase:
## Summary

PR1 of the issue #45 staged migration. Introduces the runtime-width Pauli mask primitives and converts the first algorithm (`localize_pauli`) plus its pivot helpers to use them. Updated to address review feedback (commit 806a29d).

## What's in this PR

**New types in `src/clifft/util/`:**
- `BasicMaskView<Word>` (`mask_view.h`) — non-owning runtime-width reference around `std::span<Word>`. Aliases `MaskView = BasicMaskView<const uint64_t>` and `MutableMaskView = BasicMaskView<uint64_t>` preserve a const/mutable distinction; mutating ops gated behind `requires(!std::is_const_v<Word>)`.
- `PauliMaskArena` (`pauli_arena.h`) — fixed-capacity contiguous (X, Z, sign) storage indexed by an opaque `enum class PauliMaskHandle`. Pre-allocated at construction so view references stay valid for the arena's lifetime. Parallel `x_`/`z_` arrays plus `signs_` byte vector.
- `PauliMaskView` / `MutablePauliMaskView` accessor wrappers around the arena slots.
- `view()` / `mut_view()` adapters expose any `BitMask<N>` through the runtime view API.

**Algorithm conversion + bug fix:**
- `find_dormant_pivot`, `find_active_pivot` now operate on `MaskView` and correctly handle `k >= 64`. Previously they assumed `k < 64` (the SVM peak_rank cap) and returned the wrong axis when bit 64 was set with `k > 64` — a latent bug exposed by the migration. Declarations moved into `compiler_context.h` so unit tests can call them directly.
- `localize_pauli` body operates on views; signature unchanged.

**Tests:** `tests/test_mask_view.cc` covers view bit ops across word boundaries, BitMask adapters, arena handle allocation including `width=1024` (above the current cap), and 16 pivot-helper cases that walk the boundary at `k = 0, 30, 63, 64, 70, 128`.

## Test plan

- [x] `uv tool run pre-commit run --all-files`
- [x] `ctest --test-dir build/cmake` — 681 cases pass
- [x] Local Release benches: deltas within inter-run noise envelope on this VM. `localize_pauli` is compile-time-only and the bench `BENCHMARK` lambda only runs `sample()`, so no measurable effect is expected or observed.

### DIFF
--- a/src/clifft/backend/backend.cc
+++ b/src/clifft/backend/backend.cc
@@ -1,6 +1,7 @@
 #include "clifft/backend/backend.h"
 
 #include "clifft/backend/compiler_context.h"
+#include "clifft/util/mask_view.h"
 
 #include <algorithm>
 #include <bit>
@@ -399,38 +400,39 @@ namespace internal {
 // Given a non-identity PauliString P = X^x Z^z, computes virtual Clifford V
 // such that V P V^dag = (+/-)P_v where P_v in {X_v, Z_v}.
 
-// Find a dormant pivot in a BitMask, exploiting the active rank bound.
-// Since k_max <= 30, the active_mask only affects word[0]. Any set bit
-// in words 1..N is guaranteed dormant -- return it immediately.
-static uint16_t find_dormant_pivot(const PauliBitMask& bits, uint32_t k) {
-    assert(k <= kMaxInlineQubits);
-    for (uint32_t w = 1; w < kMaxInlineWords; ++w) {
-        if (bits.w[w] != 0) {
-            return static_cast<uint16_t>(w * 64 + std::countr_zero(bits.w[w]));
+// Find a dormant pivot, exploiting the active rank bound. Any set bit in
+// word index >= 1 is guaranteed dormant when k_max < 64, so the loop
+// returns immediately on the first such bit. For larger k, the active
+// region spans into higher words and the fallback to lowest_bit() applies.
+static uint32_t find_dormant_pivot(MaskView bits, uint32_t k) {
+    for (uint32_t wi = 1; wi < bits.num_words; ++wi) {
+        if (bits.w[wi] != 0) {
+            return wi * 64 + std::countr_zero(bits.w[wi]);
         }
     }
     // k < 64: mask off the active bits in word[0] to find dormant ones.
     // k >= 64: all of word[0] is active, so no dormant bits there.
-    if (k < 64) {
+    if (k < 64 && bits.num_words > 0) {
         uint64_t active_mask = (k == 0) ? 0ULL : ((1ULL << k) - 1);
         uint64_t dormant_w0 = bits.w[0] & ~active_mask;
         if (dormant_w0 != 0) {
-            return static_cast<uint16_t>(std::countr_zero(dormant_w0));
+            return std::countr_zero(dormant_w0);
         }
     }
-    return static_cast<uint16_t>(bits.lowest_bit());
+    return bits.lowest_bit();
 }
 
-// Find an active pivot in a BitMask (prefer bits < k in word[0]).
-static uint16_t find_active_pivot(const PauliBitMask& bits, uint32_t k) {
-    assert(k <= kMaxInlineQubits);
+// Find an active pivot (prefer bits < k in word[0]).
+static uint32_t find_active_pivot(MaskView bits, uint32_t k) {
+    if (bits.num_words == 0)
+        return bits.lowest_bit();
     // k < 64: mask to just the active region of word[0].
     // k >= 64: all of word[0] is active, use it directly.
     uint64_t active_w0 = (k == 0) ? 0ULL : (k >= 64) ? bits.w[0] : bits.w[0] & ((1ULL << k) - 1);
     if (active_w0 != 0) {
-        return static_cast<uint16_t>(std::countr_zero(active_w0));
+        return std::countr_zero(active_w0);
     }
-    return static_cast<uint16_t>(bits.lowest_bit());
+    return bits.lowest_bit();
 }
 
 LocalizationResult localize_pauli(CompilerContext& ctx,
@@ -439,14 +441,16 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
     const uint32_t words = (n + 63) / 64;
 
     PauliBitMask x_bits, z_bits;
+    MutableMaskView x_view = mut_view(x_bits);
+    MutableMaskView z_view = mut_view(z_bits);
     for (uint32_t w = 0; w < words && w < kMaxInlineWords; ++w) {
-        x_bits.w[w] = pauli.xs.u64[w];
-        z_bits.w[w] = pauli.zs.u64[w];
+        x_view.w[w] = pauli.xs.u64[w];
+        z_view.w[w] = pauli.zs.u64[w];
     }
 
-    assert(!(x_bits | z_bits).is_zero() && "Cannot localize identity Pauli");
+    assert((!x_view.is_zero() || !z_view.is_zero()) && "Cannot localize identity Pauli");
 
-    uint16_t pivot;
+    uint32_t pivot;
     LocalizedBasis basis;
     bool sign = pauli.sign;
 
@@ -454,25 +458,26 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
 
     uint32_t k = ctx.reg_manager.active_k();
 
-    if (!x_bits.is_zero()) {
+    if (!x_view.is_zero()) {
         // =============================================================
         // Case 1: X-support is non-empty.
         // =============================================================
 
         // Pick pivot from X-support, preferring dormant axes (>= k).
-        pivot = find_dormant_pivot(x_bits, k);
-        bool z_pivot = z_bits.bit_get(pivot);
+        pivot = find_dormant_pivot(x_view, k);
+        bool z_pivot = z_view.bit_get(pivot);
 
         // X-localization: CNOT(pivot -> q) annihilates X_q.
         // Z propagation: Z_t -> Z_c Z_t, so z_pivot ^= z_q.
         // Sign rule: CNOT(c,t) flips sign iff x_c & z_t & (x_t ^ z_c ^ 1).
         // Here x_pivot=1, x_q=1, so: sign ^= z_q & z_pivot.
-        PauliBitMask to_clear_x = x_bits;
+        PauliBitMask to_clear_x_buf = x_bits;
+        MutableMaskView to_clear_x = mut_view(to_clear_x_buf);
         to_clear_x.bit_set(pivot, false);
         while (!to_clear_x.is_zero()) {
-            uint16_t q = static_cast<uint16_t>(to_clear_x.lowest_bit());
-            emit_cnot(ctx, pivot, q);
-            if (z_bits.bit_get(q)) {
+            uint32_t q = to_clear_x.lowest_bit();
+            emit_cnot(ctx, static_cast<uint16_t>(pivot), static_cast<uint16_t>(q));
+            if (z_view.bit_get(q)) {
                 if (z_pivot)
                     sign ^= true;
                 z_pivot ^= true;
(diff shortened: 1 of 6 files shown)


src/clifft/util/mask_view.h
```diff
@@ -0,0 +1,142 @@
...
+
+// Runtime-width Pauli mask views.
+//
+// MaskView and MutableMaskView are non-owning references to a contiguous
+// uint64_t bit array, parameterized only at runtime by `num_words`. They
+// expose the same bit-level operations as BitMask<N> but without the
+// compile-time width constraint, so algorithms written against the view
+// API work uniformly at any width.
+//
+// These types are the foundation for migrating Pauli mask storage off the
```
bachase: Don't include comments about migration or reference types we might remove later -- these comments in code should stand alone and not depend on this related, but not long term relevant, context.

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,91 @@
...
+// Each handle owns its own (X, Z, sign) slot -- no deduplication. This
+// matches the existing per-op mask ownership model in the HIR.
+//
+// Storage layout: parallel `x_` and `z_` vectors, each strided by
+// `num_words_`, plus a `signs_` byte vector. Parallel arrays (rather than
+// interleaved [x|z]) keep the hot APPLY_PAULI / NOISE kernel reading two
+// contiguous strides, which auto-vectorizes more cleanly than interleaved
+// layout when num_words is large.
+//
+// PR1 introduces the type alongside BitMask<N>. Callers (HirModule,
```
bachase: Again, drop comments that are specific to a PR/implementation plan. This code should stand alone going forward.

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,91 @@
...
+        x_.resize(x_.size() + num_words_, 0);
+        z_.resize(z_.size() + num_words_, 0);
+        signs_.push_back(0);
+        return handle;
+    }
+
+    [[nodiscard]] PauliMaskRef get(uint32_t handle) {
+        assert(handle < signs_.size());
+        size_t off = static_cast<size_t>(handle) * num_words_;
+        return {x_.data() + off, z_.data() + off, &signs_[handle], num_words_};
```
bachase: Here and below, if we have 0 qubits, isn't data() nullptr?

src/clifft/backend/backend.cc
```diff
@@ -399,38 +400,39 @@ namespace internal {
...
-    assert(k <= kMaxInlineQubits);
-    for (uint32_t w = 1; w < kMaxInlineWords; ++w) {
-        if (bits.w[w] != 0) {
-            return static_cast<uint16_t>(w * 64 + std::countr_zero(bits.w[w]));
+// Find a dormant pivot, exploiting the active rank bound. Any set bit in
+// word index >= 1 is guaranteed dormant when k_max < 64, so the loop
+// returns immediately on the first such bit. For larger k, the active
+// region spans into higher words and the fallback to lowest_bit() applies.
+static uint32_t find_dormant_pivot(MaskView bits, uint32_t k) {
+    for (uint32_t wi = 1; wi < bits.num_words; ++wi) {
```
bachase: Won't this scans word 1+ and immediately returns the first set bit, even if that bit is still in the active prefix, e.g. k = 70 and bit 64 set? Be sure to unit test the cases here more thoroughly.

src/clifft/util/mask_view.h
```diff
@@ -0,0 +1,142 @@
...
+
+/// Mutable view over a contiguous uint64_t bit array. Implicitly converts
+/// to MaskView for read-only call sites.
+struct MutableMaskView {
+    uint64_t* w;
+    uint32_t num_words;
+
+    constexpr operator MaskView() const { return {w, num_words}; }
+
+    [[nodiscard]] constexpr bool bit_get(uint32_t idx) const {
```
bachase: Duplicating all methods seems a bit off. can we template this instead (like below but not exact)

```
template <class Word>
  struct BasicMaskView {
      std::span<Word> words;

      bool bit_get(uint32_t idx) const;
      bool is_zero() const;
      int popcount() const;
      uint32_t lowest_bit() const;

      void bit_set(uint32_t idx, bool v)
          requires (!std::is_const_v<Word>);
  };
  using MaskView = BasicMaskView<const uint64_t>;
  using MutableMaskView = BasicMaskView<uint64_t>;
```

src/clifft/util/mask_view.h
```diff
@@ -0,0 +1,142 @@
...
+                return i * 64 + std::countr_zero(w[i]);
+        }
+        return num_words * 64;
+    }
+};
+
+/// Mutable view over a contiguous uint64_t bit array. Implicitly converts
+/// to MaskView for read-only call sites.
+struct MutableMaskView {
+    uint64_t* w;
```
bachase: For all of these pointer/num_words, please use std::span<uint64_t> and std::span<const uint64_t>

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,91 @@
...
+#include "clifft/util/mask_view.h"
+
+#include <cassert>
+#include <cstdint>
+#include <vector>
+
+namespace clifft {
+
+struct PauliMaskRef {
+    uint64_t* x;
```
bachase: Its kind of confusing to have both the raw pointers and the accessors, can we be more consistent?

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,91 @@
...
+        // out unique handles, but the per-mask slot has zero width.
+    }
+
+    [[nodiscard]] uint32_t num_words() const { return num_words_; }
+    [[nodiscard]] size_t num_masks() const { return signs_.size(); }
+
+    /// Allocate a new zero-initialized mask. Returns its handle.
+    uint32_t alloc_zero() {
+        uint32_t handle = static_cast<uint32_t>(signs_.size());
+        x_.resize(x_.size() + num_words_, 0);
```
bachase: This resizing will invalidate any existing PauliMaskRef. Do we know the total size we need to allocate ahead of time so we don't need dynamic allocation here?

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,91 @@
...
+    explicit PauliMaskArena(uint32_t num_qubits) : num_words_((num_qubits + 63) / 64) {
+        // num_words_ == 0 is allowed (empty circuits): alloc still hands
+        // out unique handles, but the per-mask slot has zero width.
+    }
+
+    [[nodiscard]] uint32_t num_words() const { return num_words_; }
+    [[nodiscard]] size_t num_masks() const { return signs_.size(); }
+
+    /// Allocate a new zero-initialized mask. Returns its handle.
+    uint32_t alloc_zero() {
```
bachase: This name could be more informative ... maybe create_zero_mask or something?

Also, should we type the handle with an enum class to avoid misusing?

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,91 @@
...
+
+#include "clifft/util/mask_view.h"
+
+#include <cassert>
+#include <cstdint>
+#include <vector>
+
+namespace clifft {
+
+struct PauliMaskRef {
```
bachase: I'd be consistent with the maskview names, so maybe this is MutablePauliMaskView (or Ref) and the one below is PauliMaskView (or Ref?).

src/clifft/util/mask_view.h
bachase: Done — dropped the migration comment.

src/clifft/util/pauli_arena.h
bachase: Done — dropped the PR-plan comment.

src/clifft/util/pauli_arena.h
bachase: Good catch. Switched to std::span constructed from vector iterators (`v.begin() + i * num_words_, num_words_`), which is well-formed even when num_words=0 and avoids pointer arithmetic on a possibly-null `data()`.

src/clifft/backend/backend.cc
bachase: Right — this was a latent bug that the migration exposed. The original code assumed k < 64 (the SVM peak_rank cap). Fixed both helpers to walk the partial straddle word and respect the active prefix in any word containing k. Added 16 unit tests covering k = 0, 30, 63, 64, 70, 128 across both helpers.

src/clifft/util/mask_view.h
bachase: Done — collapsed into a single `BasicMaskView<Word>` template with mutating ops gated behind `requires(!std::is_const_v<Word>)`. Aliases `MaskView = BasicMaskView<const uint64_t>` and `MutableMaskView = BasicMaskView<uint64_t>` preserve the public API.

src/clifft/util/mask_view.h
bachase: Done — switched to std::span throughout.

src/clifft/util/pauli_arena.h
bachase: Done — refs now use accessor-only API (`x()`, `z()`, `sign()`, `set_sign()`); raw pointer fields are private.

src/clifft/util/pauli_arena.h
bachase: Done — arena is now pre-allocated at construction (`PauliMaskArena(num_qubits, num_masks)`). No dynamic resize, so views stay valid for the arena's lifetime. PR2 will count masks up-front from the HIR builder.

src/clifft/util/pauli_arena.h
bachase: Removed the dynamic `alloc_zero` entirely; with fixed capacity, callers index the pre-allocated slots via `at()`/`mut_at()`. Handle is now `enum class PauliMaskHandle : uint32_t {}` for type safety.

src/clifft/util/pauli_arena.h
bachase: Done — renamed to `PauliMaskView` (const) and `MutablePauliMaskView` (mutable) to match the mask-view naming.

src/clifft/backend/compiler_context.h
```diff
@@ -326,5 +327,19 @@ struct LocalizationResult {
...
                                                 const stim::PauliString<kStimWidth>& pauli);
 
+// =============================================================================
+// Pivot selection helpers (exposed for unit testing)
+// =============================================================================
+
+/// Find a virtual axis at index >= k that has a set bit in `bits`. Prefers
+/// dormant axes over active ones. Falls back to the lowest set bit if no
+/// dormant bit is set. Caller asserts `bits` is non-empty.
+[[nodiscard]] uint32_t find_dormant_pivot(MaskView bits, uint32_t k);
```
bachase: Do these belong here, or should they be (free functions) in mask_view.h?

src/clifft/util/mask_view.h
```diff
@@ -0,0 +1,153 @@
...
+#include <concepts>
+#include <cstddef>
+#include <cstdint>
+#include <span>
+#include <type_traits>
+
+namespace clifft {
+
+template <class Word>
+    requires std::same_as<std::remove_const_t<Word>, uint64_t>
```
bachase, line 24: I can live with it, but its a bit weird to me to have this requires uint64_t. I know thats what we end up needing and it probably helps to not worry about arbitrary types here, but does seem a bit awkward.

src/clifft/util/pauli_arena.h
```diff
@@ -0,0 +1,103 @@
...
+    constexpr PauliMaskView(MaskView x, MaskView z, bool sign) : x_(x), z_(z), sign_(sign) {}
+
+    [[nodiscard]] MaskView x() const { return x_; }
+    [[nodiscard]] MaskView z() const { return z_; }
+    [[nodiscard]] bool sign() const { return sign_; }
+
+  private:
+    MaskView x_;
+    MaskView z_;
+    bool sign_;
```
bachase: Note that sign_ isn't a reference here but the other too are. Probably should match the MutableView below and store it as a const pointer to uint8_t?

src/clifft/backend/compiler_context.h
bachase: Done — moved `lowest_bit_at_or_above` and `lowest_bit_below` to `mask_view.h` as free inline functions. They are general mask operations and don't depend on any compiler state. The fallback-to-lowest-bit behavior previously baked into `find_dormant_pivot` / `find_active_pivot` is now an explicit two-step pattern in `localize_pauli` (file-local `pick_dormant_pivot` / `pick_active_pivot` wrappers). Tests now cover the primitives directly.

src/clifft/util/mask_view.h
bachase: Done — dropped the `requires(uint64_t)` constraint. The underlying word type is conveyed by the public aliases (`MaskView = BasicMaskView<const uint64_t>`, `MutableMaskView = BasicMaskView<uint64_t>`) and the operations work for any unsigned word.

src/clifft/util/pauli_arena.h
bachase: Done — `PauliMaskView::sign_` is now `const uint8_t*`, matching the mutable view's `uint8_t*` and the span-based x/z references. Reading the sign now goes through the arena slot instead of a copy taken at view construction.

bachase (commented):
👍

PR #50, opened 2026-05-01
feat(backend, frontend, noise): migrate AOT-side Pauli mask storage to arenas

bachase:
## Summary

PR2 of the issue #45 staged migration. Moves all AOT-side Pauli mask storage off compile-time-fixed `BitMask<N>` and onto runtime-width `PauliMaskArena`, shrinks `HeisenbergOp` to a fixed 16 bytes, and relocates the qubit-count ceiling from the frontend to the backend. SVM frame storage (`p_x`/`p_z` in `SchrodingerState`) and `CLIFFT_MAX_QUBITS` itself are unchanged here; PR3 finishes that.

Five commits, each independently reviewable:

1. **`chore: document arena resize invalidation`** — comment in `pauli_arena.h` flagging the fixed-capacity invariant and the upgrade path if a future pass needs to grow the arena post-construction.
2. **`feat(hir): migrate HIR Pauli mask storage to arena`** — HirModule owns a sized `pauli_masks` arena. `HeisenbergOp` shrinks from a width-dependent struct (32 B at N=64) to a fixed 16 B (4-byte `PauliMaskHandle` + 8-byte payload + 4-byte type/flags/pad), with a new `static_assert`. Op factories become private; HirModule exposes `append_*` builders that allocate the slot, copy mask data, append the op, and return a reference. Module-bound accessors `hir.destab_mask(op)` / `hir.stab_mask(op)` / `hir.sign(op)` replace the old per-op methods. Optimizer passes that repurpose existing ops (PHASE_ROTATION → T_GATE) use new `demote_to_*` helpers that preserve the mask handle. `localize_pauli`'s helpers and conjugation primitives now operate on `MaskView`.
3. **`feat(backend): migrate ConstantPool Pauli mask storage to arena`** — `ConstantPool::pauli_masks` and `exp_val_masks` become arenas, indexed by handles cast from the existing bytecode `cp_mask_idx` / `cp_exp_val_idx` fields. `lower()` pre-counts and allocates. SVM `apply_pauli_to_frame` and `exec_exp_val` read masks via `arena.at(handle)` and operate on `MaskView`. The SVM frame `p_x`/`p_z` is still inline `PauliBitMask` (PR3), so a per-word XOR/popcount bridge sits in `apply_pauli_to_frame`.
[...]

PR #52, opened 2026-05-02
feat: migrate AOT-side Pauli mask storage to runtime-width arenas

bachase:
## Summary

PR2 of the issue #45 staged migration. Replaces fixed-width `BitMask<N>` Pauli storage in HIR, NoiseChannel, and ConstantPool with runtime-width `PauliMaskArena`s, and shrinks `HeisenbergOp` to a fixed 16 bytes. The frontend writes stim PauliString rows directly into arena slots via `stim_to_mask_view`, so qubits beyond `kMaxInlineQubits` round-trip without truncation.

This is a clean rewrite of the work in #50, organized so each commit is correct on landing — no patches-on-patches. #50 will be closed in favor of this.

## Commit summary

1. **`chore: document arena resize invalidation`** — comment in `pauli_arena.h`.
2. **`feat(util): tighten copy_from semantics, add MaskBuf test helper`** — `copy_from` throws (not asserts) on truncation; production same-size case has zero overhead. 1-word `MaskBuf` test helper.
3. **`feat: migrate Pauli mask storage to runtime-width arenas`** — the big one. HirModule + ConstantPool arenas, NoiseChannel becomes `{handle, prob}`, frontend writes stim → arena directly with no PauliBitMask intermediate, optimizer/backend/SVM read masks via arena, `HeisenbergOp` shrinks to 16 B. Includes high-qubit regression tests at n = 70, 150, 200, 513.
4. **`feat(frontend): reject circuits above the VM axis ceiling early`** — `trace()` rejects `n > 65536` so we don't allocate a giant Stim TableauSimulator before `lower()` would reject anyway.
5. **`fix(python): keep HirModule alive through HeisenbergOp wrappers`** — `PyHeisenbergOp` holds `nb::object module_owner`; both `__getitem__` and `__iter__` capture `nb::borrow(self)`. Includes Python regression tests for both.
6. **`chore(optimizer): clear noise_channel_masks in RemoveNoisePass`** — replace the arena with an empty one after stripping noise.
7. **`test: add surface d=11 r=11 benchmark above the old 128-qubit cap`** — `n ≈ 274`, exercises the regime the migration unlocks.

## Test plan

- [x] `uv tool run pre-commit run --all-files`
- [x] `ctest --test-dir build/cmake -E Bench --timeout 60` — 691 cases pass, full suite in ~3 s
- [x] `uv run pytest tests/python/` — 597 cases pass
- [x] Local Release benches (n=10 samples)

## Bench (Release, this dev machine)

| Benchmark | PR0 baseline | This branch | Δ |
|---|---:|---:|---:|
| `QV-10 x100 shots` | 24.74 ms | 22.9 ms | -7.4% |
| `cultivation-d5 x1000 shots` | 38.95 ms | 39.7 ms | +1.8% |
| `surface-d7-r7 p=1e-3 x10000 shots` | 71.74 ms | 72.3 ms | +0.8% |
| `surface-d5-r5 p=0.05 x10000 shots` | 81.68 ms | 82.4 ms | +0.9% |
| `exp-val 20q 200 probes x100000 shots` | 138.59 ms | 190.4 ms | +37% |
| `surface-d11-r11 p=1e-3 x1000 shots` | (n/a, > old cap) | 18.8 ms | — |

EXP_VAL is the only meaningful regression. As flagged in the migration plan: the previously fully-unrolled `BitMask<128>` popcount/XOR is now a runtime-bounded loop, and the `(X, Z, sign)` tuple is split across three storage regions. PR3 will add template specialization on common `num_words` values for the hot `APPLY_PAULI` / `EXP_VAL` paths to recover this.

## What's still pending (PR3)

- SVM frame `p_x` / `p_z` runtime sizing.
- Removal of `CLIFFT_MAX_QUBITS` and `BitMask<N>`.
- Template specialization on `num_words` for hot mask widths.

### DIFF
--- a/src/clifft/backend/backend.cc
+++ b/src/clifft/backend/backend.cc
@@ -348,8 +348,8 @@ void emit_swap(CompilerContext& ctx, uint16_t a, uint16_t b) {
 }
 
 // Map an HIR Pauli (at t=0) into the current virtual frame.
-stim::PauliString<kStimWidth> map_to_virtual(CompilerContext& ctx, const PauliBitMask& destab_mask,
-                                             const PauliBitMask& stab_mask, bool sign, uint32_t n) {
+stim::PauliString<kStimWidth> map_to_virtual(CompilerContext& ctx, MaskView destab_mask,
+                                             MaskView stab_mask, bool sign, uint32_t n) {
     return ctx.virtual_frame.map_pauli(destab_mask, stab_mask, sign, n);
 }
 
@@ -420,10 +420,15 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
     const uint32_t n = ctx.reg_manager.num_qubits();
     const uint32_t words = (n + 63) / 64;
 
-    PauliBitMask x_bits, z_bits;
-    MutableMaskView x_view = mut_view(x_bits);
-    MutableMaskView z_view = mut_view(z_bits);
-    for (uint32_t w = 0; w < words && w < kMaxInlineWords; ++w) {
+    // Runtime-width scratch storage. PauliBitMask (BitMask<128>) would
+    // truncate the input for circuits with n > kMaxInlineQubits, hiding
+    // bits in qubits 128+ and producing a spuriously-empty Pauli that
+    // tripped the identity-Pauli assertion below.
+    std::vector<uint64_t> x_bits(words, 0);
+    std::vector<uint64_t> z_bits(words, 0);
+    MutableMaskView x_view{std::span<uint64_t>(x_bits)};
+    MutableMaskView z_view{std::span<uint64_t>(z_bits)};
+    for (uint32_t w = 0; w < words; ++w) {
         x_view.words[w] = pauli.xs.u64[w];
         z_view.words[w] = pauli.zs.u64[w];
     }
@@ -451,8 +456,8 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
         // Z propagation: Z_t -> Z_c Z_t, so z_pivot ^= z_q.
         // Sign rule: CNOT(c,t) flips sign iff x_c & z_t & (x_t ^ z_c ^ 1).
         // Here x_pivot=1, x_q=1, so: sign ^= z_q & z_pivot.
-        PauliBitMask to_clear_x_buf = x_bits;
-        MutableMaskView to_clear_x = mut_view(to_clear_x_buf);
+        std::vector<uint64_t> to_clear_x_buf(x_bits);
+        MutableMaskView to_clear_x{std::span<uint64_t>(to_clear_x_buf)};
         to_clear_x.bit_set(pivot, false);
         while (!to_clear_x.is_zero()) {
             uint32_t q = to_clear_x.lowest_bit();
@@ -468,8 +473,8 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
         // Z-localization: CZ(pivot, q) clears residual Z_q.
         // Sign rule: CZ(a,b) flips sign iff x_a & x_b & (z_a ^ z_b).
         // Here x_q=0 (already cleared), so sign NEVER flips.
-        PauliBitMask to_clear_z_buf = z_bits;
-        MutableMaskView to_clear_z = mut_view(to_clear_z_buf);
+        std::vector<uint64_t> to_clear_z_buf(z_bits);
+        MutableMaskView to_clear_z{std::span<uint64_t>(to_clear_z_buf)};
         to_clear_z.bit_set(pivot, false);
         while (!to_clear_z.is_zero()) {
             uint32_t q = to_clear_z.lowest_bit();
@@ -497,8 +502,8 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
         // Z-localization: CNOT(q -> pivot) folds Z_q onto pivot.
         // Sign rule: CNOT(c,t) flips sign iff x_c & z_t & (x_t ^ z_c ^ 1).
         // Here x_c=x_q=0, so sign NEVER flips.
-        PauliBitMask to_clear_z_buf = z_bits;
-        MutableMaskView to_clear_z = mut_view(to_clear_z_buf);
+        std::vector<uint64_t> to_clear_z_buf(z_bits);
+        MutableMaskView to_clear_z{std::span<uint64_t>(to_clear_z_buf)};
         to_clear_z.bit_set(pivot, false);
         while (!to_clear_z.is_zero()) {
             uint32_t q = to_clear_z.lowest_bit();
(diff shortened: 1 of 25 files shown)


src/clifft/util/mask_view.h
```diff
@@ -128,12 +129,24 @@ struct BasicMaskView {
             words[i] |= other.words[i];
     }
 
+    /// Copy `other` into this view. The source must fit within the
+    /// destination width: if `other` is narrower, trailing destination
+    /// words are zero-padded; if `other` is wider, the call throws unless
+    /// every excess word is already zero (so no information is lost).
+    /// Use `std::span` truncation at the call site if you genuinely want
+    /// to drop bits (e.g. `dst.copy_from({other.words.first(dst.num_words())})`).
```
bachase: i think the only reason the different widths is supported is for unit tests -- can we rewrite to only support same width copies? For unit tests, is there some other test helper/simpler approach that avoids the need for copy between mixed widths?

```
constexpr void copy_from_same_width(MaskView other) {
    assert(words.size() == other.words.size());
    for (size_t i = 0; i < words.size(); ++i)
        words[i] = other.words[i];
}
```

src/clifft/frontend/hir.h
```diff
@@ -63,10 +59,26 @@ enum class ExpValIdx : uint32_t {};
 // controls only the SIMD register width for internal bit operations.
 constexpr size_t kStimWidth = 64;
 
-// Inline Pauli mask type used throughout the HIR and SVM.
+// Inline Pauli mask retained for legacy debug paths and tests that build
+// expected fixed-width references. New code should use MaskView / arena.
```
bachase: Are these still needed now? Or just temporarily until we finish the SVM migration? If need for SVM migration, leave some clearer marker or todo to remove this when that lands

src/clifft/frontend/hir.h
```diff
@@ -122,38 +130,35 @@ enum class OpType : uint8_t {
     NUM_OP_TYPES        // Sentinel: must remain last for binding completeness checks
 };
 
+/// Sentinel handle value indicating that an op carries no Pauli mask.
+inline constexpr PauliMaskHandle kNoMask = static_cast<PauliMaskHandle>(~uint32_t{0});
```
bachase, line 108: Should this live with/be defined in the pauli mask code?

src/clifft/frontend/hir.h
```diff
@@ -220,235 +213,353 @@ struct HeisenbergOp {
...
-    // --- PHASE_ROTATION accessor (debug-asserted) ---
-
     [[nodiscard]] double alpha() const {
         assert(type_ == OpType::PHASE_ROTATION && "alpha called on non-PHASE_ROTATION op");
         return phase_.alpha;
     }
 
-    // --- Factory Methods ---
+  private:
+    friend struct HirModule;
```
bachase: What do we need this `friend` for?

src/clifft/frontend/hir.h
```diff
@@ -220,235 +213,353 @@ struct HeisenbergOp {
...
+    // Used by the Front-End to write rewound-Pauli data straight from a
+    // stim::PauliString into the arena slot via mask_at(op), avoiding any
+    // fixed-width PauliBitMask intermediate that would silently truncate
+    // qubits beyond kMaxInlineQubits. Pattern:
+    //
+    //     auto& op = hir.append_tgate_empty(dagger);
+    //     auto slot = hir.mask_at(op);
+    //     stim_to_mask_view(rewound.xs, n, slot.x());
+    //     stim_to_mask_view(rewound.zs, n, slot.z());
+    //     slot.set_sign(rewound.sign);
```
bachase: I'm not a fan of creating an incomplete op and then the caller needing to do these things. Is it possible

1. Template with a lambda so the user does it all at the callsite?
2. Return a handle to just the mutable mask view  for that op that the caller sets?

src/clifft/util/introspection.h
```diff
@@ -12,11 +12,11 @@ namespace clifft {
 
 // Format a HeisenbergOp's Pauli mask as a human-readable string.
 // Example: "+X0*Z3" for destab bit 0 and stab bit 3.
-std::string format_pauli_mask(const HeisenbergOp& op);
+std::string format_pauli_mask(const HeisenbergOp& op, const HirModule& hir);
```
bachase: I don't like exposing the HirModule here -- this only needs the pauli mask and sign for the heisenberg op. Can we instead pass the MaskRef or something narrower?

src/clifft/backend/backend.h
```diff
@@ -255,17 +246,24 @@ struct ConstantPool {
     // Global scalar gamma accumulated during compilation
     std::complex<double> global_weight = {1.0, 0.0};
 
-    // Full N-bit Pauli masks for OP_APPLY_PAULI (indexed by cp_mask_idx)
-    std::vector<PauliMask> pauli_masks;
+    // Full N-bit Pauli masks for OP_APPLY_PAULI. Bytecode references slots
+    // by cp_mask_idx, treated as a PauliMaskHandle into this arena.
+    PauliMaskArena pauli_masks;
```
bachase, line 251: This might be overkill, but we now have a bunch of pauli masks for different purposes. Should we leverage the type system to ensure we don't mix the index types from different arenas? Probably not worth it?

src/clifft/backend/backend.cc
```diff
@@ -348,8 +348,8 @@ void emit_swap(CompilerContext& ctx, uint16_t a, uint16_t b) {
 }
 
 // Map an HIR Pauli (at t=0) into the current virtual frame.
-stim::PauliString<kStimWidth> map_to_virtual(CompilerContext& ctx, const PauliBitMask& destab_mask,
-                                             const PauliBitMask& stab_mask, bool sign, uint32_t n) {
+stim::PauliString<kStimWidth> map_to_virtual(CompilerContext& ctx, MaskView destab_mask,
```
bachase, line 351: Do any callsites use the PauliString that is returned? Or should we have this return a clifft "native" type? It coudl be heap allocated/tempory, but confirm to the interface for our mask views.

(Ok to skip if this is not that impactfuly/will get cleaned up in a the final PR)

src/clifft/frontend/frontend.cc
```diff
@@ -75,199 +66,289 @@ void apply_two_qubit_clifford(stim::TableauSimulator<kStimWidth>& sim, GateType
...
 
-HirModule trace(const Circuit& circuit) {
-    if (circuit.num_qubits > kMaxInlineQubits) {
-        throw std::runtime_error(
-            "Circuit exceeds " + std::to_string(kMaxInlineQubits) +
-            "-qubit compile-time limit: " + std::to_string(circuit.num_qubits) + " qubits");
+/// Conservative upper bound on the number of noise channel masks the
+/// trace will emit. Some channels with prob = 0 are skipped, so the
+/// actual count may be lower; the unused arena slots stay zero-init.
+size_t count_noise_channels(const Circuit& circuit) {
```
bachase, line 271: What happens if we had a new noise gate but forget to add it here?

src/clifft/frontend/frontend.cc
```diff
@@ -75,199 +66,289 @@ void apply_two_qubit_clifford(stim::TableauSimulator<kStimWidth>& sim, GateType
...
+                break;
+        }
+    }
+    return count;
+}
+
+/// Pre-count the number of mask-carrying HIR ops the trace will emit.
+/// Must mirror the dispatch in trace().
+size_t count_pauli_masks(const Circuit& circuit) {
+    size_t count = 0;
```
bachase, line 299: What happens if we had a new pauli-mask related gate but forget to add it here?

src/clifft/frontend/frontend.cc
bachase, line 1: Why are so many comments removed in this file? (Asking to understand)

src/clifft/util/introspection.h
```diff
@@ -10,13 +10,13 @@
...
-// Format a HeisenbergOp's Pauli mask as a human-readable string.
+// Format a Pauli mask (X bits, Z bits, sign) as a human-readable string.
 // Example: "+X0*Z3" for destab bit 0 and stab bit 3.
-std::string format_pauli_mask(const HeisenbergOp& op);
+std::string format_pauli_mask(PauliMaskView mask);
 
 std::string op_type_to_str(OpType type);
 
-std::string format_hir_op(const HeisenbergOp& op);
+std::string format_hir_op(const HeisenbergOp& op, const HirModule& hir);
```
bachase: Can we reduce the HirModule dependency here by having the caller have to call hir.mask_view themselves?

PR #56, opened 2026-05-04
perf(playground): O(1) cursor highlight via reverse source maps

bachase:
## Summary

Every click in the source / HIR / bytecode editor was scanning the full forward source maps three times before the decoration set call. For large circuits (e.g. the SHYPS r=3 logical-memory file from PhotonicInc/ComputingEfficientlyInQLDPCCodes, which produces tens of thousands of HIR ops and bytecode instructions) that gave a noticeable hitch on every click.

The three offenders, all in `App.tsx`:

1. **Bidirectional highlight `useEffect`** — two `forEach` scans over `hir_source_map` and `bytecode_source_map` doing `Array.includes(srcLine)` on each entry.
2. **`highlightPC` `useMemo`** — `findIndex` over `bytecode_source_map` doing the same `includes` check.

All three were O(n_hir + n_bc) per cursor change.

## Fix

Build inverse maps once per compile result in a single `useMemo`:

- `hir: Map<srcLine, monacoLineNumbers[]>`
- `bc: Map<srcLine, monacoLineNumbers[]>`
- `firstBc: Map<srcLine, bytecodePc>` for the k-history chart

The highlight effect and `highlightPC` become O(1) `Map.get` lookups. Build cost is O(n_hir + n_bc) once per compile — a rounding error compared to the wasm compile step that produced the data.

Forward maps stay for the HIR / bytecode cursor handlers (which already index into them in O(1)).

## Test plan

- [x] `npm run build` clean
- [x] `npm run lint` clean
- [x] Manual: clicking through source / HIR / bytecode lines on a small circuit produces the same highlights as before.
- [ ] Manual: click latency on the SHYPS r=3 file is no longer noticeable. (Reviewer to verify after the playground rebuilds and deploys.)

### DIFF
--- a/playground/src/App.tsx
+++ b/playground/src/App.tsx
@@ -175,8 +175,7 @@ export default function App() {
   // once when an upstream pass merged provenance from fused / unrolled
   // ops (e.g. REPEAT-block expansion). Dedupe per entry via Set so the
   // reverse arrays don't grow with the duplication factor and Monaco's
-  // decoration set isn't fed redundant ranges -- exactly the fused /
-  // large cases this PR is meant to speed up.
+  // decoration set isn't fed redundant ranges.
   const reverseMaps = useMemo(() => {
     const empty = {
       hir: new Map<number, number[]>(),
--- a/playground/src/App.tsx
+++ b/playground/src/App.tsx
@@ -170,6 +170,13 @@ export default function App() {
   // HIR/bytecode lines correspond to this source line" in O(1) instead
   // of scanning every output line on every click. firstBc is the first
   // bytecode index for a given source line, used by the k-history chart.
+  //
+  // A single forward entry can contain the same source line more than
+  // once when an upstream pass merged provenance from fused / unrolled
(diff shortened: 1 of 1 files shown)


playground/src/App.tsx
```diff
@@ -164,15 +164,56 @@ export default function App() {
...
+  // HIR/bytecode lines correspond to this source line" in O(1) instead
+  // of scanning every output line on every click. firstBc is the first
+  // bytecode index for a given source line, used by the k-history chart.
+  //
+  // A single forward entry can contain the same source line more than
+  // once when an upstream pass merged provenance from fused / unrolled
+  // ops (e.g. REPEAT-block expansion). Dedupe per entry via Set so the
+  // reverse arrays don't grow with the duplication factor and Monaco's
+  // decoration set isn't fed redundant ranges -- exactly the fused /
+  // large cases this PR is meant to speed up.
```
bachase: This comment shouldn't reference a "PR" -- its meant for posterity.

PR #58, opened 2026-05-05
perf(playground): decouple K-history highlight from chart re-render

bachase:
## Summary

The playground hangs (~700-3000 ms) on every click in the source / HIR / bytecode panes for large circuits. Profiling the SHYPS r=3 circuit showed **56% of per-click CPU was inside `getBoundingClientRect`**, called recursively from React commit during a full recharts re-render on every cursor move.

The driver was `<ReferenceLine x={highlightPC}>` sitting inside `<AreaChart>`: any change to `highlightPC` triggered recharts' full measurement pass over every axis tick and SVG element.

## Fix

Two-part change in `playground/src/components/KHistoryChart.tsx`:

1. **Memoize the chart body.** Extract everything inside `<AreaChart>` (axes, area, baseline line, memory-limit reference line) into a `BaseChart = memo(...)` depending only on the data, baseline flag, and theme colors. The `data` + `maxK` arrays are built in a `useMemo` at the parent keyed on `history` / `baselineHistory`. Cursor moves no longer touch this subtree.
2. **Replace the cursor highlight ReferenceLine with a text annotation** rendered in the chart's top-right corner: `PC 1505 · k=0`. Updates cheaply on every click, conveys *more* information than a vertical line on a possibly-flat trajectory, and — critically — does not re-enter recharts.

## Bench (headless Chromium, SHYPS r=3, 10 clicks)

| metric | before (main) | after | Δ |
|---|---:|---:|---:|
| per-click rAF latency median | 725 ms | 16 ms | **45× faster** |
| per-click rAF latency p95 | 3047 ms | 33 ms | **92× faster** |
| long-tasks (≥50 ms) | 18 | **0** | — |
| long-tasks total | 25,967 ms | 0 ms | — |

User-perceived hang gone.

## Visual

`PC 1505 · k=0` annotation lives in the top-right of the chart; the rest of the chart is unchanged. For Clifford-only circuits (where the trajectory is flat at k=0) the annotation is the only useful per-cursor signal anyway. For T-gate-heavy circuits with non-trivial `k(PC)`, the chart's existing area trace remains visible and the annotation calls out the cursor's specific PC and k value.

## Test plan

- [x] `npm run build` clean
- [x] `npm run lint` clean
- [x] Playwright bench: 0 long-tasks, 16 ms median per-click
- [ ] Manual verification by reviewer in real Chrome on the SHYPS r=3 URL

### DIFF
--- a/playground/src/components/KHistoryChart.tsx
+++ b/playground/src/components/KHistoryChart.tsx
@@ -1,3 +1,4 @@
+import { memo, useMemo } from "react";
 import {
   AreaChart,
   Area,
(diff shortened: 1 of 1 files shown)


bachase:
Replaced by #59 — cleaner fix using recharts 3 public hooks (`usePlotArea` / `useXAxisScale`) to render the cursor highlight as a real SVG line, instead of the in-corner text label. Same perf win (16 ms median per-click, 0 long-tasks), better visual.

PR #60, opened 2026-05-05
feat: add exact full-bitstring probabilities

### LINKED ISSUE #51: strong simulation / bitstring probability
### Problem or motivation

Thanks for developing this nice simulator. 

I would like to propose strong simulation as a useful feature. Currently I use Quizx or TSim for this task, but since Clifft seems to be faster than those (at least at sampling), it would be great if this feature was available too.

The task is to query the exact probability of a specified measurement outcome $x$, i.e. $p(x)=|\langle x|\psi\rangle|^2$.  This would make Clifft useful not only as a weak simulator, but also for tasks such as validating individual output probabilities. 

### Proposed solution

This could be exposed with an interface like `clifft.amplitude(program, bitstring)` or `clifft.probability(program, bitstring)`.

A possible route is to propagate the computational-basis effect $\langle x|$ backward through the Clifford frame and projector/Pauli structure, reducing the query to an overlap with only the active state $|\phi\rangle_A$.

### Alternatives considered

In tsim this would be via [probability_of](https://queracomputing.github.io/tsim/latest/reference/tsim/sampler/?h=probability_of#tsim.sampler.CompiledStateProbs.probability_of).

bachase:
## Summary

Implements issue #51 as a unitary-only exact probability query API:

- add `clifft.probabilities(program, bitstrings, *, bit_order="big")` returning a 1D `float64` NumPy array for full computational-basis bitstrings
- add `Program.num_qubits` and support string inputs plus 2D `bool` / `uint8` arrays with consistent `big` / `little` bit-order handling
- evaluate probabilities from the factored SVM state and final tableau instead of expanding through `get_statevector()`
- pack query bitstrings as runtime-width `uint64_t` word masks, so total bitstring length / qubit count is not capped at 64
- reject non-unitary programs in `probabilities()`, and expose `MakeUnitaryPass` for users who intentionally want to query the unitary-only skeleton of a mixed circuit
- document the new API, bit ordering, limitations, and `MakeUnitaryPass`

## Notes

This intentionally supports only full-register basis-state probabilities in v1. Marginals/subsets are left for a separate API because they represent projector marginals rather than single basis-state probabilities.

## Test plan

- [x] `cmake --build build -j`
- [x] `PYTHONPATH=src/python:.venv/lib/python3.13/site-packages .venv/bin/python -m pytest tests/python/test_probabilities.py -q`
- [x] `PYTHONPATH=src/python:.venv/lib/python3.13/site-packages .venv/bin/python -m pytest tests/python/ -q`
- [x] `ctest --test-dir build -E '^Bench:' --output-on-failure`
- [x] `uv run pre-commit run --all-files`

Closes #51


docs/guide/simulation.md
```diff
@@ -76,6 +77,59 @@ print(sv)  # [0.707+0j, 0+0j, 0+0j, 0.707+0j]
...
+""")
+
+ps = clifft.probabilities(program, ["00", "01", "10", "11"])
+print(ps)  # [0.5, 0.0, 0.0, 0.5]
+```
+
+Bitstrings must cover all `program.num_qubits` qubits. By default,
+`bit_order="big"` maps the first character, or first column of a 2D NumPy
+`bool`/`uint8` array, to qubit 0. With `bit_order="little"`, the last character
+or column maps to qubit 0. Bitstrings can be longer than 64 qubits; the
```
bachase: Remove the sentence about bitstrings can be longer than 64 qubits and beyond -- that one is an artifact of a refactor and is not necessary for user context going forward. They simply query whatever bitstrings they want.

src/clifft/optimizer/drop_non_unitary_pass.cc
```diff
@@ -0,0 +1,41 @@
...
+            case OpType::NOISE:
+            case OpType::READOUT_NOISE:
+            case OpType::DETECTOR:
+            case OpType::OBSERVABLE:
+            case OpType::EXP_VAL:
+                return true;
+            case OpType::NUM_OP_TYPES:
+                return true;
+        }
+        return true;
```
bachase, line 27: Would it be safer/more protective to assert if there's a new opcode added but not included here?

tests/python/test_probabilities.py
bachase, line 1: Should we consider any more fuzz style testing against a small state vector simulation reference?

src/python/bindings.cc
```diff
@@ -1068,4 +1076,18 @@ NB_MODULE(_clifft_core, m) {
             return vec_to_numpy(std::move(sv), {n});
         },
         nb::arg("program"), nb::arg("state"), "Expand the SVM state into a dense statevector.");
+
+    m.def(
+        "_probabilities_from_bitmasks",
+        [](const clifft::CompiledModule& program,
+           const std::vector<std::vector<uint64_t>>& basis_masks) {
```
bachase: the raw unit64_t is probably ok, but should we relate this to a specific type in the PauliMask style? I know we don't have an arena so its different. Maybe overkill for now though.

src/python/clifft/__init__.py
```diff
@@ -98,6 +105,87 @@ def _check_multiprocessing_omp() -> None:
 from clifft._sample_result import SampleResult
 
 
+def _basis_masks_from_bitstrings(
+    program: Program,
+    bitstrings: list[str] | tuple[str, ...] | str | object,
```
bachase: THis is pretty broad with object .. should it be a named type like:

`BasisBitstrings = str | Sequence[str] | npt.NDArray[np.bool_] | npt.NDArray[np.uint8]`

src/python/clifft/__init__.py
```diff
@@ -98,6 +105,87 @@ def _check_multiprocessing_omp() -> None:
...
+    bit_order: str,
+) -> list[list[int]]:
```
bachase: List of list is unnecessarily allocated objects -- can we do `npt.NDArray[np.uint64]`? C++ should happily consume that falt view and we can tell it the num extent to loop it?

src/clifft/api/probabilities.cc
```diff
@@ -0,0 +1,473 @@
...
+struct IdentityConstraint {
+    bool static_sign = false;
+    BasisMask sign_mask;
+};
+
+[[nodiscard]] size_t basis_word_count(uint32_t n) {
+    return (static_cast<size_t>(n) + 63U) / 64U;
+}
+
+[[nodiscard]] BasisMask zero_basis_mask(uint32_t n) {
```
bachase, line 57: If we have BasisMask instead be a mask view over a local bitvector -- do we get most of these functions for "free"?

PR #70, opened 2026-05-10
ci: add scheduled benchmark history workflow (#38)

### LINKED ISSUE #38: Add scheduled benchmark history workflow
## Problem or motivation

Clifft has C++ Catch2 benchmarks in `tests/test_benchmarks.cc` and Python pytest-benchmark cases in `tools/bench/`, but we do not currently keep a historical record of benchmark results. This makes performance regressions harder to spot during normal maintenance and release prep.

We should add a lightweight scheduled benchmark workflow that records benchmark results over time.

## Proposed solution

Add a GitHub Actions workflow that can be run manually and on a nightly schedule. The workflow should run the existing benchmark suites and store their results in a way maintainers can inspect over time.

Suggested direction:

- Add a new workflow under `.github/workflows/`.
- Trigger it with `workflow_dispatch` and a nightly `schedule`.
- Use `ubuntu-24.04` for the first version.
- Run existing C++ and/or Python benchmarks without adding new benchmark fixtures.
- Store benchmark history outside the docs deployment branch.
- Add a short docs/development page describing where maintainers can find the benchmark history.

One possible implementation is to use `benchmark-action/github-action-benchmark`, which supports Catch2 and pytest-benchmark output formats, but contributors may propose another simple approach if it satisfies the requirements.

## Acceptance criteria

- A scheduled/manual GitHub Actions workflow runs existing Clifft benchmarks.
- Benchmark history is persisted somewhere maintainers can inspect across runs.
- The implementation does not write to or depend on the existing `gh-pages` docs deployment branch.
- The workflow does not run on every pull request.
- A short docs page explains how to find and interpret the benchmark history.
- The PR explains any benchmark output format/tooling choices.

## Out of scope

- PR-time benchmark comparison comments.
- Failing CI on benchmark regressions.
- Alerting integrations.
- Self-hosted runners.
- Adding new benchmark workloads.
- Reworking the existing benchmark suite.

## Open questions

- Should the first version track both Catch2 and pytest-benchmark results, or start with one suite?
- What storage/viewing mechanism should we use for the history?
- Should the history live on a dedicated branch such as `bench-data`?

bachase:
Closes #38.

## Summary

- New `.github/workflows/bench.yml` runs the existing C++ Catch2 benchmarks (`tests/test_benchmarks.cc`, `[bench]` tag) and the Python pytest-benchmark suites (`tools/bench/`) on a daily schedule (06:17 UTC) and on `workflow_dispatch`. Not run on PRs or push to `main`.
- Results append to a dedicated `bench-data` orphan branch (already bootstrapped on the remote) via `benchmark-action/github-action-benchmark`, in two subdirs so the chart pages do not mash incomparable units:
  - `dev/bench/cpp/` — Catch2 console (ms)
  - `dev/bench/python/` — pytest-benchmark JSON (s)
- Shortens one BENCHMARK label that exceeded Catch2's name column (broke the publisher's parser).
- Pins `CLIFFT_CPU_BASELINE=x86-64-v3` in the bench job for cross-runner consistency.
- New `docs/development/benchmark-history.md` explains where data lives, how to view the chart locally, runner-noise caveats, and that the local-server step is temporary pending a follow-up that hosts the chart.
- `mkdocs.yml` nav entry under Development.

## End-to-end verified pre-merge

A temporary `push:` trigger was added on this branch (then reverted) so we could exercise the full path before writing to `bench-data` was no longer cheaply recoverable. Three runs across the iteration found and fixed:

| # | Failure | Fix |
|---|---|---|
| 1 | C++ publish: action's catch2 parser couldn't parse the bench output | Found via review, not by CI: `tool: pytestbenchmark` is invalid (use `pytest`) and the action expects console reporter, not XML. Fixed in commit `1aa0949`. |
| 2 | C++ publish: parser failed mid-suite — "No benchmark found" | The label `"exp-val 20q 200 probes x100000 shots"` (36 chars) wraps onto two lines in Catch2's console reporter, and the parser bails on any suite that yields 0 results. Shortened to `"exp-val 20q 200 probes x100k"` (28 chars) and added a guarding comment. Commit `69d28cc`. |
| 3 | C++ build: `catch_discover_tests` SIGILL | Default `CLIFFT_CPU_BASELINE=native` emits ISA the next runner's CPU may reject. Pinned to `x86-64-v3`. Commit `7bb9071`. As a bonus this also makes day-to-day timings more comparable. |

After fix 3, both jobs published successfully. Final state of `bench-data`:

```
README.md
dev/bench/cpp/{data.js,index.html}     1 entry
dev/bench/python/{data.js,index.html}  3 entries
```

The four pre-merge entries reference feature-branch SHAs. They are valid runs but if you'd prefer a clean history before merge, force-push `bench-data` back to `6a65f6f` (the bootstrap commit) and the workflow will repopulate from main.

## Design choices worth flagging

- **Two jobs, not one.** A Python failure should not gate the C++ history and vice versa. Confirmed valuable in practice — Python kept publishing while C++ iterated through three failure modes.
- **`bench-data` over `gh-pages`.** Required by #38.
- **Record-only.** `alert-threshold: 200%`, `fail-on-alert: false`, `comment-on-alert: false`.
- **No GitHub Pages serving for the chart.** Pages already serves docs from `gh-pages` (one source per repo). The action still generates a working `index.html`; the docs page explains how to view it locally with `python3 -m http.server`. Hosting the chart is deferred to a follow-up — likely a sibling `clifft-benchmarks` repo, which the data layout (`dev/bench/{cpp,python}/`) already supports without changes.

## Test plan

- [x] `uv tool run pre-commit run --all-files` passes.
- [x] Phase 1 local verification: Catch2 console-out and pytest-benchmark JSON shapes match what the action's parsers consume (verified by simulating the parser locally with `node`).
- [x] `bench-data` orphan branch bootstrapped on origin in a separate commit (`6a65f6f`).
- [x] **End-to-end run via temporary push trigger.** Both jobs succeeded; data landed in expected paths; chart accumulates across runs (Python has 3 entries).
- [x] Temporary push trigger reverted (`2fa2676`).


bachase:
Updated to publish directly to \`gh-pages\` under \`bench/{cpp,python}/\` rather than a separate \`bench-data\` branch. The earlier rationale (gh-pages is the docs site's exclusive territory) no longer holds — after versioned-docs/mike landed, gh-pages already hosts mike's versioned docs, the playground, PR-preview deploys, and the release workflow's stable refresh, all under non-overlapping subpaths. Bench charts fit in the same model.

Changes since the original branch:

- Rebased onto current main.
- \`bench.yml\`: \`gh-pages-branch: gh-pages\`, \`benchmark-data-dir-path: bench/{cpp,python}\`, concurrency group switched to \`docs-deploy\` (shared with docs.yml/release.yml/PR-preview).
- \`docs/development/benchmark-history.md\`: drops the local-server workaround, links directly to the soon-to-be-live charts at <https://unitaryfoundation.github.io/clifft/bench/cpp/> and <…/python/>.

Post-merge cleanup: delete the orphan \`bench-data\` branch on origin. The four entries currently there reference feature-branch SHAs and aren't worth migrating; the schedule will start a fresh history on \`gh-pages\` the next time it fires.

PR #75, opened 2026-05-12
test: cover SchrodingerState::reset() reuse semantics

bachase:
## Summary

\`reset()\` is implicitly exercised by \`sample()\`'s multi-shot loop, but the invariant it must guarantee — *after \`reset()\`, the state is logically equivalent to a freshly constructed instance with the same config* — has no dedicated test coverage today. The forthcoming \`probability_of()\` API will reuse a single \`SchrodingerState\` across many records via \`reset()\`, so it's worth locking down that invariant before the new path lands.

Seven new directed tests in \`tests/test_state_reset.cc\`:

- \`reset()\` on a freshly-constructed state is a no-op.
- \`reset()\` returns a mutated unitary state to t=0.
- \`reset()\` returns a mutated measured state to t=0.
- **\`execute → reset → execute\` matches \`execute\` on a fresh state** for a Clifford+T circuit (the load-bearing reuse invariant).
- Same for a measured circuit (with PRNG reseed for outcome reproducibility).
- \`reset()\` preserves the \`v_\` allocation pointer.
- \`reset()\` preserves \`p_x\` / \`p_z\` buffer sizing.

No implementation changes. The existing \`reset()\` already meets the contract; this PR just locks it down.

## Test plan

- [x] \`ctest --test-dir build -E Bench\` — 719/719 pass (+7 new).
- [x] Pre-commit clean.

### DIFF
--- a/tests/test_state_reset.cc
+++ b/tests/test_state_reset.cc
@@ -1,0 +1,211 @@
+// Tests for the SchrodingerState::reset() reuse invariant:
+// after reset() the state must be safe to feed to execute() again, with
+// the result matching what execute() would produce on a freshly
+// constructed instance with the same config. The underlying v_, p_x,
+// and p_z allocations must be preserved across the call so that reuse
+// is allocation-free.
+//
+// Note: reset() does not unconditionally clear meas_record / det_record.
+// Those buffers may carry stale outcome bits from a prior non-discarded
+// run; execute() overwrites each slot it touches before any subsequent
+// kernel reads it.
+
+#include "clifft/backend/backend.h"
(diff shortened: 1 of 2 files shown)


tests/test_state_reset.cc
```diff
@@ -0,0 +1,251 @@
+// Directed tests for SchrodingerState::reset() reuse semantics.
+//
```
bachase: this comment must stand alone for future readers to understand what the file is, regardless of the sate of probability_of. Make this comment just a brief sumamry that this tests the semantics of reset/reuse

PR #81, opened 2026-05-13
feat: add probability_of Python API and measurement_qubits accessor

bachase:
## Summary

Exposes \`probability_of\` to Python and adds a \`measurement_qubits\` accessor on \`Program\`, completing the user-facing surface for the strong-simulation API.

### Public API

\`\`\`python
clifft.probability_of(program, records, *, return_log=False)
\`\`\`

Returns the probability that \`sample()\` would emit each measurement record, with the same dust-clamping semantics as the sampling path. Polymorphic on input:

| Input shape | Output |
|---|---|
| Single string \`"010"\` | 1D array of length 1 |
| Sequence of strings | 1D array of length \`len(records)\` |
| 2D \`bool\` or \`uint8\` array \`(N, num_measurements)\` | 1D array of length \`N\` |

Returns linear probabilities by default; pass \`return_log=True\` to get natural-log values that survive underflow on deep circuits.

\`\`\`python
program.measurement_qubits  # list[int]
\`\`\`

For each visible measurement record slot \`i\`, the qubit \`measurement_qubits[i]\` that was measured. Lets users map record positions back to physical qubits without re-parsing the source.

### Wired in

- New \`_probability_of_from_records\` nanobind helper takes a 2D \`uint8\` array and returns a 1D \`float64\` array of log-probabilities.
- New \`_records_from_outcomes\` Python helper normalizes the polymorphic input into a row-major \`uint8\` array (parallel to the existing \`_basis_masks_from_bitstrings\`).
- \`measurement_qubits\` is populated in \`HirModule\` during \`trace()\` from the per-target qubit indices on \`M\` / \`MX\` / \`MY\` ops (hidden reset measurements are excluded by virtue of the existing reject in \`probability_of\`), then copied through \`lower()\` to \`CompiledModule\`, then exposed via nanobind.

### Tests

**31 new Python tests** in \`tests/python/test_probability_of.py\` covering:

- Closed-form correctness (Bell state, |+>, deterministic, feedback).
- Input polymorphism (string / sequence / bool array / uint8 array / empty).
- \`return_log\` toggle.
- Equivalence with \`probabilities()\` on terminal-M-all circuits, both Clifford and Clifford+T.
- Qiskit cross-check for seeded random circuits (\`n in {2, 3, 4}\`).
- Empirical \`sample()\` frequency match against \`probability_of\` predictions (200K shots).
- \`measurement_qubits\` accessor on a measured program and an empty list on a unitary one.
- Full rejection contract: zero-measurement programs, hidden slots, noise, detectors, observables, wrong record length, invalid characters, wrong array shape / dtype / contents, invalid input types.

**2 new C++ tests** in \`tests/test_probability_of.cc\` covering the underlying \`CompiledModule.measurement_qubits\` field directly.

## Test plan

- [x] \`ctest --test-dir build -E Bench\` — 764/764 pass.
- [x] \`uv run pytest tests/python/\` — 666/666 pass.
- [x] Pre-commit clean.

### DIFF
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -29,6 +29,17 @@ add_library(clifft_core STATIC
     util/introspection.cc
 )
 
+# probability_of() uses -inf as an in-band sentinel for unreachable records
+# and surfaces it through the Python API. The Release-only -ffast-math (set
+# in the top-level CMakeLists.txt) implies -ffinite-math-only, which lets
+# the compiler assume infinities never occur and folds away the sentinel.
+# Undo fast-math for just this TU to preserve the contract.
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    set_source_files_properties(api/probability_of.cc
+        PROPERTIES COMPILE_OPTIONS "-fno-fast-math"
+    )
+endif()
+
 # x86-64 runtime dispatch: compile AVX2 and AVX-512 TUs with SIMD flags.
 if((CMAKE_SYSTEM_PROCESSOR MATCHES "(x86_64|AMD64|amd64)") AND NOT MSVC)
     target_sources(clifft_core PRIVATE svm/svm_avx2.cc svm/svm_avx512.cc)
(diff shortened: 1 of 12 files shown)


bachase:
Superseded by #82 (cleaner diff: drops measurement_qubits, replaces -inf sentinel with a finite kUnreachableLogProb so we keep -ffast-math globally).

PR #84, opened 2026-05-13
feat!: rename probability APIs and unify strong-simulation docs

bachase:
## Summary

Breaking rename of the two strong-simulation entry points, plus a documentation restructure that surfaces both side-by-side with their performance tradeoff.

\`\`\`
clifft.probabilities()  -> clifft.basis_probabilities()
clifft.probability_of() -> clifft.record_probabilities()
\`\`\`

The new names make the *object being assigned probability* explicit and put the two APIs on a sharp shared criterion: **is the program's final state pure?** Any measurement breaks that, so:

- \`basis_probabilities()\` — Born amplitudes of computational-basis bitstrings. Requires a unitary program; rejects any measurement (terminal or otherwise).
- \`record_probabilities()\` — exact joint probabilities of measurement records (with optional classical feedback). Requires at least one measurement.

C++ symbols, Python wrappers, internal helpers, source filenames, test filenames, and all doc references renamed in lockstep. No compatibility aliases.

## Documentation restructure

- **\`docs/guide/strong-simulation.md\`** now covers both APIs as one tutorial. New \"When to use which\" decision table at the top, then walkthroughs for each API, then a \"Performance on overlapping circuits\" section explaining when one beats the other on terminal-\`M\`-all circuits (StatevectorSqueezePass lowers active rank for the measured form; \`basis_probabilities()\` amortizes evolve cost across queries).
- **\`docs/guide/simulation.md\`** now lists both in the API summary and the \"Exact Probabilities\" subsection states the pure-state / records distinction up front.
- **\`docs/theory/probabilities.md\` -> \`docs/theory/basis_probabilities.md\`** rename plus cross-link to \`record_probabilities()\` for the measurement-bearing side.

## Why this rename

The previous \`probability_of()\` was grammatically incomplete (\"probability of *what*?\") and the previous \`probabilities()\` didn't signal that it requires a unitary program. The pair \`basis_probabilities\` / \`record_probabilities\` names the queried object and lines up directly with the \"is the final state pure?\" decision.

## Test plan

- [x] All 768 C++ tests pass under Release
- [x] All 664 Python tests pass
- [x] \`uv run pytest --codeblocks docs/\` clean (only preexisting README \`pip install\` failure unrelated to this PR)
- [x] \`uv run mkdocs build --strict\` clean
- [x] pre-commit clean

### DIFF
--- a/docs/guide/simulation.md
+++ b/docs/guide/simulation.md
@@ -4,7 +4,7 @@ Clifft's Schrödinger Virtual Machine (SVM) executes compiled programs. The main
 
 - `sample()` for ordinary shot-based sampling
 - `sample_survivors()` for post-selected sampling
-- `basis_probabilities()` for exact computational-basis amplitudes of a unitary program
+- `basis_probabilities()` for exact computational-basis probabilities of a unitary program
 - `record_probabilities()` for exact joint probabilities of measurement records
 - `execute()` and `get_statevector()` for inspecting small final states
 - `sample_k()` and `sample_k_survivors()` for stratified importance sampling
@@ -104,10 +104,14 @@ qs = clifft.record_probabilities(measured, ["00", "01", "10", "11"])
 print(qs)  # [0.5, 0.0, 0.0, 0.5]
 ```
 
-On terminal `M`-all circuits the two return the same numbers, but their
-algorithms differ in important ways covered in
-[Strong Simulation: Exact Probabilities](strong-simulation.md), including
-when one is meaningfully faster than the other.
+Each program goes to exactly one API — `basis_probabilities()` rejects any
+measurement, including terminal `M`-all. To compare them on the same
+circuit, compile two variants (unitary into `basis_probabilities()`,
+unitary + terminal `M`-all into `record_probabilities()`); the returned
+distributions match on the bitstrings the records encode. Their execution
+costs can differ by 100× either way, depending on circuit structure; see
+[Strong Simulation: Exact Probabilities](strong-simulation.md) for the
+tradeoff and when one is meaningfully faster.
 
 If you intentionally want to query the unitary skeleton of a mixed circuit
 with `basis_probabilities()`, compile with a custom HIR pass manager:
--- a/docs/guide/simulation.md
+++ b/docs/guide/simulation.md
@@ -85,7 +85,8 @@ Clifft offers two exact-probability APIs:
 - **`clifft.basis_probabilities(program, bitstrings)`** — computational-basis
   probabilities for a unitary program.
 - **`clifft.record_probabilities(program, records)`** — joint probabilities
-  of measurement records for a circuit that has measurements.
+  of measurement records for a unitary circuit that has measurements (no
+  noise).
 
 ```python
 import clifft
--- a/docs/guide/simulation.md
+++ b/docs/guide/simulation.md
@@ -80,17 +80,12 @@ print(sv)  # [0.707+0j, 0+0j, 0+0j, 0.707+0j]
 
 ## Exact Probabilities
 
-Clifft offers two exact-probability APIs sitting on either side of one
-question: *is the final state pure?*
+Clifft offers two exact-probability APIs:
 
-- **`clifft.basis_probabilities(program, bitstrings)`** — for **unitary**
-  programs only. Returns $|\langle x | U | 0 \rangle|^2$ for each queried
-  bitstring. Rejects any measurement, feedback, noise, detector, observable,
-  or post-selection (`EXP_VAL` probes are allowed but ignored).
-
-- **`clifft.record_probabilities(program, records)`** — for programs **with
-  measurements** (and, optionally, classical feedback). Returns the exact
-  joint probability that `sample()` would assign to each measurement record.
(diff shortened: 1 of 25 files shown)


docs/index.md
```diff
@@ -87,7 +87,7 @@ For QEC workflows, Clifft also supports detector-based post-selection, survivor
 ## What's New in 0.3.0
 
 Clifft 0.3.0 adds strong simulation for unitary circuits with
-`clifft.probabilities()`, which computes exact probabilities for selected
+`clifft.basis_probabilities()`, which computes exact probabilities for selected
```
bachase: We already released 0.3.0 with this name, so don't change it (you can instead say it will be renamed in the next release)

docs/guide/simulation.md
```diff
@@ -77,36 +78,43 @@ print(sv)  # [0.707+0j, 0+0j, 0+0j, 0.707+0j]
...
 ```
 
-Bitstrings must cover all `program.num_qubits` qubits. By default,
-`bit_order="big"` maps the first character, or first column of a 2D NumPy
-`bool`/`uint8` array, to qubit 0. With `bit_order="little"`, the last character
-or column maps to qubit 0. The practical cost is driven by the number of queried
-bitstrings, the circuit size, and the final active rank rather than by dense
-state-vector expansion. Each call re-executes the program once, so pass all
-bitstrings you want to query in one batch.
+Each program goes to exactly one API — `basis_probabilities()` rejects any
```
bachase: I'd drop this paragraph or be briefer and direct users to the dedicated guide/tutorial page. The main message is performance will depend on the circuit>.  Open to alternatives, but this seems a bit too much info. 

it might also make sense to remove the "DropNonUnitary" pass as well; perhaps too much info?

docs/guide/strong-simulation.md
```diff
@@ -1,17 +1,43 @@
...
-
-- You care about a sparse set of output bitstrings.
-- The circuit is wider than is convenient for dense statevector extraction.
-- The active rank stays small, so exact sparse queries are cheap.
-- You want deterministic checks for compiled circuits.
-
-## Warm-up: Bell State
+# Strong Simulation: Exact Probabilities
+
+Clifft provides two exact-probability APIs. They sit on opposite sides of a
```
bachase: I disagree with sides of the question. For here and other spots, I'd instead say the two APIs support only unitary with computaitonal basis probabilities , or probability of measurement records for unitary mixed with measurement.. You can map the former to the latter by adding explicit terminal measurements, but the performance depends on the circuit as discussed below.

docs/guide/strong-simulation.md
```diff
@@ -1,17 +1,43 @@
...
+## When to use which
+
+| Your circuit… | Use |
+|---|---|
+| has no measurements | `basis_probabilities()` |
+| has any measurement (terminal or intermediate) | `record_probabilities()` |
+| has classical feedback (`CX rec[-1] q`, etc.) | `record_probabilities()` |
+| has noise, detectors, observables, or post-selection | neither — use `sample()` |
+
+Each program goes through exactly one API: `basis_probabilities()` rejects
```
bachase: I'd drop this paragraph -- not sure it adds much

src/python/clifft/__init__.py
```diff
@@ -252,13 +256,16 @@ def fill_string_record(out: npt.NDArray[np.uint8], record: str, row: int) -> Non
     return np.ascontiguousarray(bit_array.astype(np.uint8, copy=False))
 
 
-def probability_of(
+def record_probabilities(
     program: Program,
     records: MeasurementRecords,
     *,
     return_log: bool = False,
```
bachase, line 277: In hindsight, is it confusing for this to offer return_log but basis_probabilities doesn't?

docs/guide/strong-simulation.md
```diff
@@ -1,22 +1,18 @@
...
-disjoint regions of program-space. They overlap mathematically when the
-circuit is a unitary prefix followed by terminal `M`-all — but, as the
-[performance section below](#performance-on-overlapping-circuits) explains,
-even there the two strategies have meaningfully different execution costs.
+Clifft provides two exact-probability APIs covering complementary regimes:
+
+- **`clifft.basis_probabilities(unitary_program, bitstrings)`** — exact
+  computational-basis probabilities for a unitary program.
+- **`clifft.record_probabilities(measured_program, records)`** — exact
+  joint probabilities of measurement records for a circuit that contains
```
bachase: how about a "for a unitary circuit" here and in the other section, since we don't support noise at this time?

src/python/clifft/__init__.py
```diff
@@ -194,13 +195,20 @@ def basis_probabilities(
...
+    probabilities map to ``-inf`` in log output.
     """
-    return cast(
+    probs = cast(
         npt.NDArray[np.float64],
         _basis_probabilities_from_bitmasks(
             program, _basis_masks_from_bitstrings(program, bitstrings, bit_order)
         ),
     )
+    if return_log:
```
bachase, line 208: I'm ok with this hack for now, but maybe leave a TODO in the code that we should consider working in logspace in the c++ implementation instead (as that is where the benefit would be)

Issue #104, opened 2026-05-26
Implement noncomputational structural histories MVP


bachase:
## Goal

Implement the first Clifft-native noncomputational structural-history MVP on a long-running feature branch, without merging partial pieces to `main`.

The intent is to model the same broad category of loss/leakage/status noise that sqale-sim is targeting, but in a way that preserves Clifft's current AOT/static-HIR architecture for the first implementation. Python should provide the ergonomic UX/configuration layer; the C++ core should own validation, history sampling, circuit mutation, compilation, and sampling.

## Scope

Create a C++ core model for exact, pre-sampleable noncomputational histories:

- state-independent structural loss;
- known-source transitions;
- classical status-to-status transitions;
- reset/reload restoring a site when the policy allows it;
- concrete per-trajectory statuses, not per-site probability distributions.

The first implementation should reject unknown coherent state-dependent transitions by default. Exact diagonal filters and segmented replan/resume are important later work, but are out of scope for this issue.

## Model Semantics

Use narrow names and keep HIR static.

Core concepts:

- `NonComputationalModel`
- `TransitionInstrument`
- `NonComputationalHistory`
- a status enum covering computational / known computational / leaked / lost

Transition matrices may be accepted as a shorthand for instruments using the convention `T[to, from]`, but the implementation should keep these concepts separate:

- hidden instrument branch;
- visible noncomputational status;
- reported symbol / classifier output.

Important branch rule:

If multiple hidden source-resolving branches map to the same visible status and aggregating those branches would produce a mixed conditional state, the trajectory implementation must sample the hidden instrument branch, not merely the visible status.

For this MVP, avoid that complexity by supporting exact pre-sampleable cases and rejecting unknown coherent source-dependent transitions.

## Circuit Rewrite Semantics

The C++ core should rewrite a sampled structural history into an ordinary static Clifft circuit before the normal trace/lower pipeline.

At structural loss points:

- insert a hidden computational-basis measurement/collapse of the lost qubit;
- discard the hidden outcome;
- do not shift visible measurement records;
- mark the site as lost for downstream rewrite policy.

Downstream operations touching lost/leaked sites should use explicit policy. Ambiguous behavior should reject by default. Visible measurement, detector, and observable layouts must remain stable.

Potential simple MVP policies:

- lost single-qubit gates/noise: drop;
- lost two-qubit gates: drop the affected operation or reject if policy is ambiguous;
[...]

PR #105, opened 2026-05-26
docs(design): noncomputational MVP API and rewrite-policy contract

bachase:
First commit on the noncomputational MVP feature branch (#104). Captures the resolved API and policy shape from the kickoff thread + two rounds of review so subsequent implementation PRs land against an agreed contract.

**Targets the feature branch, not main.** This branch is the long-running integration point; per the issue, partial work does not merge to main.

## What's in the note

- **Two distinct enums (§2.1).** `LevelCategory` (3 values: `Computational`, `Leaked`, `Lost`) is a *level* property in the model. `QubitStatusKind` (4 values: `ComputationalUnknown`, `ComputationalKnown`, `Leaked`, `Lost`) is the *per-qubit runtime* tag carried in the trajectory.
- **Tagged `QubitStatus` (§2.3).** `{kind, level_id}` struct with structural invariants — `ComputationalUnknown` carries `kInvalidLevel`, so source-dependent reads on unresolved qubits cannot accidentally consult a stale level id. Invariants enforced by named accessors (`known_source_level`, `require_classical_source_level`), not by convention.
- **Default sqale-aligned level set**: `g`, `e`, `leak_g`, `leak_e`, `lost` with `basis_bit` for the computational levels.
- **Model-attached configuration** via `NonComputationalModel(levels, initial_state, transitions, classifier, policy)`. nanobind-bound C++ classes (no pydantic). No new parser instructions in this MVP.
- **Transitions fire after the gate, per qubit operand (per-qubit marginal).** Joint correlated multi-qubit instruments are out of scope.
- **`MeasurementClassifier` is column-substochastic**: per-level column sums in [0, 1], deficit = implicit reject probability. Default identity + reject is just `[[1,0,0,0,0], [0,1,0,0,0]]`. No separate `lost_measurement` policy knob — lost-qubit measurement is fully specified by the classifier's `lost` column.
- **Validation split (§4)**:
  - Construction time: shape + probability checks (column sums, classifier substochasticity, unknown-gate rejection, etc.). Source-dependent matrices construct fine.
  - Sample time: §4.2 source-context check. Applying a source-dependent transition to a `ComputationalUnknown` qubit rejects cleanly with op/qubit/instrument named; applying it to a `ComputationalKnown` qubit succeeds.
- **Sample-history → rewrite-to-ordinary-clifft-circuit → compile → sample** pipeline (§5.1). Compile cache deferred.
- **Default rewrite-policy table (§5.2)** keyed on `(operation, QubitStatusKind)` with explicit reject-by-default rows for ambiguous cases.
- **Hidden trace-out at coherent-qubit loss (§5.3)** uses clifft's existing hidden-measurement infrastructure (`src/clifft/frontend/frontend.cc:618-694`): the rewriter inserts an ordinary `R` op; the frontend lowers it to a hidden measurement with its own slot counter and conditional correction. No `rec[-k]` renumbering, no output stripping, no new HIR op.
- **C++ header dependency order** under a new `src/clifft/noncomp/` tree (§6), with a matching **test plan** (§7) including QubitStatus invariant tests and column-substochastic classifier checks.

## Resolved by the two review rounds

- `RL` (loss reload) is a `policy.reset_restores_lost` flag, not a new circuit op.
- Source-independence is *cached* on each `TransitionInstrument` at construction; it is not a construction-time validation gate.
- `KnownComputational` is not a `LevelCategory` enum value; it lives in `QubitStatusKind` as the runtime tag.
- Knownness is energy-basis-specific: `M`/`R` set it; `MX`/`MY`/`RX`/`RY` do not.
- Sqale-compat equalized-rates approximation deferred to §9 as a future opt-in policy knob; MVP rejects strictly.

## Test plan

- [x] Pre-commit clean
- [ ] Reviewer agrees that the tagged `QubitStatus` representation is the implementation contract
- [ ] Reviewer agrees with the §5.2 rewrite-policy table and the §5.3 R-insertion trace-out strategy

### DIFF
--- a/design/noncomputational-mvp.md
+++ b/design/noncomputational-mvp.md
@@ -93,11 +93,14 @@ simulation:
   the energy basis demotes `Known → Unknown`, discarding the level id
   (Hadamard on `|0>` is the canonical example).
 - `(anything) → Leaked` or `Lost`: via a transition jump.
-- `Leaked / Lost → ComputationalKnown`: only via reset/reload, gated
-  by policy. `Leaked → Known` via `R`/`RX`/`RY` is allowed; `Lost →
-  Known` only when `policy.reset_restores_lost` is set. There is no
-  spontaneous coherent return from a leaked level back into a
-  superposition with `g`/`e`.
+- `Leaked / Lost → Computational`: only via reset/reload, gated by
+  policy. Z-basis reset (`R`) lands in `ComputationalKnown(g)`;
+  X/Y-basis reset (`RX`/`RY`) lands in `ComputationalUnknown` (the
+  post-reset state is in a non-energy basis). `Leaked → Computational`
+  via any of these is allowed; `Lost → Computational` only when
+  `policy.reset_restores_lost` is set. There is no spontaneous
+  coherent return from a leaked level back into a superposition with
+  `g`/`e`.
 
 A `Computational`-category level must also declare a `basis_bit`
 (0 or 1) identifying which computational basis state it represents.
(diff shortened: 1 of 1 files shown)


design/noncomputational-mvp.md
```diff
@@ -0,0 +1,602 @@
...
+sample places the site in a `Computational` level whose `basis_bit`
+is not 0 (the SVM default initialization). `Leaked` and `Lost` levels
+have no `basis_bit`.
+
+Default level set for the MVP (sqale-aligned):
+
+| id | label    | intrinsic category | basis_bit | notes                          |
+|----|----------|--------------------|-----------|--------------------------------|
+| 0  | `g`      | Computational      | 0         | logical 0                      |
+| 1  | `e`      | Computational      | 1         | logical 1                      |
```
bachase: Here and below, what is the thinking to not have KnownComputational and Computational as explicit intrinsic categories? Would that be easier to manage downstream semantics?

design/noncomputational-mvp.md
```diff
@@ -0,0 +1,602 @@
...
+| 2  | `leak_g` | Leaked             | —         | metastable / Rydberg "leak g"  |
+| 3  | `leak_e` | Leaked             | —         | metastable / Rydberg "leak e"  |
+| 4  | `lost`   | Lost               | —         | empty trap / vacuum            |
+
+Users may construct a `NonComputationalModel` with a different level
+set, but the default ships unchanged.
+
+### 2.3 Per-trajectory site state
```
bachase: Here and throughout we introduce the concept of a "site". Clifft doesn't currently do that and states qubit instead. I know this is considering additional states outside the normal qubit computational basis, but for consistency would it be clearer to use site?

design/noncomputational-mvp.md
```diff
@@ -0,0 +1,602 @@
...
+    ],
+
+    # Independent per-site initial level distribution.
+    # Indices into levels[]; sums to 1 per site (validated in C++).
+    initial_state=[0.0065, 0.96, 0.0065, 0.027, 0.0],
+
+    # Per-gate transition instruments. Matrix shorthand uses T[to, from]
+    # convention. Keys are gate names from the existing clifft circuit
+    # vocabulary; unknown keys reject at validation time.
+    transitions={
```
bachase, line 205: Is this meaning here "apply this transition after `CZ`"? And which qubits does it get applied to?

design/noncomputational-mvp.md
```diff
@@ -0,0 +1,602 @@
...
+    },
+
+    # Level -> visible-symbol classifier applied at measurement.
+    # Row-stochastic `symbols x levels` matrix using
+    # matrix[symbol_idx, level_id] = P(symbol | level).
+    # Optional; default is identity-on-computational + reject on
+    # leaked/lost (i.e. policy must say what M of a noncomputational
+    # site returns).
+    classifier=clifft.MeasurementClassifier(
+        symbols=["0", "1"],     # 2-symbol binary record by default
```
bachase: You mention a default of identiy + reject, but is that expressible in this matrix form? Otherwise that seems off to not be able to represent teh default behavior in the way we are asking the user to specify behavior

design/noncomputational-mvp.md
```diff
@@ -0,0 +1,602 @@
...
+
+### 4.2 Sample-time check (per (transition, target-site) firing)
+
+When a transition fires on a target site at category `c`:
+
+- `c == KnownComputational`: use the column matching the known level
+  directly. Always allowed.
+- `c == Leaked` or `c == Lost`: use the column matching the level
+  directly. Classical Markov update. Always allowed.
+- `c == Computational` (coherent, unknown): require
```
bachase: Do we need a mode where we force state-independence by reweighting?

bachase (commented):
Also note that the PR description is stale

PR #106, opened 2026-05-27
feat(noncomp): Level, NonComputationalPolicy, QubitStatus value types

bachase:
First implementation PR on the noncomputational MVP feature branch. Lands the foundational value types from §6 of the design note (#105).

## What's in

Three header-only modules under `src/clifft/noncomp/`:

| File | Contents |
|---|---|
| `qubit_status.h` | `QubitStatusKind` enum (4 values — the runtime tag), `QubitStatus` class with private constructor and `_unchecked` static factories, `kInvalidLevel` sentinel, member accessors (`kind()`, `level_id()`, `is_unknown_computational()`, `known_source_level()`, `require_classical_source_level()`). Aggregate-init `QubitStatus{kind, level_id}` does not compile; user code goes through `LevelSet`. |
| `level.h` | `LevelCategory` enum (3 values), `BasisBit` enum (`Zero`/`One`), `Level` struct, **`LevelSet`** — validated wrapper around `std::vector<Level>` that owns the `QubitStatus` factories so level ids stay bound to a specific table. `default_set()` for the `g`/`e`/`leak_g`/`leak_e`/`lost` set. |
| `policy.h` | `NonComputationalPolicy { reset_restores_lost, unknown_source_policy }`. `UnknownSourcePolicy` is a one-value enum (`Reject`) holding the slot for future opt-in modes. |

Plus `tests/test_noncomp_value_types.cc` — 23 Catch2 cases covering construction/validation, the `LevelSet` status factories, the `QubitStatus` accessors and `_unchecked` factories, and the policy defaults.

## Validation rules (enforced in `LevelSet` ctor)

- Level set non-empty, `<=` 128 entries.
- Every `LevelCategory` value is one of the named enum members (unrecognized values reject).
- Every `Computational` level declares a `basis_bit`. `BasisBit` is an enum so the value is structurally `Zero` or `One`.
- `Leaked` levels may carry a `basis_bit` as optional origin metadata.
- `Lost` levels must not carry a `basis_bit`.
- The table contains **exactly one** Computational level with `basis_bit == Zero` and **exactly one** with `basis_bit == One`. Downstream code (visible Z-basis `M`/`R`, initial prep, classifier defaults) needs unambiguous `g`/`e` ids; duplicates or missing canonical levels reject.

`LevelSet`'s `computational_known(id)` / `leaked(id)` / `lost(id)` factories then validate that `id` refers to a level of the matching category in this table before returning a `QubitStatus`. No more `(level_id, levels)` free-function pairs.

## What's not in

Per the design-note staging plan: no `TransitionInstrument`, no `MeasurementClassifier`, no `NonComputationalModel`, no sampler/rewriter, no nanobind bindings. Each lands in a later PR; bindings come once the full surface exists so Python users see one coherent API drop.

## Test plan

- [x] Debug build: full suite passes (786/786 tests, including the 23 new ones)
- [x] Release build: 23 new tests pass

## Notes

- `clifft::` namespace, no nested `noncomp` sub-namespace — matches existing project convention.
- Design note (`design/noncomputational-mvp.md`) updated in the same branch to match: §2.2 (Lost rejects basis_bit, canonical Zero/One rule), §2.3 (`QubitStatus` non-aggregate with `_unchecked` factories), §3.2 (`LevelSet` C++ surface), §4.1 (validation rules), §6 (header dependency order — `qubit_status.h` now precedes `level.h`).
- `validate_levels` is no longer exposed as a free function; its job moved into `LevelSet`'s constructor.

### DIFF
--- a/src/clifft/noncomp/level.h
+++ b/src/clifft/noncomp/level.h
@@ -12,8 +12,16 @@
 //     for basis_bit == One initial samples.
 //   - For LevelCategory::Leaked, basis_bit is optional metadata
 //     (origin / classifier hint). No current code path consumes it.
-//   - For LevelCategory::Lost, basis_bit should be left empty by
-//     convention. A vacated trap carries no useful state.
+//   - For LevelCategory::Lost, basis_bit MUST be empty. "Lost from
+//     |1>" provenance, if ever needed, belongs in an event record or
+//     in distinct lost levels — not the default lost level.
+//
+// LevelSet validation also enforces that the level table contains
+// exactly one Computational level with basis_bit == Zero and exactly
+// one with basis_bit == One. Downstream paths (visible Z-basis
+// measurement, Z-basis reset, initial prep, classifier defaults)
+// need unambiguous g/e ids; duplicates or missing canonical levels
+// reject at LevelSet construction.
 //
 // LevelSet wraps a std::vector<Level>, runs validation in its ctor,
 // and owns the QubitStatus factories that bind a level id to this
@@ -81,17 +89,17 @@ class LevelSet {
     // of the matching category in this table.
     QubitStatus computational_known(uint8_t level_id) const {
         check_kind(level_id, LevelCategory::Computational, "computational_known");
-        return QubitStatus{QubitStatusKind::ComputationalKnown, level_id};
+        return QubitStatus::computational_known_unchecked(level_id);
     }
 
     QubitStatus leaked(uint8_t level_id) const {
         check_kind(level_id, LevelCategory::Leaked, "leaked");
-        return QubitStatus{QubitStatusKind::Leaked, level_id};
+        return QubitStatus::leaked_unchecked(level_id);
     }
 
     QubitStatus lost(uint8_t level_id) const {
         check_kind(level_id, LevelCategory::Lost, "lost");
-        return QubitStatus{QubitStatusKind::Lost, level_id};
+        return QubitStatus::lost_unchecked(level_id);
     }
 
   private:
@@ -106,6 +114,8 @@ class LevelSet {
                                         std::to_string(levels_.size()) +
                                         " entries; max supported is 128");
         }
+        size_t computational_zero_count = 0;
+        size_t computational_one_count = 0;
         for (size_t i = 0; i < levels_.size(); ++i) {
             const Level& lv = levels_[i];
             switch (lv.category) {
@@ -115,18 +125,39 @@ class LevelSet {
                                                     std::to_string(i) +
                                                     ") is Computational but has no basis_bit");
                     }
+                    if (*lv.basis_bit == BasisBit::Zero) {
+                        ++computational_zero_count;
+                    } else {
+                        ++computational_one_count;
+                    }
                     break;
                 case LevelCategory::Leaked:
+                    // basis_bit allowed as optional origin metadata.
+                    break;
                 case LevelCategory::Lost:
-                    // Leaked: basis_bit allowed as optional metadata.
-                    // Lost: basis_bit should be empty by convention but
-                    // is not structurally enforced.
+                    if (lv.basis_bit.has_value()) {
+                        throw std::invalid_argument("LevelSet: level '" + lv.label + "' (id " +
+                                                    std::to_string(i) +
+                                                    ") is Lost and must not carry basis_bit");
+                    }
                     break;
                 default:
                     throw std::invalid_argument("LevelSet: level '" + lv.label + "' (id " +
                                                 std::to_string(i) +
                                                 ") has an unrecognized LevelCategory value");
             }
+        }
+        if (computational_zero_count != 1) {
+            throw std::invalid_argument(
+                "LevelSet: expected exactly one Computational level with basis_bit == Zero, "
+                "got " +
+                std::to_string(computational_zero_count));
+        }
+        if (computational_one_count != 1) {
+            throw std::invalid_argument(
+                "LevelSet: expected exactly one Computational level with basis_bit == One, "
+                "got " +
+                std::to_string(computational_one_count));
         }
     }
 
--- a/src/clifft/noncomp/level.h
+++ b/src/clifft/noncomp/level.h
@@ -1,0 +1,88 @@
+#pragma once
+
+// Level categories and the per-level model record for the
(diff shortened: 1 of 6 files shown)


src/clifft/noncomp/level.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
+
+// Level categories and the per-level model record for the
+// noncomputational trajectory MVP. See design/noncomputational-mvp.md.
```
bachase: Get rid of this comment that is specific to MVP/the implementation plan.

src/clifft/noncomp/level.h
```diff
@@ -0,0 +1,88 @@
...
+    std::string label;
+    LevelCategory category;
+    // Required when category == Computational; must be omitted otherwise.
+    // Identifies the computational basis state (0 or 1) this level
+    // represents. Used by the rewriter to prepend an X prep for
+    // basis_bit == 1 initial samples.
+    std::optional<uint8_t> basis_bit;
+};
+
+// Sqale-aligned default level set: g, e, leak_g, leak_e, lost.
```
bachase: Get rid of reference to sqale. Even though we are following/targetting its capabilities, this should stand alone

src/clifft/noncomp/level.h
```diff
@@ -0,0 +1,88 @@
...
+        throw std::invalid_argument("validate_levels: level set is empty");
+    }
+    if (levels.size() > 128) {
+        throw std::invalid_argument("validate_levels: level set has " +
+                                    std::to_string(levels.size()) +
+                                    " entries; max supported is 128");
+    }
+    for (size_t i = 0; i < levels.size(); ++i) {
+        const Level& lv = levels[i];
+        if (lv.category == LevelCategory::Computational) {
```
bachase: What about invalid enum values?

src/clifft/noncomp/level.h
```diff
@@ -0,0 +1,88 @@
...
+enum class LevelCategory : uint8_t {
+    Computational = 0,
+    Leaked = 1,
+    Lost = 2,
+};
+
+struct Level {
+    std::string label;
+    LevelCategory category;
+    // Required when category == Computational; must be omitted otherwise.
```
bachase: I can live with this, but is there a straightforward way to make illegal states not representable? Like, the computational basis is probably always 0 or 1, not arbitrary number of basis levels.

src/clifft/noncomp/policy.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+// Policy knobs for the noncomputational trajectory MVP. See
```
bachase: Get rid of this transient comment

src/clifft/noncomp/policy.h
```diff
@@ -0,0 +1,29 @@
...
+// design/noncomputational-mvp.md sections 4.2 and 5.2.
+
+#include <cstdint>
+
+namespace clifft {
+
+// How the runtime sampler handles a source-dependent transition that
+// fires on a ComputationalUnknown qubit. The MVP exposes only Reject;
+// the EqualizeRates compat mode (section 9 of the design note) is
```
bachase: Get rid of future work/MVP/ABI language. (This is not an ABI really anyway)

src/clifft/noncomp/qubit_status.h
```diff
@@ -0,0 +1,114 @@
+#pragma once
+
+// Per-qubit runtime status for the noncomputational trajectory MVP.
+// See design/noncomputational-mvp.md section 2.1 (enums) and 2.3
```
bachase: Again, drop the comments that mention MVP or reference the design doc. This should standa lone as a comment describing this impelemntation

src/clifft/noncomp/qubit_status.h
```diff
@@ -0,0 +1,114 @@
...
+// (tagged QubitStatus + invariants).
+//
+// The tag QubitStatusKind splits the computational case by whether
```
bachase: energy-basis is a weird phrasing. I'd be more direction -- Unknown means its in an arbitrary quantum state, computational known means its in a known computational basis state (either |0> or |1>)

src/clifft/noncomp/qubit_status.h
```diff
@@ -0,0 +1,114 @@
...
+    ComputationalKnown = 1,
+    Leaked = 2,
+    Lost = 3,
+};
+
+// Sentinel for the level_id field when no specific level is resolved.
+// Valid only with kind == ComputationalUnknown.
+constexpr uint8_t kInvalidLevel = 0xFF;
+
+struct QubitStatus {
```
bachase: This can be arbitrariliy initialized, so someone could create a `QubitStatus{QubitStatusKind::ComputationalUnknown, 0}`. Should we go with a private struct that has factory methods?

src/clifft/noncomp/qubit_status.h
```diff
@@ -0,0 +1,114 @@
...
+    }
+    if (levels[level_id].category != expected) {
+        throw std::invalid_argument(std::string(factory_name) + ": level '" +
+                                    levels[level_id].label +
+                                    "' has category that does not match the requested kind");
+    }
+}
+}  // namespace detail
+
+inline QubitStatus make_computational_known(uint8_t level_id, std::span<const Level> levels) {
```
bachase: This is a weird API design ... why are we passing a level_id/levels in versus just the specific status to initialize?

tests/test_noncomp_value_types.cc
```diff
@@ -0,0 +1,218 @@
...
+using clifft::NonComputationalPolicy;
+using clifft::QubitStatusKind;
+using clifft::require_classical_source_level;
+using clifft::UnknownSourcePolicy;
+
+// =========================================================================
+// Level
+// =========================================================================
+
+TEST_CASE("Level: default set matches the sqale-aligned schema") {
```
bachase: Not sure we need this test -- default is what it is

bachase (commented):
Ok to merge but have mixed feelings on

1. Whether to model knowncomputational state versus just treat as opaque computational
2. enum BasisBit/Level optional -- confusing to have this for computational but not leaked/lost. 


Will revisit as we get actual examples.

PR #107, opened 2026-05-27
feat(noncomp): TransitionInstrument with cached source-independence flag

bachase:
Second implementation PR on the noncomputational MVP feature branch. Lands `TransitionInstrument` per §3.2 + §4.1 of the design note.

## What's in

`src/clifft/noncomp/transition_instrument.{h,cc}` — square `T[to, from]` matrix bound to a `LevelSet` at construction.

- **`TransitionInstrument::from_matrix(matrix, levels)`** validates:
  - matrix is square with dimension == `levels.size()`
  - every entry is in `[0, 1]`
  - every column sum is in `[0, 1]` (deficit = implicit no-jump weight)
- **Cached flag** `is_source_independent_on_computational`: true iff every `Computational`-category source column is bit-identical within `1e-12` to the others. Source-dependent matrices construct cleanly; applicability is enforced at sample time per §4.2.
- **Accessors**: `num_levels()`, `prob(to, from)`, `column_sum(from)`, `no_jump_weight(from)`, `is_source_independent_on_computational()`. Out-of-range accessors throw with named fields.

The `LevelSet` is consumed at construction time to size-check the matrix and identify Computational columns. The instrument doesn't carry a `LevelSet` reference afterward; downstream consistency is the model's responsibility (next PR).

## What's not in

- Per-source branch enumeration / sampling API — deferred to the sampler PR, where the concrete use shape will pin down what's needed.
- No nanobind bindings yet (whole-surface drop after the model lands).

## Test plan

- [x] Debug build: full suite passes (802/802 tests, including the 16 new ones)
- [x] Release build: 16 new tests pass

16 cases:
- Shape: zero matrix accepted; wrong row count, jagged row width rejected.
- Bounds: negative entry, entry > 1, column sum > 1 all rejected with named fields.
- Accessors: `column_sum` / `no_jump_weight` match hand-built matrices; `prob` respects `T[to, from]` orientation; out-of-range `prob` and `column_sum` throw.
- Source-independence flag: true for all-zero matrix and for matched g/e columns; false when g/e columns differ; only Computational-source columns affect the flag (leak_g column differing has no effect); 1e-13 difference still independent, 1e-9 not.

## Next

PR C: `classifier.{h,cc}` — column-substochastic `(symbols, levels)` map with derived `reject_probability(level_id)`.

### DIFF
--- a/src/clifft/noncomp/transition_instrument.h
+++ b/src/clifft/noncomp/transition_instrument.h
@@ -9,6 +9,18 @@
 // gives the total jump probability out of that source, and the
 // deficit `1 - column_sum` is the implicit no-jump weight.
 //
+// Convention: every matrix entry, diagonal included, represents a
+// discrete *transition event*. A non-zero matrix[a][a] is NOT the
+// probability that source `a` stays at level `a`; it is the
+// probability of a transition event that happens to land back at
+// the source level (e.g., a depolarizing error that ends in the
+// same level). The "nothing happened" branch ("no-jump") lives in
+// the column deficit, not the diagonal. This mirrors the Kraus
+// structure in which each jump is its own Kraus operator and
+// no-jump is the complement; it also makes the no-jump back-action
+// under unknown-coherent sources (the aI + bZ filter) a single
+// special case rather than something carved out of the diagonal.
+//
 // Construction binds the instrument to a LevelSet so the matrix
 // shape can be checked against the level table and the
 // is_source_independent_on_computational flag can be computed.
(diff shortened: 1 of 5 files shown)


src/clifft/noncomp/transition_instrument.h
```diff
@@ -0,0 +1,66 @@
...
+
+    bool is_source_independent_on_computational() const {
+        return is_source_independent_on_computational_;
+    }
+
+  private:
+    TransitionInstrument(std::vector<std::vector<double>> matrix, std::vector<double> column_sums,
+                         bool is_source_independent_on_computational);
+
+    std::vector<std::vector<double>> matrix_;
```
bachase: Can we use flat storage? Probably not speed sensitive here, but every little bit helps.

Later on we might even consider an arena.

PR #108, opened 2026-05-27
feat(noncomp): MeasurementClassifier with cached reject probability

bachase:
Third implementation PR on the noncomputational MVP feature branch. Lands `MeasurementClassifier` per §3.2 + §4.1 check 5 of the design note.

## What's in

`src/clifft/noncomp/classifier.{h,cc}` — column-substochastic `(symbols, levels)` map bound to a `LevelSet` at construction.

- **`MeasurementClassifier::from_matrix(symbols, matrix, levels)`** validates:
  - `symbols` non-empty, no duplicate labels
  - matrix has shape `(symbols.size(), levels.size())`
  - every entry is finite and strictly in `[0, 1]` (NaN-robust under `-ffast-math` via the IEEE bit-pattern finite check; `static_assert(is_iec559)` documents the assumption)
  - every per-level column sum is in `[0, 1]` (tolerated to `1 + 1e-12` then clamped)
- **`reject_probability(level_id)`** is cached at construction as `1 - column_sum`, clamped so it never reports a negative number under floating drift.
- **Accessors**: `num_symbols`, `num_levels`, `symbol_label`, `prob(symbol, level)`, `reject_probability(level)`. All out-of-range accessors throw with a named field.

Internal storage is a single row-major `std::vector<double>` of size `num_symbols * num_levels`, mirroring `TransitionInstrument`'s arena-friendly layout from PR #107.

## What this enables

`reject_probability(lost)` for the default identity-style classifier is `1.0` — that's the "reject" outcome that lets the model say "the policy does not specify what `M` on a lost qubit should return." Users who want random-bit-on-lost configure the classifier's `lost` column to `[0.5, 0.5]`; deterministic-0 is `[1, 0]`; reject is `[0, 0]`. One specification surface, no separate `lost_measurement` policy knob (per the design-note resolution in PR #105).

## What's not in

- No `NonComputationalModel` yet (PR D); the classifier is a building block.
- No nanobind bindings yet (whole-surface Python drop once the model lands).

## Test plan

- [x] Debug build: full suite passes (824/824 tests, including 18 new)
- [x] Release build: NaN-rejection test passes under `-ffast-math`

18 cases cover:
- Construction validation: identity-style accepted; empty symbols, duplicate labels, wrong row/column count, negative/NaN/infinity entry, `nextafter(1.0)` entry, column sum > 1 all rejected with named fields; column-sum-within-tolerance clamps `reject_probability` to 0 (with the same `a + b > 1.0` guard pattern from PR #107).
- Accessors: `prob(s, l)` orientation; `symbol_label` round-trips constructed strings; out-of-range accessors throw.
- `reject_probability`: identity matrix gives 0 on g/e and 1 on leak_g/leak_e/lost; all-zero matrix gives 1 everywhere; random-bit-on-lost (0.5/0.5 column) gives 0 on lost.

## Next

PR D: `model.{h,cc}` — `NonComputationalModel` ties together `LevelSet`, `TransitionInstrument` map, `MeasurementClassifier`, `NonComputationalPolicy`. Construction-time validation per §4.1.

### DIFF
--- a/src/clifft/noncomp/classifier.cc
+++ b/src/clifft/noncomp/classifier.cc
@@ -41,6 +41,14 @@ MeasurementClassifier MeasurementClassifier::from_matrix(std::vector<std::string
     if (symbols.empty()) {
         throw std::invalid_argument("MeasurementClassifier::from_matrix: symbols list is empty");
     }
+    if (symbols.size() > 256) {
+        // The public symbol index APIs use uint8_t. Allowing more than
+        // 256 symbols would make indices >= 256 unaddressable through
+        // prob() / symbol_label() and silently wrap on conversion.
+        throw std::invalid_argument("MeasurementClassifier::from_matrix: symbols list has " +
+                                    std::to_string(symbols.size()) +
+                                    " entries; max supported is 256");
+    }
     {
         std::unordered_set<std::string> seen;
         seen.reserve(symbols.size());
--- a/src/clifft/noncomp/classifier.cc
+++ b/src/clifft/noncomp/classifier.cc
@@ -1,0 +1,146 @@
+#include "clifft/noncomp/classifier.h"
+
+#include <cstdint>
+#include <cstring>
+#include <limits>
+#include <stdexcept>
+#include <string>
+#include <unordered_set>
+#include <utility>
+
+namespace clifft {
+
+// is_finite_robust below assumes IEEE 754 doubles (every Clifft
+// target satisfies this). Make the assumption explicit.
+static_assert(std::numeric_limits<double>::is_iec559,
+              "MeasurementClassifier requires IEEE 754 doubles");
+
+namespace {
+
+// Tolerance for column-sum bounds and the reject-probability clamp.
+// Matches TransitionInstrument's behavior: raw user entries are
+// strict, derived sums tolerate floating drift and clamp on overshoot.
+constexpr double kProbTolerance = 1e-12;
+
+// Release builds use -ffast-math, which implies -ffinite-math-only.
+// That folds away std::isfinite() and lets `v >= 0.0 && v <= 1.0`
+// pass NaN through. Inspect the IEEE 754 bit pattern instead: a
+// non-finite double has all exponent bits set.
+bool is_finite_robust(double v) {
+    uint64_t bits;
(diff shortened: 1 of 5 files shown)


src/clifft/noncomp/classifier.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+// MeasurementClassifier: column-substochastic map from levels to
+// user-facing measurement symbols.
+//
+// matrix[symbol_index][level_id] = P(symbol | level). For each level,
+// the column (taken across symbols) sums to at most 1; the deficit
+// per column is the implicit P(reject | level). A reject outcome
+// means the model has no opinion about what symbol a measurement on
+// a qubit at this level should produce, and the sampler raises
```
bachase, line 10: Why is raising the right semantics here? Shouldn't we require users to fully specify?

src/clifft/noncomp/classifier.h
```diff
@@ -0,0 +1,62 @@
...
+    // floating drift in the column sum.
+    double reject_probability(uint8_t level_id) const;
+
+  private:
+    MeasurementClassifier(std::vector<std::string> symbols, std::vector<double> matrix_flat,
+                          std::vector<double> reject_probs);
+
+    std::vector<std::string> symbols_;
+    // Row-major flat storage: matrix_flat_[symbol * num_levels() + level_id].
+    // Single allocation, contiguous; mirrors TransitionInstrument's
```
bachase: drop this comment on single allocaitno/arena-friendly; just keep the row-major comment.

src/clifft/noncomp/classifier.cc
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+    }
+
+    const size_t s_n = symbols.size();
+    const size_t l_n = levels.size();
+
+    if (matrix.size() != s_n) {
+        throw std::invalid_argument("MeasurementClassifier::from_matrix: matrix has " +
+                                    std::to_string(matrix.size()) + " rows; expected " +
```
bachase, line 68: Can we use std::format instead of the string concatenation throughout? 

YOu can also consider another commit that does that for other code added in this branch.

PR #111, opened 2026-06-01
chore(noncomp): use std::format for validation error messages

bachase:
Replaces `std::to_string` string concatenation with `std::format` across the noncomp validation code added on this branch:

- `classifier.cc` — `MeasurementClassifier::from_matrix` and the index accessors
- `transition_instrument.cc` — `TransitionInstrument::from_matrix` and accessors
- `level.h` — `LevelSet::validate`, `require_in_range`, `check_kind`

`uint8_t` indices are cast to `unsigned` in the format args so they render as numbers — `std::format` treats `unsigned char` as a character type and would otherwise print the glyph (`std::to_string` promoted to int).

Error-message text is preserved; the `ContainsSubstring` assertions in the noncomp test suites still pass (130 assertions, 63 cases, Debug).

### DIFF
--- a/src/clifft/noncomp/classifier.cc
+++ b/src/clifft/noncomp/classifier.cc
@@ -2,6 +2,7 @@
 
 #include <cstdint>
 #include <cstring>
+#include <format>
 #include <limits>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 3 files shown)


bachase:
Abandoning: this is the first `std::format` use in the repo and would raise the source-build compiler floor to GCC 13 / Clang 17 / Xcode 16, breaking the documented GCC 10+/Clang 12+ support (Ubuntu 22.04, Debian 12, older Xcode). The std::to_string concatenation is fine; the cleanup is a convenience that isn't worth narrowing supported compilers.

PR #112, opened 2026-06-01
feat(noncomp): NonComputationalModel with construction-time validation

bachase:
PR D in the noncomputational MVP sequence (after #108). Adds `NonComputationalModel` in `src/clifft/noncomp/model.{h,cc}` — the assembled, validated model that ties together the pieces built in #106–#108 — plus the review-driven hardening below.

## What it does
Constructor takes a `LevelSet`, a per-level `initial_state` distribution, a gate-keyed `std::map<std::string, TransitionInstrument>`, an optional `MeasurementClassifier`, and a `NonComputationalPolicy`, and runs the **construction-time consistency checks** that no single component can see on its own.

- **Initial state is a probability vector** over the levels: one entry per level, each finite and in `[0, 1]` (strict), summing to 1 within `kProbTolerance`. The stored distribution is then **normalized** so no sampler has to compensate for within-tolerance drift.
- **Transition keys are canonicalized to `GateType`.** Stim aliases (`CNOT`→`CX`, `RZ`→`R`, `MZ`→`M`, …) resolve to their canonical gate, and two keys that collapse to the same gate are rejected. The model stores `std::map<GateType, TransitionInstrument>`, so the sampler resolves transitions against the parsed circuit's `GateType` directly — a stored alias spelling can't silently shadow a configured transition.
- **Transition keys must name a gate that supports a transition instrument.** An explicit allowlist that defaults to false (Cliffords, non-Clifford unitaries, measurements except `MPAD`, resets); annotations, identity no-ops, `MPAD`, `EXP_VAL`, and noise channels are rejected, and any newly-added `GateType` is rejected until deliberately classified. Leakage riding on a noise instruction is out of MVP scope until the sampler defines composition semantics.
- **Components are bound by level-table fingerprint.** `LevelSet` exposes a deterministic `fingerprint()` (FNV-1a over each level's `(label, category, basis_bit)` in order); `TransitionInstrument`/`MeasurementClassifier` record it at construction, and the model rejects any component whose fingerprint differs. This catches a same-sized but differently ordered/labeled table that a level-count check accepts.

Per the design contract, **source-context validity stays a sample-time concern** (the sampler checks it against the target qubit's `QubitStatusKind`) and is not enforced at construction.

The `LevelSet` fingerprint guards the **compositional** C++ constructor only. The intended Python-facing path is spec-based (raw matrices built against the model's single `LevelSet`), documented in design note §3.3 and planned for the bindings step — fingerprints will not surface in the Python API.

## Review changes
Multi-LLM review across two rounds:
- **Round 1:** alias canonicalization + duplicate rejection; table-fingerprint binding instead of size-only checks; initial-state normalization; the hookable-gate allowlist excluding noise for the MVP; zero-matrix no-op test helper.
- **Round 2:** hoisted the duplicated `is_finite_robust` / `kProbTolerance` / IEEE-754 `static_assert` into `clifft/noncomp/numeric.h`; inverted the transition-gate check to a fail-closed allowlist; renamed `is_hookable_transition_gate` → `supports_transition` (`"hook"` collides with the QEC hook-error term).

Touches the merged `TransitionInstrument`/`MeasurementClassifier` to add the fingerprint accessor and share the numeric header.

## Tests
`tests/test_noncomp_model.cc` covers happy paths, every rejection branch (length/range/NaN/sum, unknown key, non-supported noise + annotation keys, duplicate canonical keys, wrong-size and wrong-fingerprint instrument/classifier), normalization, alias canonicalization, and accessors. Full Debug suite green (853 cases).

### DIFF
--- a/src/clifft/noncomp/model.cc
+++ b/src/clifft/noncomp/model.cc
@@ -5,6 +5,7 @@
 #include <cstdint>
 #include <cstring>
 #include <limits>
+#include <map>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -35,6 +36,28 @@ bool is_finite_robust(double v) {
     return (bits & kExpMask) != kExpMask;
 }
 
+// A transition instrument models the noncomputational side effects of a
+// physical qubit operation, so it may only key a gate that represents
+// one. Annotations, identity no-ops, the classical measurement pad, the
+// expectation-value probe, and noise channels (whose composition with a
+// level transition is deliberately left undefined in the MVP) are not
+// hookable.
+bool is_hookable_transition_gate(GateType g) {
+    if (g == GateType::UNKNOWN) {
+        return false;
+    }
+    if (gate_arity(g) == GateArity::ANNOTATION) {
+        return false;
+    }
+    if (is_identity_noop(g) || is_noise_gate(g) || is_exp_val(g)) {
+        return false;
+    }
+    if (g == GateType::MPAD) {
+        return false;
+    }
+    return true;
+}
+
 }  // namespace
 
 NonComputationalModel::NonComputationalModel(
@@ -43,10 +66,10 @@ NonComputationalModel::NonComputationalModel(
     std::optional<MeasurementClassifier> classifier, NonComputationalPolicy policy)
     : levels_(std::move(levels)),
       initial_state_(std::move(initial_state)),
-      transitions_(std::move(transitions)),
       classifier_(std::move(classifier)),
       policy_(policy) {
     const size_t n = levels_.size();
+    const uint64_t fingerprint = levels_.fingerprint();
 
     // Initial state must be a probability vector over the levels.
     if (initial_state_.size() != n) {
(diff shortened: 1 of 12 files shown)


src/clifft/noncomp/model.cc
```diff
@@ -0,0 +1,184 @@
...
+// derived-quantity tolerance used by TransitionInstrument and
+// MeasurementClassifier: raw entries are strict, the derived sum
+// tolerates floating drift.
+constexpr double kProbTolerance = 1e-12;
+
+// Release builds use -ffast-math, which implies -ffinite-math-only.
+// That folds away std::isfinite() and lets `v >= 0.0 && v <= 1.0`
+// pass NaN through. Inspect the IEEE 754 bit pattern instead: a
+// non-finite double has all exponent bits set.
+bool is_finite_robust(double v) {
```
bachase: This and the tolerances are repeated in several places. Should we unify to one spot?

src/clifft/noncomp/model.cc
```diff
@@ -0,0 +1,184 @@
...
+    if (gate_arity(g) == GateArity::ANNOTATION) {
+        return false;
+    }
+    if (is_identity_noop(g) || is_noise_gate(g) || is_exp_val(g)) {
+        return false;
+    }
+    if (g == GateType::MPAD) {
+        return false;
+    }
+    return true;
```
bachase: Should we invert and default to false? I wonder about adding a new gatetype and it being "supported" incorrectly out of the box? Or should we do some other check to ensure we at least consider it.

src/clifft/noncomp/model.cc
```diff
@@ -0,0 +1,184 @@
...
+    // and no two spellings of one gate shadow each other. Each instrument
+    // must have been built against this model's level table.
+    std::map<GateType, std::string> first_spelling;
+    for (auto& [name, instrument] : transitions) {
+        const GateType gate = parse_gate_name(name);
+        if (gate == GateType::UNKNOWN) {
+            throw std::invalid_argument("NonComputationalModel: transition key '" + name +
+                                        "' is not a recognized gate name");
+        }
+        if (!is_hookable_transition_gate(gate)) {
```
bachase: hook has a specific meaning in quantum error correction (hook errors), maybe use a different name for this.

PR #113, opened 2026-06-01
feat(noncomp): history sampler + shared status stepper (PR E)

bachase:
PR E in the noncomputational MVP sequence (after #112). Adds the step-3 **history sampler** and the **shared status stepper** it relies on, bundling design steps 7 (`history.h`) and 8 (`sampler.h/.cc`) per the agreed scope.

## What it does
`sample_history(circuit, model, seed)` walks a parsed `Circuit` once and samples one noncomputational trajectory:
1. Sample each qubit's initial level from `model.initial_state()` (independent per qubit).
2. Walk ops in order; for each qubit operand, consult the gate's transition instrument (if any), run the §4.2 source-context check, sample no-jump/jump, and advance the qubit's status.

It records the **canonical, minimal** `NonComputationalHistory` — initial statuses + sampled transition outcomes (one record per consulted operand, jump or no-jump). Per-op statuses are intentionally *not* materialized; they're a derived view via replay through the shared stepper. The sampler does **not** rewrite, compile, or run the SVM, and never consults a measurement outcome. Deterministic in `seed` (`std::mt19937_64` + a portable `[0,1)` extraction, so fixed seed → fixed history across platforms).

## Shared status stepper (`status_step.{h,cc}`)
The §5.2.1 / §5.2.2 status-transition rules (plus §5.2 reset-restore) live in one place so the sampler and the future rewriter advance status identically:
- A transition's **source level is the status at op entry**; the gate's normal effect (e.g. demotion) is applied after.
- A sampled **jump destination wins** over the normal post-op status.
- A Z-basis `M` keeps the **pre-SVM-known** status (§5.2.2): `Known` stays `Known`, `Unknown` stays `Unknown` — it does not promote, since the outcome lives in the SVM.
- `R`/`MR` → `Known(g)`; `RX`/`RY`/`MRX`/`MRY` → `Unknown`; non-destructive probes preserve status; a Leaked/Lost qubit changes only on a restoring reset (Lost gated by `policy.reset_restores_lost`).

## §4.2 enforcement
A source-dependent transition firing on a `ComputationalUnknown` qubit raises, naming the op index, qubit, and gate. `Known`/`Leaked`/`Lost` sources use their level's column directly; source-independent instruments are allowed on `Unknown` (columns are identical, so `g`'s column is used).

## Supporting change
`LevelSet` gains `computational_zero_id()` / `computational_one_id()` and `status_for(level_id)` (category-dispatched `QubitStatus`), which the stepper needs.

## Design note
Includes the §5.2.2 clarification (history status is pre-SVM-known, not trajectory-physical) settled in review.

## Tests
`test_noncomp_status_step.cc` (stepper rules) and `test_noncomp_sampler.cc` (deterministic seed, initial-state marginals, transition firing, §4.2 allow/reject). Full Debug suite green.

## Review changes

Address a multi-LLM review of the sampler:

- **Operand roles (the key fix).** Classical Pauli feedback (`CX`/`CZ` with a record control) was being treated as a physical gate — it would consult the gate's transition instrument and demote the target like an entangler. A new shared `op_role.{h,cc}` classifier resolves each node's qubit operands and their role (`Physical` vs `Feedback`), skipping record references and `MPAD` pad literals. The sampler fires **no transition** on a `Feedback` operand, and the stepper demotes a known qubit there (a conditional X can flip `g<->e` on a pre-SVM-unknown control bit) while leaving noncomputational qubits untouched. The stepper now takes the role, not bare `GateType`.
- A qubit operand outside the circuit's range now **throws** instead of being silently skipped.
- Expanded tests: parsed-circuit feedback, `M`-on-Unknown then rejecting source-dependent transition, reset-before-transition acceptance, a **probabilistic** jump frequency that pins `T[to, from]` orientation, multi-target record ordering, and policy-gated lost-qubit reset restoration through the sampler.

## Review changes (round 2)

- **Conditional-Z feedback fix.** Feedback was uniformly demoting the target, but a conditional Z (`CZ rec q`) is phase-only and cannot change the energy level, so it must preserve `ComputationalKnown(g/e)`. Split the feedback role into `FeedbackX` (demotes, may flip `g<->e`) and `FeedbackZ` (preserves); the classifier assigns it from the gate and the stepper branches on the role.
- **`qubit_operands()` invariant enforced.** A record target accompanies a qubit operand only as CX/CZ feedback; rec-only annotations (`DETECTOR`/`OBSERVABLE_INCLUDE`/`READOUT_NOISE`) yield no qubit operands, and a record control with a qubit operand on any other gate now throws rather than being silently treated as feedback.
- Added `tests/test_noncomp_op_role.cc` (feedback roles, rec-only annotations, `MPAD`, the rejection invariant) plus sampler/stepper coverage for conditional-Z preservation.

## Review changes (round 3)

Addresses the two review comments on the sampler (commit `ad78e2a`):

- **Shared xoshiro RNG.** The sampler drew from `std::mt19937_64`; it now uses `Xoshiro256PlusPlus`, the same scientific PRNG the SVM sampler uses. The generator is hoisted out of `svm.h` into a shared `util/xoshiro.h` that both the SVM and the sampler include, so there is one cross-platform-reproducible generator instead of two. The portable `[0, 1)` extraction is unchanged, so a fixed seed still maps to a fixed history.
- **Self-contained comments.** Dropped the design-doc section/step references (`section 4.2`, `step 3`, "the MVP") from the sampler, stepper, and model comments so they stand on their own once the design note is removed. No behavior change; full Debug suite green.

### Added coverage (commit `a316227`)

Pinned the MPP corner that review flagged as easy to get wrong: `qubit_operands` returns an `MPP`'s Pauli-tagged qubits in target order as `Physical`, and an `MPP` with no attached transition demotes each `Known(g)` operand to `Unknown` (the stepper's measurement-collapse fallthrough). Behavior was already correct; these lock it in.

### RNG follow-up (commit `7a8ce1e`)

Made `Xoshiro256PlusPlus` a fully self-contained utility:

- Moved `seed_from_entropy()` from `svm/svm_state.cc` into a new `util/xoshiro.cc`, so the generator no longer has an out-of-line member defined in an SVM translation unit (`<random>` is no longer needed in `svm_state.cc`). This also lets the noncomp side default to entropy seeding later without depending on the SVM.
- Added a canonical `next_double()` to the generator; the SVM's `random_double()` and the sampler both draw through it, so the `[0, 1)` extraction has one definition instead of two copies of the same bit math. Header-inline, so the SVM hot path is unchanged.

No behavior change; Debug and Release suites green.

_Build note (commit `f1aba72`): the wasm target keeps its own `clifft_core` source list in `src/wasm/CMakeLists.txt`; since it compiles `svm_state.cc` (which calls `seed_from_entropy()`), it also needs `util/xoshiro.cc`. Added there too; verified against the `emscripten/emsdk:3.1.74` image._

### DIFF
--- a/src/clifft/noncomp/sampler.cc
+++ b/src/clifft/noncomp/sampler.cc
@@ -76,7 +76,7 @@ HistorySample sample_history(const Circuit& circuit, const NonComputationalModel
             // Feedback corrections are virtual: they never fire a
             // transition. Physical operands consult the instrument (if
             // any) for this gate.
-            if (operand.role == OperandRole::Feedback || instrument == nullptr) {
+            if (operand.role != OperandRole::Physical || instrument == nullptr) {
                 status[qubit] = normal_post_op_status(s_in, gate, operand.role, policy, levels);
                 continue;
             }
--- a/src/clifft/noncomp/sampler.cc
+++ b/src/clifft/noncomp/sampler.cc
@@ -12,19 +12,6 @@
 
 namespace clifft {
 
-namespace {
-
-// Portable [0, 1) draw: the top 53 bits of a 64-bit word scaled to the
-// unit interval. Xoshiro256PlusPlus (the same generator the SVM uses)
-// produces an identical sequence across compilers, and this extraction
-// avoids std::uniform_real_distribution, whose algorithm varies across
-// standard libraries -- together they keep a fixed seed reproducible.
-double next_unit(Xoshiro256PlusPlus& rng) {
-    return static_cast<double>(rng() >> 11) * 0x1.0p-53;
-}
-
-}  // namespace
-
 HistorySample sample_history(const Circuit& circuit, const NonComputationalModel& model,
                              uint64_t seed) {
     const LevelSet& levels = model.levels();
@@ -40,7 +27,7 @@ HistorySample sample_history(const Circuit& circuit, const NonComputationalModel
     // initial-state distribution. The last level catches any
     // floating-point tail so a draw always resolves to a level.
     for (uint32_t q = 0; q < circuit.num_qubits; ++q) {
-        const double u = next_unit(rng);
+        const double u = rng.next_double();
         double acc = 0.0;
         uint8_t chosen = static_cast<uint8_t>(num_levels - 1);
         for (uint8_t l = 0; l < num_levels; ++l) {
@@ -104,7 +91,7 @@ HistorySample sample_history(const Circuit& circuit, const NonComputationalModel
             // Sample the outcome: the no-jump weight occupies [0, w), the
             // jump targets partition [w, 1). last_positive catches a
             // floating-point tail so u >= w always resolves to a jump.
-            const double u = next_unit(rng);
+            const double u = rng.next_double();
             const double no_jump = instrument->no_jump_weight(source_col);
             TransitionOutcome outcome;
             if (u >= no_jump) {
--- a/src/clifft/noncomp/sampler.cc
+++ b/src/clifft/noncomp/sampler.cc
@@ -1,0 +1,137 @@
+#include "clifft/noncomp/sampler.h"
+
+#include "clifft/circuit/gate_data.h"
+#include "clifft/circuit/target.h"
+#include "clifft/noncomp/status_step.h"
+#include "clifft/noncomp/transition_instrument.h"
+
+#include <cstdint>
+#include <random>
+#include <stdexcept>
+#include <string>
+
+namespace clifft {
+
+namespace {
+
+// Portable [0, 1) draw: the top 53 bits of a 64-bit word scaled to the
+// unit interval. std::mt19937_64's sequence is standardized and this
+// extraction avoids std::uniform_real_distribution, whose algorithm
+// varies across standard libraries -- both together keep "fixed seed ->
+// fixed history" stable across platforms.
+double next_unit(std::mt19937_64& rng) {
+    return static_cast<double>(rng() >> 11) * 0x1.0p-53;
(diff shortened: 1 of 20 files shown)


src/clifft/noncomp/sampler.cc
```diff
@@ -0,0 +1,140 @@
...
+
+namespace {
+
+// Portable [0, 1) draw: the top 53 bits of a 64-bit word scaled to the
+// unit interval. std::mt19937_64's sequence is standardized and this
+// extraction avoids std::uniform_real_distribution, whose algorithm
+// varies across standard libraries -- both together keep "fixed seed ->
+// fixed history" stable across platforms.
+double next_unit(std::mt19937_64& rng) {
+    return static_cast<double>(rng() >> 11) * 0x1.0p-53;
```
bachase: Why aren't we using the xoshiro RNG we use for the SVM sampler?

src/clifft/noncomp/sampler.cc
```diff
@@ -0,0 +1,140 @@
...
+
+            // Feedback corrections are virtual: they never fire a
+            // transition. Physical operands consult the instrument (if
+            // any) for this gate.
+            if (operand.role != OperandRole::Physical || instrument == nullptr) {
+                status[qubit] = normal_post_op_status(s_in, gate, operand.role, policy, levels);
+                continue;
+            }
+
+            // Section 4.2 source-context check and column selection.
```
bachase: Don't reference design doc sections/numbers. These comments should be self-contained to the codebase, as we will eventually delete/remove the design document

PR #114, opened 2026-06-02
feat(noncomp): circuit rewriter (PR F)

bachase:
PR F in the noncomputational MVP sequence (after #113). Adds the **circuit rewriter** that replays a sampled trajectory into an ordinary `clifft::Circuit` for the existing compile pipeline.

`rewrite(original, history, model)` does three things, introducing no new instruction kinds:

1. **Initial-state prep** — a leading `X` for each qubit whose sampled initial status is `ComputationalKnown` with a `basis_bit == One` level, so the SVM's `|0...0>` matches the sampled known level.
2. **Per-op policy** — replays each op's per-qubit status through the shared `status_step` stepper, then keeps / drops / rejects the op. Rejections name the op index, qubit, gate, and status.
3. **Hidden trace-out** — inserts an `R` when a coherent carrier jumps to a noncomputational level (see the rule below). The existing reset lowering turns it into a hidden measurement + corrective Pauli — no new visible measurement, no record renumbering.

## Trace-out rule (settled in review, design §5.3 updated)
The `R` decision is made on the **carrier state after the base op's normal effect**, not the entry status, because a gate can make a `Known` qubit coherent before its after-gate transition fires:

```
pre             = status entering the op
post_if_no_jump = normal_post_op_status(pre, op)   # entry status still selects the transition source column
insert R iff outcome jumps to Leaked/Lost AND post_if_no_jump is ComputationalUnknown
```

So `Known(g) → demoted by the gate → jump to lost` **does** get a trace-out, while a still-`Known` atom that is lost does not. The `docs(design)` commit updates the §5.3 summary rule to match.

## Two scoping calls to flag for review
- **Measurements are always kept**, even on leaked/lost operands — dropping one would shift the visible record and break every `rec[-k]` reference. The classifier's *reinterpretation* of a leaked/lost measurement outcome (symbol / reject) is sampling + sidecar, so it lands in the orchestrator (PR G), not here.
- **Ambiguous leaked/lost ops reject by default.** The policy function covers the explicit table cells (single-qubit gate/noise on Lost → drop; reset restores per `reset_restores_lost`; measurements kept) and rejects everything else (two-qubit gates, multi-qubit ops, feedback onto a vacated site). No new override knobs — deferred.

## Why R-insertion + policy ship together
The inserted `R` resets the residual SVM qubit to `|0>`; that is only sound because downstream policy never lets a later op treat that residual as a real computational atom (lost single-qubit gates drop, two-qubit ops reject).

## Tests (`test_noncomp_rewriter.cc`, 11 cases)
Driven from real sampler histories. Cover: `|1>` prep emitted / `|0>` not; the post-op carrier trace-out rule (incl. the `Known`-demoted-then-lost case and the still-`Known`-lost no-`R` case); **`R` survives the default compile pipeline as exactly one hidden measurement**; **visible measurement + detector indices unchanged with detectors before and after the loss point**; two-qubit-on-lost and single-qubit-on-leaked **reject**; single-qubit-on-lost **drops**; lost-qubit reset rejects by default and restores under policy; measurement-on-lost kept. Full Debug and Release suites green (888 cases).

## Next (PR G)
`orchestrator.{h,cc}` + bindings: drive sampler → rewriter → compile → SVM, apply the classifier to leaked/lost measurement outcomes, and assemble the sidecar. Seed handling per the separation note (entropy by default; domain-separate or thread one RNG when an explicit seed is supplied).

---
_Review follow-up (`d6ff1ae`): strengthened the record-invariance test to assert `hir.detector_targets` (not just the count) is unchanged, and documented `rewrite()`'s parser-normalized-circuit precondition. The classifier-vs-detectors point from review is a PR G design item (the classifier-sampled bit must be injected in-circuit before the SVM, not postprocessed) and is captured for that PR; it does not change PR F._

### DIFF
--- a/src/clifft/noncomp/rewriter.cc
+++ b/src/clifft/noncomp/rewriter.cc
@@ -18,20 +18,6 @@ namespace {
 // How the base operation is handled for one qubit operand.
 enum class OpAction { Apply, Drop, Reject };
 
-const char* kind_name(QubitStatusKind k) {
-    switch (k) {
-        case QubitStatusKind::ComputationalUnknown:
-            return "ComputationalUnknown";
-        case QubitStatusKind::ComputationalKnown:
-            return "ComputationalKnown";
-        case QubitStatusKind::Leaked:
-            return "Leaked";
-        case QubitStatusKind::Lost:
-            return "Lost";
-    }
-    return "unknown";
-}
-
 // Trajectory policy for one operand, keyed on the operand's status at op
 // entry. A computational qubit (known or unknown) runs every operation; the
 // table below only governs leaked and lost operands. Aggregated across an
@@ -95,14 +81,12 @@ Circuit rewrite(const Circuit& original, const NonComputationalHistory& history,
                                     std::to_string(original.num_qubits) + " qubits");
     }
 
-    Circuit out;
-    out.num_qubits = original.num_qubits;
-    // Only X-prep and trace-out R ops are inserted, neither of which is a
-    // visible measurement, so the record-layout counts carry over unchanged.
-    out.num_measurements = original.num_measurements;
-    out.num_detectors = original.num_detectors;
-    out.num_observables = original.num_observables;
-    out.num_exp_vals = original.num_exp_vals;
+    // Copy so every circuit-level field carries over (and any field added to
+    // Circuit later is not silently dropped); only the node list is rebuilt.
+    // The inserted X-prep and trace-out R ops are not visible measurements, so
+    // the record-layout counts stay valid.
+    Circuit out = original;
+    out.nodes.clear();
     out.nodes.reserve(original.nodes.size() + original.num_qubits);
 
     std::vector<QubitStatus> status = history.initial_status;
--- a/src/clifft/noncomp/rewriter.cc
+++ b/src/clifft/noncomp/rewriter.cc
@@ -1,0 +1,206 @@
+#include "clifft/noncomp/rewriter.h"
+
+#include "clifft/circuit/gate_data.h"
+#include "clifft/circuit/target.h"
+#include "clifft/noncomp/op_role.h"
+#include "clifft/noncomp/status_step.h"
+#include "clifft/noncomp/transition_instrument.h"
+
+#include <cstdint>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace clifft {
+
+namespace {
+
+// How the base operation is handled for one qubit operand.
+enum class OpAction { Apply, Drop, Reject };
+
+const char* kind_name(QubitStatusKind k) {
+    switch (k) {
+        case QubitStatusKind::ComputationalUnknown:
+            return "ComputationalUnknown";
+        case QubitStatusKind::ComputationalKnown:
+            return "ComputationalKnown";
+        case QubitStatusKind::Leaked:
+            return "Leaked";
+        case QubitStatusKind::Lost:
+            return "Lost";
+    }
+    return "unknown";
+}
+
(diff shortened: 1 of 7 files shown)


src/clifft/noncomp/rewriter.cc
```diff
@@ -0,0 +1,206 @@
...
+#include <vector>
+
+namespace clifft {
+
+namespace {
+
+// How the base operation is handled for one qubit operand.
+enum class OpAction { Apply, Drop, Reject };
+
+const char* kind_name(QubitStatusKind k) {
```
bachase: Why does kind_name live here versus with the QubitStatusKind definition?

src/clifft/noncomp/rewriter.cc
```diff
@@ -0,0 +1,206 @@
...
+            return OpAction::Drop;
+        }
+        if (lost) {
+            return OpAction::Drop;
+        }
+        // A single-qubit gate on a leaked qubit rejects by default.
+        return OpAction::Reject;
+    }
+    // Anything else touching a leaked or lost operand -- a two-qubit gate,
+    // a multi-qubit measurement-class op, classical feedback onto a vacated
```
bachase, line 61: Confirming that later on, if we have more complex policies, such as adding decoherent noise to a two-qubit gate for the remaining qubit, we would plug that in here?

src/clifft/noncomp/rewriter.cc
```diff
@@ -0,0 +1,206 @@
...
+
+    Circuit out;
+    out.num_qubits = original.num_qubits;
+    // Only X-prep and trace-out R ops are inserted, neither of which is a
+    // visible measurement, so the record-layout counts carry over unchanged.
+    out.num_measurements = original.num_measurements;
+    out.num_detectors = original.num_detectors;
+    out.num_observables = original.num_observables;
+    out.num_exp_vals = original.num_exp_vals;
+    out.nodes.reserve(original.nodes.size() + original.num_qubits);
```
bachase, line 90: Should Circuit itself make a "clone" or similar function to ensure if any fields change they are properly copied?

src/clifft/noncomp/rewriter.cc
bachase: Good call — moved it to `qubit_status.h` next to the `QubitStatusKind` enum as an inline `kind_name(QubitStatusKind)`, so it lives with the type it names and is reusable. Done in ab3c3df.

src/clifft/noncomp/rewriter.cc
bachase: Yes — `operand_action` is the locus for that. One nuance: "apply decoherent noise to the surviving qubit" is not a keep/drop/reject, it emits a *different* op (a noise channel on the survivor), so it would be a **replace** action. Concretely that means adding an `OpAction::Replace` (or letting the per-operand step return substitute ops for the loop to emit) plus a knob on `NonComputationalPolicy` to turn it on — the §5.2 table already earmarks these as "policy override" cells. So: right place, but it needs a replace path + a policy field, not just another branch here. Left as-is for this PR.

src/clifft/noncomp/rewriter.cc
bachase: Addressed in ab3c3df — `rewrite()` now does `Circuit out = original; out.nodes.clear();` and rebuilds only the node list, so every circuit-level field (including any added to `Circuit` later) carries over automatically. The implicitly-generated copy constructor already gives clone semantics, so I used that rather than adding a `Circuit::clone()`; happy to add a named `clone()` if you would rather it be explicit.

PR #116, opened 2026-06-03
docs: auto-generate Python API reference

### LINKED ISSUE #63: docs: auto-generate Python API reference via mkdocstrings
### Problem or motivation

The MkDocs site has a Reference section for gates, VM instructions, and optimization passes (auto-generated from data tables via `docs/macros.py`), but no auto-generated reference for the Python API. The public surface (`clifft.compile`, `clifft.sample`, `clifft.basis_probabilities`, `clifft.record_probabilities`, `Program`, `State`, etc.) is documented through prose in the User Guide, tests, and docstrings, but there is no single page that answers "what is available in `clifft`?"

This works at the current scale, but it drifts: signatures can change in the code without the docs catching up, and users have to piece the API together from examples.

### Proposed solution

Add a first Python API reference page using [`mkdocstrings`](https://mkdocstrings.github.io/) with the Python handler. `mkdocstrings` imports the package at build time and renders signatures plus docstrings for selected public symbols. This is a standard pairing with MkDocs Material and is the right shape for an API reference that should follow the implementation.

This issue is intended to produce a useful v1 reference page, not perfect prose documentation for every binding.

### Implementation sketch

1. Add `mkdocstrings[python]` to the `docs` dependency group in `pyproject.toml`.
2. Register the plugin in `mkdocs.yml` with sensible defaults (for example heading levels, signature display, member ordering, and cross-reference behavior).
3. Add `docs/reference/python-api.md` and organize the public API by topic using `::: clifft.<symbol>` directives.
4. Add the new page to the Reference nav in `mkdocs.yml`.
5. Update docs CI so the `clifft` package is installed/built before MkDocs imports it. This is required because the public package imports the nanobind C++ extension.

Suggested page organization:

- Compilation: `compile`, `parse`, `parse_file`, `trace`, `lower`
- Sampling: `sample`, `sample_survivors`, `sample_k`, `sample_k_survivors`
- Strong simulation: `basis_probabilities`, `record_probabilities`
- State inspection: `execute`, `get_statevector`
- Result types: `SampleResult`
- Compiled programs and execution state: `Program`, `State`
- Circuit and IR inspection: `Circuit`, `AstNode`, `Target`, `HirModule`, `HeisenbergOp`, `Instruction`
- Pass managers: `HirPassManager`, `BytecodePassManager`, `default_hir_pass_manager`, `default_bytecode_pass_manager`
- HIR passes: `PeepholeFusionPass`, `StatevectorSqueezePass`, `RemoveNoisePass`, `DropNonUnitaryPass`
- Bytecode passes: `NoiseBlockPass`, `ExpandTPass`, `ExpandRotPass`, `SwapMeasPass`, `MultiGatePass`, `SingleAxisFusionPass`
- Utilities: `get_num_threads`, `set_num_threads`, `svm_backend`, `version`

### CI requirements

This is in scope for the issue.

Today some docs workflows can run with only the docs dependency group because the docs do not import `clifft`. Once `mkdocstrings` is enabled, docs builds must be able to import the package. Update the workflows that build or deploy docs (for example PR preview, main docs deploy, and release docs deploy as applicable) so they install/build the package before running `mkdocs build` or `mike deploy`.

The exact command may depend on the existing workflow environment, but the important requirement is:

```bash
uv pip install -e .
uv run --group docs mkdocs build --strict
```

or the workflow-specific equivalent. Avoid leaving CI in a state where the new page works only on machines that already have `clifft` installed.

### Known gotchas to plan around

- **The C++ extension build becomes a docs prerequisite.** `mkdocstrings` imports `clifft`, and `clifft` imports `clifft._clifft_core`. CI must build/install the package before the docs command runs.
- **nanobind-bound classes render through runtime docstrings and introspection.** Top-level functions already have useful docstrings, but classes such as `Program`, `State`, and `Instruction` currently have terse one-line class docs and some minimally documented properties/methods. That is acceptable for this v1 reference page.
- **Do not backfill every binding docstring in this PR.** Improving C++ binding docstrings is good follow-up work, but it should not be required to land the reference scaffold.
- **Source links for bound symbols may be imperfect.** Symbols re-exported from `clifft._clifft_core` may point at the binary module or have no useful source link. Acceptable for v1; can be tuned later with mkdocstrings options if needed.

### Done when

- `docs/reference/python-api.md` exists and is included under **Reference → Python API**.
- The page renders signatures and docstrings for the public symbols listed above.
- `mkdocs build --strict` passes locally in a clean environment after installing the package and docs dependencies.
- Docs CI workflows that build/deploy the site install/build `clifft` before running MkDocs/Mike.
- A user landing on **Reference → Python API** can navigate the public Python surface without reading the User Guide first.

### Out of scope

- Fully rewriting or expanding all nanobind docstrings.
- Perfect source links for C++/nanobind symbols.
- Changing public APIs to make the docs prettier.
- Replacing the narrative User Guide pages.

### Alternatives considered

- **Hand-authoring per-symbol pages** (with or without `mkdocs-gen-files`). This drifts the same way the current prose-in-User-Guide approach does, but with more text to maintain. `mkdocstrings` is a better fit for "generate an API reference from the importable package."
- **Backfilling docstrings on every nanobind binding before adding the page.** Useful eventually, but it couples two pieces of work that do not need to land together. Better to land the reference scaffold first and let the rendered output guide focused docstring follow-ups.

an1k3sh:
## Summary

Configure mkdocstrings to auto-generate cleaner Python API reference documentation from pybind11 docstrings. Changes to `mkdocs.yml`:

- Add member filters to suppress internal Python attributes (`__doc__`,` __module__`, `__new__`, `__repr__`, `__init__` )
- Enable category grouping and symbol-type TOC labels
- Switch docstring parser to numpy style

## Test plan

## Contributor agreement

  Closes https://github.com/unitaryfoundation/clifft/issues/63


bachase:
Hi @an1k3sh. Thanks for your interest in contributing. I enabled CI workflows. Please address those failures. Once that CI is clean, I can do a deeper review of your PR.

pyproject.toml
bachase, line 1: Note that this changes the uv.lock file, which is now stale. Be sure to check in the updated version.

mkdocs.yml
```diff
@@ -69,6 +69,33 @@ plugins:
...
+          options:
+            show_source: false
+            show_root_heading: true
+            heading_level: 3
+            members_order: source
+            separate_signature: true
+            docstring_style: numpy
+            docstring_section_style: spacy
+            filters:
+              - "!^__doc__$"
```
bachase, line 84: How come doc and module are listed twice?

docs/reference/python-api.md
bachase, line 1: Do you have any good suggestions on keeping this up to date?

mkdocs.yml
bachase, line 84: Would it be simpler to filter in any __.*__ method versus enumerating these?

bachase (changes requested):
Good start, however there are still some gaps in the build I think


`docs.yml`, the release docs steps, and the `just docs-build` / `just docs-serve` recipes still use `uv run --only-group docs`, which no longer installs/imports `clifft` and will fail on `reference/python-api.md`. Please update those to install the package and include the docs group, matching the PR preview/CI approach.

Also, the pr_preview was previously only running on changes to docs, but now it might change with changes to these bindings. Can you think of a good filter to add to cover thoes cases?

bachase:
Thanks for your work @an1k3sh but I've accepted #121 for this issue.

PR #118, opened 2026-06-04
feat(playground): add starter circuit catalog

### LINKED ISSUE #36: Add starter circuit catalog to the playground
## Problem or motivation

The playground is one of the first surfaces new users encounter, but it currently opens without an obvious starter-circuit catalog. Users can paste code, load a URL, or use saved circuits, but there is no built-in way to pick a small known-good circuit and immediately explore the compiler and simulator behavior.

This would make the playground more useful for onboarding, demos, and quick smoke checks.

## Proposed solution

Add a small starter-circuit catalog to the playground UI.

Suggested shape:

- Add a small typed registry of example circuits, with metadata such as title, short description, and source text or source URL.
- Add a compact examples/presets control to the playground toolbar or another appropriate existing UI surface.
- Loading an example should update the editor through the same path as other source loads where practical.
- Preserve share/deep-link behavior for loaded examples, either by encoding the source or by using the existing `?url=` style when appropriate.

Open design questions for the implementer (please propose your suggestions as a comment before starting work):

- Which circuits should be included initially? Possible candidates include Bell, GHZ, a small T-gate/non-Clifford circuit, a tiny detector/QEC-flavored circuit, and one larger hosted example.
- Should examples be embedded and shipped directly with the playground bundle, or should larger examples load from GitHub/raw URLs through the existing CORS-compatible URL-loading path?

Suggested acceptance criteria:

- A user can select at least a few starter circuits from the playground UI.
- Starter circuits compile/run in the playground.
- The implementation works in both local Vite development and GitHub Pages deployment.
- Add or update lightweight tests where practical for example registry/deep-link behavior.

## Alternatives considered

Relying only on docs examples keeps the playground simpler, but it misses the chance to make the interactive page self-contained for first-time users. Another option is to add examples only to docs and deep-link into the playground; that may still be useful, but does not solve the blank-playground onboarding gap by itself.

Samfresh-ai:
## Summary

Closes #36.

Adds a compact starter-circuit catalog to the playground toolbar. The examples
load through the existing `onLoadCircuit` path, so Share, Save, and Recents stay
on the same source flow as manually loaded circuits.

The current starter set includes Bell pair and GHZ baseline Clifford examples,
plus T-gate interference and arbitrary-angle `R_ZZ(0.25)` examples that show
non-Clifford behavior in Clifft.

## Test plan

The checklist commands above are not a clean fit for this playground TypeScript
UI change, so I left them unchecked unless fully proven. Two bounded
`uv run pre-commit run --all-files` attempts timed out during hook setup/mypy;
the second run had already passed clang-format, ruff, and ruff-format before
the timeout.

Playground checks that passed:

- `npm run lint`
- `npm run build`
- `git diff --check`

## Contributor agreement

### DIFF
--- a/playground/src/examples.ts
+++ b/playground/src/examples.ts
@@ -27,8 +27,8 @@ M 0 1 2
   },
   {
     id: "t-gate-probability",
-    title: "T-gate probability",
-    description: "Small non-Clifford circuit that activates Clifft's VM.",
+    title: "T-gate interference",
+    description: "T between Hadamards changes the final Z-basis probability.",
     source: `H 0
 T 0
 H 0
@@ -36,13 +36,12 @@ M 0
 `,
   },
   {
-    id: "noisy-bell",
-    title: "Noisy Bell",
-    description: "Bell circuit with depolarizing noise before measurement.",
+    id: "zz-rotation",
+    title: "ZZ rotation",
+    description: "Arbitrary-angle two-qubit rotation outside the Clifford gate set.",
     source: `H 0
-DEPOLARIZE1(0.01) 0
-CNOT 0 1
-DEPOLARIZE2(0.01) 0 1
+H 1
+R_ZZ(0.25) 0 1
 M 0 1
 `,
   },
--- a/playground/src/examples.ts
+++ b/playground/src/examples.ts
@@ -1,0 +1,49 @@
+export interface StarterCircuit {
+  id: string;
+  title: string;
+  description: string;
+  source: string;
+}
+
+export const STARTER_CIRCUITS: StarterCircuit[] = [
+  {
+    id: "bell-pair",
+    title: "Bell pair",
+    description: "Two-qubit entanglement with correlated Z measurements.",
+    source: `H 0
+CNOT 0 1
+M 0 1
+`,
+  },
+  {
+    id: "ghz-state",
+    title: "GHZ state",
+    description: "Three-qubit fanout circuit with shared parity.",
+    source: `H 0
+CNOT 0 1
+CNOT 1 2
+M 0 1 2
+`,
+  },
+  {
+    id: "t-gate-probability",
+    title: "T-gate probability",
+    description: "Small non-Clifford circuit that activates Clifft's VM.",
+    source: `H 0
+T 0
+H 0
+M 0
+`,
+  },
+  {
+    id: "noisy-bell",
+    title: "Noisy Bell",
+    description: "Bell circuit with depolarizing noise before measurement.",
+    source: `H 0
+DEPOLARIZE1(0.01) 0
+CNOT 0 1
+DEPOLARIZE2(0.01) 0 1
+M 0 1
+`,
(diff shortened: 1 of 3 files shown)


playground/src/examples.ts
```diff
@@ -0,0 +1,49 @@
...
+    source: `H 0
+CNOT 0 1
+CNOT 1 2
+M 0 1 2
+`,
+  },
+  {
+    id: "t-gate-probability",
+    title: "T-gate probability",
+    description: "Small non-Clifford circuit that activates Clifft's VM.",
```
bachase: Can you explain what you mean by "activates Clifft's VM"?

playground/src/examples.ts
```diff
@@ -0,0 +1,49 @@
...
+    source: `H 0
+T 0
+H 0
+M 0
+`,
+  },
+  {
+    id: "noisy-bell",
+    title: "Noisy Bell",
+    description: "Bell circuit with depolarizing noise before measurement.",
```
bachase: I don't think this adds much alongside the non-noisy bell circuit. I would just include this one or the other one.

bachase:
Hi @Samfresh-ai  it looks like you have more than the limit of 4 PRs open for unitary hack at this time, so I won't be reviewing this for now

Samfresh-ai:
I reduced my open GitHub unitaryHACK PR queue to four and kept this PR in the active queue. The closed branches are preserved for later, but this one should now be under the GitHub-side 4-open-PR limit.

bachase (commented):
Thanks @Samfresh-ai. This is looking like a great start. I like the style and approach.

Before merging, it would be nice to have at two examples that more clearly demonstrate the non-clifford capabilities of clifft. Do you have any ideas for ones to use?

Samfresh-ai:
Thanks Brad, addressed in a93ed95.

Changes:
- replaced the redundant noisy Bell starter with an `R_ZZ(0.25)` two-qubit rotation example
- renamed the T example to `T-gate interference` and replaced the VM wording with a measurement-bias description
- kept the Clifford Bell/GHZ starters as baseline examples, with two non-Clifford starters now in the catalog

Checks:
- `npm run build`
- `npm run lint`
- `git diff --check`

bachase:
looks great!

Before I do a squash merge,  can you update the PR description to follow our default PR template. In particular, checking of the tick boxes.

Also if you used AI:, when merging, we will do a squash merge and I will want to add an assisted-by trailer with the name of the AI agent you used, so please either update your commits to have that trailer or just share in github which agent you used.

Samfresh-ai:
Updated the PR description to the default template and checked the contributor
agreement boxes.

I also amended the latest commit to include:

`Assisted-by: OpenAI Codex`

I left the full pre-commit box unchecked because two bounded
`uv run pre-commit run --all-files` attempts timed out during hook setup/mypy.
On the second run, clang-format, ruff, and ruff-format had already passed.

PR #121, opened 2026-06-04
docs: auto-generate Python API reference via mkdocstrings

### LINKED ISSUE #63: docs: auto-generate Python API reference via mkdocstrings

manasa-manoj-nbr:
## Summary

Closes #63
Adds a first Python API reference page auto-generated from docstrings
and runtime signatures using `mkdocstrings[python]`.
- Adds `mkdocstrings[python]>=0.26` to the `docs` dependency group in `pyproject.toml`
- Registers the mkdocstrings plugin in `mkdocs.yml` with Google docstring style, full signature annotations, dunder member filtering, and `merge_init_into_class`
- Adds `docs/reference/python-api.md` organized into 12 sections matching the issue's suggested layout (compilation, sampling, strong simulation, state inspection, result types, programs, IR inspection, pass managers, HIR passes, bytecode passes, utilities, type aliases)
- Adds `docs/hooks.py` — a MkDocs hook that suppresses griffe's nanobind type-annotation warnings so `mkdocs build --strict` passes cleanly
- Updates `justfile` `docs-serve` and `docs-build` recipes from `--only-group docs` to `--group docs`
- Updates all three doc CI workflows (`docs.yml`, `pr_preview.yml`, `release.yml`) to run `uv pip install -e .` before any mkdocs/mike command, since mkdocstrings now imports `clifft` (and therefore `clifft._clifft_core`) at build time
- Adds `src/python/**` to the `pr_preview` path trigger so the preview rebuilds when Python bindings change

## Test plan

- [x] `mkdocs build --strict` passes locally with zero warnings
- [x] `mkdocs serve` — Reference -> Python API page renders with correct signatures, docstrings, and section groupings
- [x] Existing Reference pages (Supported Gates, Instruction Reference, Optimization Passes) unaffected
- [x] `uv.lock` updated

## Contributor agreement

## AI Disclosure
I used Claude to help implement and debug this PR : reviewing the repo structure, identifying the right mkdocstrings options, tracing the MkDocs logging chain to fix the strict-mode warning suppression, and diagnosing the CI deployment failures. I reviewed all generated changes, verified the build locally, and confirmed the rendered output before submitting.

### DIFF
--- a/.github/workflows/pr_preview.yml
+++ b/.github/workflows/pr_preview.yml
@@ -53,7 +53,7 @@ jobs:
           enable-cache: true
 
       - name: Install clifft and docs deps
-        run: uv sync --group docs
+        run: uv sync --frozen --group docs
 
       - name: Build docs
         run: uv run --group docs mkdocs build --strict
--- a/.github/workflows/pr_preview.yml
+++ b/.github/workflows/pr_preview.yml
@@ -52,8 +52,8 @@ jobs:
         with:
           enable-cache: true
 
-      - name: Install clifft
-        run: uv pip install --system -e .
+      - name: Install clifft and docs deps
+        run: uv sync --group docs
 
       - name: Build docs
         run: uv run --group docs mkdocs build --strict
--- a/.github/workflows/pr_preview.yml
+++ b/.github/workflows/pr_preview.yml
@@ -8,6 +8,7 @@ on:
       - 'docs/**'
       - 'playground/**'
       - 'src/wasm/**'
+      - 'src/python/**'
       - 'mkdocs.yml'
       - '.github/workflows/pr_preview.yml'
 
@@ -46,14 +47,16 @@ jobs:
           cd playground && npm ci
           VITE_BASE_PATH=/clifft/pr-preview/pr-${{ github.event.number }}/playground/ VITE_BUILD_SHA=${{ github.sha }} npm run build
 
-      # --- Build MkDocs (docs group only, no C++ extension needed) ---
       - name: Install uv
         uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
 
+      - name: Install clifft
+        run: uv pip install -e .
+
       - name: Build docs
-        run: uv run --only-group docs mkdocs build --strict
+        run: uv run --group docs mkdocs build --strict
         env:
           SITE_URL: https://unitaryfoundation.github.io/clifft/pr-preview/pr-${{ github.event.number }}/
 
--- a/.github/workflows/pr_preview.yml
+++ b/.github/workflows/pr_preview.yml
@@ -53,7 +53,7 @@ jobs:
           enable-cache: true
 
       - name: Install clifft
-        run: uv pip install -e .
+        run: uv pip install --system -e .
 
       - name: Build docs
         run: uv run --group docs mkdocs build --strict
--- a/docs/hooks.py
+++ b/docs/hooks.py
@@ -1,0 +1,16 @@
+import logging
+
+
+class _SuppressGriffeNanobindWarnings(logging.Filter):
+    # nanobind-bound parameters lack Python type annotations; griffe warns
+    # about this when inspecting the compiled extension. These are not real
+    # doc errors, so drop them so mkdocs build --strict passes.
+    def filter(self, record: logging.LogRecord) -> bool:
+        return "No type or annotation for parameter" not in record.getMessage()
+
+
+def on_config(config):
+    filt = _SuppressGriffeNanobindWarnings()
+    for handler in logging.getLogger("mkdocs").handlers:
+        handler.addFilter(filt)
+    return config
--- a/docs/hooks.py
+++ b/docs/hooks.py
@@ -10,7 +10,7 @@ class _SuppressNanobindAnnotationWarnings(logging.Filter):
     def filter(self, record: logging.LogRecord) -> bool:
         if record.levelno > logging.WARNING:
             return True
-        if not record.name.startswith(("griffe", "mkdocs.plugins.mkdocstrings")):
+        if not record.name.startswith(("griffe", "mkdocs.plugins.griffe", "mkdocs.plugins.mkdocstrings")):
             return True
         return "No type or annotation for parameter" not in record.getMessage()
 
--- a/docs/hooks.py
+++ b/docs/hooks.py
@@ -10,7 +10,9 @@ class _SuppressNanobindAnnotationWarnings(logging.Filter):
     def filter(self, record: logging.LogRecord) -> bool:
         if record.levelno > logging.WARNING:
             return True
-        if not record.name.startswith(("griffe", "mkdocs.plugins.griffe", "mkdocs.plugins.mkdocstrings")):
+        if not record.name.startswith(
+            ("griffe", "mkdocs.plugins.griffe", "mkdocs.plugins.mkdocstrings")
+        ):
             return True
         return "No type or annotation for parameter" not in record.getMessage()
 
--- a/docs/hooks.py
+++ b/docs/hooks.py
@@ -1,16 +1,22 @@
 import logging
 
 
-class _SuppressGriffeNanobindWarnings(logging.Filter):
+class _SuppressNanobindAnnotationWarnings(logging.Filter):
     # nanobind-bound parameters lack Python type annotations; griffe warns
     # about this when inspecting the compiled extension. These are not real
     # doc errors, so drop them so mkdocs build --strict passes.
+    # Scoped to WARNING level and griffe/mkdocstrings origins only so that
+    # ERROR-level records and unrelated loggers are never silenced.
     def filter(self, record: logging.LogRecord) -> bool:
+        if record.levelno > logging.WARNING:
+            return True
+        if not record.name.startswith(("griffe", "mkdocs.plugins.mkdocstrings")):
+            return True
         return "No type or annotation for parameter" not in record.getMessage()
 
 
 def on_config(config):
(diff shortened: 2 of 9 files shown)


.github/workflows/pr_preview.yml
```diff
@@ -46,14 +47,16 @@ jobs:
...
         uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
 
+      - name: Install clifft and docs deps
+        run: uv sync --group docs
+
       - name: Build docs
-        run: uv run --only-group docs mkdocs build --strict
+        run: uv run --group docs mkdocs build --strict
```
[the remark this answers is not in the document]
manasa-manoj-nbr: ubuntu-24.04 has gcc, cmake, and ninja pre-installed, so the C++ extension builds natively without any extra toolchain setup. The workflow already runs a heavier Docker-based WASM build, so this adds negligible overhead.

mkdocs.yml
```diff
@@ -63,12 +63,29 @@ extra:
...
+            docstring_style: google
+            docstring_options:
+              warn_unknown_params: false
+            show_root_heading: true
+            show_symbol_type_heading: true
+            show_symbol_type_toc: true
+            show_signature_annotations: true
+            merge_init_into_class: true
+            filters:
+              - "!^__"
```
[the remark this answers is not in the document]
manasa-manoj-nbr: CI uses uv sync --frozen so exact versions are already pinned by the lockfile. Tightened the lower bound from >=0.26 to >=1.0 to reflect the version these options were verified against. No upper bound added since the lockfile handles reproducibility for CI.

docs/hooks.py
```diff
@@ -0,0 +1,16 @@
+import logging
+
+
+class _SuppressGriffeNanobindWarnings(logging.Filter):
```
bachase: What does Griffe mean in this class name?

docs/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+    # nanobind-bound parameters lack Python type annotations; griffe warns
+    # about this when inspecting the compiled extension. These are not real
+    # doc errors, so drop them so mkdocs build --strict passes.
+    def filter(self, record: logging.LogRecord) -> bool:
+        return "No type or annotation for parameter" not in record.getMessage()
```
bachase, line 17: Asking to understand, do you think this would ever suppress an error we do care about seeing?

bachase (commented):
Looking really good -- just a few questions to be sure I understand some choices.

docs/hooks.py
manasa-manoj-nbr: `Griffe` refers to the static analysis library used by mkdocstrings to inspect Python packages when generating API documentation. During documentation builds, Griffe loads `clifft._clifft_core` (the nanobind-compiled extension) and emits warnings about parameters that lack Python type annotations. These warnings are expected for nanobind-generated bindings and are not actual documentation issues, so this filter suppresses them to avoid noisy output during `mkdocs build --strict`.  The class name refers to that specific source.

That said, happy to rename it to something more self-explanatory  `_SuppressNanobindAnnotationWarnings` , since the underlying reason for the warnings is nanobind-generated bindings rather than Griffe itself.

bachase: Got it; yes I think that would be a clearer name.

docs/hooks.py
manasa-manoj-nbr: Good catch. The current filter is broader than necessary , it would also suppress ERROR-level records or the same message text from unrelated loggers. Hardening it to only suppress WARNING-level records originating from Griffe/mkdocstrings makes the intent explicit and prevents it from silently hiding real errors. Updating now.

bachase (commented):
Great - thanks for making these changes.

One list style nit if you don't mind. For me, the function signatures font size is a bit larger relative to other aspects. Is there a way to change the CSS/generated content so it fits in better with the rest of the page?

Image

manasa-manoj-nbr:
> Great - thanks for making these changes.
> 
> One list style nit if you don't mind. For me, the function signatures font size is a bit larger relative to other aspects. Is there a way to change the CSS/generated content so it fits in better with the rest of the page?
> 
> Image

The large font was because mkdocstrings renders the full signature inside the heading by default, so it inherited the heading size. I enabled `separate_signature: true` , which keeps just the symbol name in the heading and renders the full signature as a standalone code block at the normal code font size so it now matches the rest of the page. No custom CSS needed.
I've included this change in the latest PR. Please let me know if anything is unclear or if there are any additional changes you'd like me to make. Thank you!
image

bachase:
Fantastic work -- thanks @manasa-manoj-nbr !

manasa-manoj-nbr:
> Fantastic work -- thanks @manasa-manoj-nbr !

Thank You!!

Issue #40, opened 2026-04-30
Explore HIR-level global T-count optimization


bachase:
## Problem or motivation

Clifft currently has local HIR optimizations, such as adjacent or commuting T/T-dagger cancellation and fusion, but it does not have a global T-count optimization pass.

T gates are an important resource in fault-tolerant quantum computing, and there is a substantial literature on global T-count reduction techniques. Clifft's HIR represents non-Clifford operations as Pauli-string operations, which may be a useful substrate for some of these techniques.

We would like to understand whether an HIR-level optimization can reduce T count on representative circuits without requiring changes to the HIR, parser, bytecode, or VM.

## Proposed solution

Prototype and evaluate one HIR-level approach to T-count reduction.

The implementation approach is intentionally open. A contributor should pick a tractable technique or subset of a technique, explain why it fits Clifft's current HIR, and demonstrate whether it reduces T count on small representative examples.

The goal is a small prototype or draft implementation plus enough evaluation to decide whether this direction is worth productionizing.

Suggested direction:

- Work within the existing `HirPass`/`HirModule` representation.
- Focus on `T_GATE` operations and any directly relevant Clifford-frame/HIR behavior.
- Keep the pass opt-in if implemented as a pass.
- Measure T count before and after on a small set of examples.
- Validate semantic equivalence on small circuits using statevector comparison or another existing test oracle.

## Constraints

- Do not redesign HIR data structures.
- Do not change the parser, bytecode, or VM.
- Do not add the prototype to the default optimization pipeline.
- Do not add new third-party C++ dependencies without prior maintainer discussion.
- Do not optimize unrelated metrics such as CX count or depth in this issue.

## Acceptance criteria

- The PR explains the chosen approach, cites prior work where applicable, and explains why the approach can be expressed correctly in current HIR.
- The implementation is opt-in or clearly marked experimental.
- The evaluation includes a before/after T-count table on representative circuits, not only minimal constructed rewrite examples. Note we are seeking a scientific level of rigor here -- not just showing a few handcrafted/toy examples. You should show how this pass works on real-world circuits as a way to evaluate whether its worth adding.
- If multiple optimization phases are present, the evaluation reports each phase’s contribution separately.
- Semantic equivalence is checked for evaluated circuits where the new pass actually changes the program.
- Existing tests, new tests, and pre-commit checks pass before maintainer review is requested.
[...]

PR #123, opened 2026-06-06
feat: Add experimental MCR T Count Pass

EthanFeld:
## Summary
For #40 
This PR adds an opt-in experimental HIR pass, `ExperimentalMcrTCountPass`, to prototype global T-count reduction without changing the parser, HIR layout, bytecode, or VM.

Here, **MCR** means **multiproduct commutator relation**. In this context, it is a four-Pauli-rotation identity where two commuting pairs can be reordered as a unit because their Pauli axes satisfy a specific commutation / anti-commutation pattern. That reorder can expose adjacent same-axis T gates that then cancel or fuse through Clifft's existing local peephole machinery.

This is based on Mori, Yusei, et al. “Nontrivial Multiproduct Commutation Relation toward Reducing T-Count in Sequential Pauli-Based Computation.” PRX Quantum, edited by , vol. 7, no. 2, June 2026. Crossref, https://doi.org/10.1103/qw5z-3bwz. but it is more of a approximation of the algorithm the paper actually uses due to the HIR constraints.

The current implementation is a lot slower than I would like, there is a clean math way to check the safety of a swap but implementing it has proved challenging.

## Theory

The main idea is that the current peephole pass cannot merge through anticommuting axis, MCR is essentially just a rule for swapping through these axis. This exposes optimizations that would otherwise be hidden. The paper its based on does a layer structure with this. The operations required for that layer structure are not possible within the constraints of the issue so I went with a more simple approach.

## Approach

The pass works inside barrier-bounded windows of contiguous T rotations:
`collect_windows` cuts the op stream at any non-rotation op (measurement,
conditional Pauli, phase rotation, noise, …) and keeps each maximal run with at
least four T gates.

Within a window it slides an anchor across a bounded lookahead horizon — capped
in both T-gate count and raw op distance — and enumerates distinct-axis
quadruples `A, B, C, D`. Such a quadruple is an MCR candidate when the two
commuting pairs `(A, B)` and `(C, D)` can be exchanged as units even though
every cross pair (`A`–`C`, `A`–`D`, `B`–`C`, `B`–`D`) anticommutes: each pair
collapses to a single exponential and the two sums commute, so the block swap
`(A, B)(C, D) → (C, D)(A, B)` preserves the circuit despite the gates not
commuting pairwise.

Rather than test that algebraic predicate, the pass certifies each candidate by
construction. `exact_span_swap_rewrite_is_valid` expands the entire anchor-to-`D`
span into a symbolic Pauli operator before and after the swap and accepts the
candidate only if the two agree up to global phase. This validates the swap
against the real span — intervening rotations included — not the four axes in
isolation, and results are memoized per span signature. A cheap
merge-reachability pre-filter (`window_swap_has_merge_potential`) discards
candidates that wouldn't bring a same-axis pair within fusion reach before the
symbolic check runs.

The swap is T-count-neutral, so it is never applied speculatively. The four ops
are permuted in place (`(A, B, C, D) → (C, D, A, B)`, leaving the intervening
gates untouched), the candidate is applied to a trial copy, the existing
`PeepholeFusionPass` is run, and the rewrite is kept only if the T-count
strictly decreases. The pass iterates to a fixed point.

## Why MCR fits current HIR

Clifft's HIR already represents non-Clifford quarter turns as Pauli-string operations. That makes MCR a good fit because the prototype only needs:

- Pauli mask commutation / anti-commutation checks
- Local rewrites over `HirModule`
- Existing virtual-S downstream absorption for Clifford fusion
- Existing peephole cleanup after rewrite

No HIR data structure changes were needed.

Other T-count optimization methods were explored(Path-Sum, ZX calculus, ZH calculus) but, the HIR does not pass enough information to use these efficiently.

## What changed

- Added `ExperimentalMcrTCountPass` as a disabled-by-default HIR pass.
- Registered the pass in the pass registry.
- Added C++ tests for:
  - successful MCR-driven reductions
  - barrier-separated windows
  - non-MCR quadruples
  - blocked-gather regressions
  - mixed-direction rejection
  - late-window anchor search
- Added evaluation coverage for:
  - statevector equivalence on small unitary examples
  - measurement-distribution equivalence on a measured example
  - before/after T-count examples
  - confirmation that the pass is not in the default pipeline
- Added an evaluation script to replicate results

## Evaluation

Using `uv run python docs/guide/scripts/mcr_t_count_evaluation.py`:

| example | source T-like ops | after peephole | after ExperimentalMcrTCountPass + peephole | window scans | over-cap scans | quadruples | swaps | merges | T removed |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| toggle_sandwich | 6 | 6 | 2 | 1 | 0 | 1 | 1 | 2 | 4 |
| kicked_xy_block | 5 | 5 | 3 | 1 | 0 | 1 | 1 | 1 | 2 |
| negative_sign_block | 5 | 5 | 3 | 1 | 0 | 1 | 1 | 1 | 2 |
| late_window_block | 23 | 23 | 21 | 2 | 2 | 1 | 1 | 1 | 2 |
| two_disjoint_pair_blocks | 10 | 10 | 6 | 3 | 0 | 2 | 2 | 2 | 4 |
| three_disjoint_pair_blocks | 15 | 15 | 9 | 4 | 0 | 3 | 3 | 3 | 6 |
| two_disjoint_pairs_x2 | 20 | 16 | 12 | 3 | 0 | 2 | 2 | 2 | 4 |
| fermionic_swap_net_6_l3 | 40 | 40 | 24 | 9 | 9 | 8 | 8 | 8 | 16 |
| fermionic_swap_net_8_l3 | 55 | 55 | 33 | 12 | 12 | 11 | 11 | 11 | 22 |
| fermionic_swap_net_onsite_6_l3 | 58 | 54 | 38 | 9 | 9 | 8 | 8 | 8 | 16 |
| fermionic_swap_net_hubbard_6_l3 | 66 | 62 | 18 | 9 | 9 | 8 | 8 | 22 | 44 |
| star_hub_4_l1 | 15 | 15 | 9 | 4 | 0 | 3 | 3 | 3 | 6 |
| bell_pumping_r2 | 24 | 24 | 16 | 8 | 0 | 4 | 4 | 4 | 8 |
| inject_entangle_measure_r2 | 30 | 30 | 18 | 11 | 0 | 6 | 6 | 6 | 12 |
| inject_bell_cultivate_r2 | 32 | 32 | 20 | 11 | 0 | 6 | 6 | 6 | 12 |

Using `uv run python docs/guide/scripts/mcr_t_count_corpus_evaluation.py`, the broader corpus improved `10 / 23` entries. Cases with no benefit were:

| family | circuit | source T-like ops | after peephole | after ExperimentalMcrTCountPass + peephole | delta T | reduction |
|---|---|---:|---:|---:|---:|---:|
| ripple_adder | full_3 | 42 | 24 | 24 | 0 | 0.0% |
| ripple_adder | full_4 | 56 | 32 | 32 | 0 | 0.0% |
| comparator | n4_v7 | 37 | 23 | 23 | 0 | 0.0% |
| comparator | n5_v13 | 51 | 31 | 31 | 0 | 0.0% |
| weighted_adder | n3_weights_1_2_3 | 156 | 110 | 110 | 0 | 0.0% |
| mcx | controls_4 | 27 | 23 | 23 | 0 | 0.0% |
| mcx | controls_5 | 39 | 31 | 31 | 0 | 0.0% |
| hubbard_chain | n4_l2 | 38 | 38 | 38 | 0 | 0.0% |
| hubbard_chain | n6_l2 | 62 | 62 | 62 | 0 | 0.0% |
| qaoa_ring | n6_p2 | 24 | 24 | 24 | 0 | 0.0% |
| qaoa_ring | n8_p2 | 32 | 32 | 32 | 0 | 0.0% |
| double_excitation | reps_1 | 8 | 8 | 8 | 0 | 0.0% |
| double_excitation | reps_2 | 16 | 0 | 0 | 0 | 0.0% |

These results show that the bounded MCR search is not universally useful: it helps when the circuit contains reorderable T windows that can expose same-axis cancellations after commuting-style rewrites, and it does not help on several arithmetic, MCX, QAOA-ring, Hubbard-chain, and compact excitation examples where peephole already saturates or the needed MCR structure is absent. In the cases where it does help, including swap-network and entanglement/distillation-style workloads such as star-hub entanglers, Bell pumping, inject-entangle-measure, and inject-bell-cultivate, every accepted rewrite is still gated by an exact local equivalence check and a post-peephole T-count reduction.
## Validation

Checked with:

- new C++ tests for pass behavior and registry exposure
- T-count reduction examples
- statevector equivalence on small unitary examples
- measurement-record probability equivalence on a measured example
- successful build with the new pass compiled in

## Conclusion

This direction looks promising for structured Pauli-rotation workloads. MCR appears expressible in the current HIR and can reduce T count without changes below the optimizer layer.

This could be further improved through employing a larger window, though this would require a improved heuristic.

The evaluation script is not meant to actually go in the production repo, it is just left so that results can be reproduced

## AI Disclosure
Codex was utilized to test, prototype and document the code. All code was human reviewed.


EthanFeld:
Sorry about the error before, it should work now

bachase:
Thanks for this @EthanFeld .

Some overall comments before a deeper review (that I'm asking of all the PRs for this issue).

1. Can you confirm if MCR is a novel approach? If not, please be sure to cite any prior work.
2. Can you add a brief "theory" type section that explains MCR in mathematical terms and why its a valid optimization pass?
3. I'm glad to see the smaller MCR-shaped examples, but for this bounty/issue, I'm also looking for a stronger scientific level of study to accept a contribution. Can you evaluate on a broader set of circuits/more real world ones? It will be useful to know where it works well and where it might not.

EthanFeld:
@bachase I added the theory section, citation and expanded evidence you asked for.

EthanFeld:
@bachase so sorry about that, I fixed the error

bachase (commented):
@EthanFeld thanks for these updates, it makes it much clearer on the work so far and how it relates to broader circuits. I'm still focused on high-level feedback versus any lower level implementation suggestions

1. It sounds like this implements a subset of the techniques in the original paper. Were the other approaches infeasible due to clifft architecture/HIR? Do you know how much those are responsible for the results in the paper?
2. You also mention the runtime concern. Have you done any benchmarks/see any opportunities to improve or is that fundamental to how the MCR approach works?

EthanFeld:
@bachase 
1. The main difference between this and the paper is finding the candidates to optimize. The layer technique the paper uses is theoretically faster but it requires doing a Pauli sum that Clifft cannot do or store without a significant change to the HIR or adding a ton of bolted on structure. The compiler that the paper is about is built intentionally for this so it gives a significant benefit there. In practice, it probably finds a few more of the harder optimizations but, the bulk of the value seems to be retained with my method.
2. I just fixed the runtime concern, before it was evaluating way too many potential candidates so I added more pruning and that seems to have fixed it. Now runtime is very low(<.2 seconds for anything I tried).

docs/guide/scripts/run_benchmark.py
bachase, line 1: Can you share more on why these changes are included? They seem unrelated to your work.

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+namespace clifft {
+
+namespace {
+
+// =========================================================================
+// Windowed bounded MCR search
+// =========================================================================
+
+/// Conjugate Pauli Q by S_P in place.
+inline void conjugate_pauli_by_S(MaskView x_p, MaskView z_p, bool sign_p, MutableMaskView x_q,
```
bachase, line 29: It looks like a good bit of these initial functions are copied from `peephole.cc`, but never used. Should you just remove them?

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+        double mag = std::min(mag_lhs, mag_rhs);
+        if (mag <= std::max(best_mag, kEps))
+            continue;
+        phase = coeff_lhs / it->second;
+        best_mag = mag;
+    }
+
+    if (best_mag <= kEps)
+        return false;
+    phase /= std::abs(phase);
```
bachase, line 430: What is this phase represent? Does it need to be pulled into the overall global phase after the pass runs?

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+        }
+
+        terms = std::move(next_terms);
+    }
+
+    return terms;
+}
+
+bool equal_up_to_global_phase(const SymbolicUnitary& lhs, const SymbolicUnitary& rhs) {
+    constexpr double kEps = 1e-9;
```
bachase, line 411: How is this epsilon determined? Does it depend on the window/span you've selected? Wondering how we would detect if this silently declares equality when we are looking at longer sequences that might have smaller absolute floating point diffs.

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+        auto src_b = hir.source_map[cand.b];
+        auto src_c = hir.source_map[cand.c];
+        auto src_d = hir.source_map[cand.d];
+        hir.source_map[cand.a] = std::move(src_c);
+        hir.source_map[cand.b] = std::move(src_d);
+        hir.source_map[cand.c] = std::move(src_a);
+        hir.source_map[cand.d] = std::move(src_b);
+    }
+
+    PeepholeFusionPass peephole;
```
bachase, line 674: How is this trying to compare and decide? The peephole fusion is over the entire HIR, but the comment at the top says MCR was done locally?

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+        case OpType::DETECTOR:
+        case OpType::OBSERVABLE:
+        case OpType::EXP_VAL:
+        case OpType::NUM_OP_TYPES:
+            return true;
+    }
+    return true;
+}
+
+int mul_phase_mod4(MaskView x1, MaskView z1, MaskView x2, MaskView z2) {
```
bachase, line 295: Is this the same as what is inside the `conjugate_pauli_by_s` function above? Might be worth a common function versus copy+paste.

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+
+            apply_to_ps(tab.xs[q]);
+            apply_to_ps(tab.zs[q]);
+        }
+    }
+}
+
+void normalize_t_sign(HirModule& hir, HeisenbergOp& op, std::complex<double>& global_weight) {
+    if (op.op_type() == OpType::T_GATE && hir.sign(op)) {
+        global_weight *= op.is_dagger() ? kExpMinusIPiOver4 : kExpIPiOver4;
```
bachase, line 168: This look to modify the op and the overall phase. Even though equivalent, does this mean the pass might modify the HIR even if it doesn't do any MCR/t-gate reductions?

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+                LocalFusionStats stats;
+                if (!apply_candidate(trial, *cand, stats))
+                    continue;
+
+                size_t before_t = hir.num_t_gates();
+                size_t after_t = trial.num_t_gates();
+                if (after_t >= before_t)
+                    continue;
+
+                hir = std::move(trial);
```
bachase, line 726: Does this move invalidate any references to what was in hir prior?

src/clifft/optimizer/experimental_mcr_t_count_pass.cc
```diff
@@ -0,0 +1,740 @@
...
+                 ++anchor_t_idx) {
+                auto cand = find_candidate_from_anchor(
+                    hir, window, anchor_t_idx, kLookaheadCap, candidates_considered_,
+                    merge_potential_rejects_, equivalence_checks_, gate_infos, original_span_cache);
+                if (!cand.has_value())
+                    continue;
+
+                ++quadruples_found_;
+
+                HirModule trial = hir;
```
bachase, line 716: This seems like it could be an expensive copy. Is there an alternative approach?

bachase (commented):
Thanks @EthanFeld for the continued work. I've left some specific comments below, but overall

1. You may have noticed I've closed some issues recently, in part for lack of sufficient human review of LLM generated work (or not even acknowledging the use of AI). I don't think that is the case here, but there are some spots that look like the overall PR could've used more human review. Please take that into consideration if you continue to work on this.
2. Setting aside the performance/speed, it doesn't seem we have evidence of this pass working on any circuits outside the ones built around the same style of a rotation kernel (which is good for correctness). Given some other PRs that do show some improvements on real-world circuits, not to mention the original references that show T-gate reductions on real world circuits, I'm not inclined to accept this as is/for the bounty. Its great to have the study and level of evaluation, but doesn't yet meet the bar for the issue. Sharing this feedback now, so you know that addressing some of the code questions I added would not be enough to meet the bar.

PR #127, opened 2026-06-08
feat(noncomp): Python bindings + spec-based model builder (G2a)

bachase:
> ⚠️ **Prototype — feature-branch work.** Targets the long-running `codex/noncomputational-structural-mvp` branch, not `main`, and is reviewed less strictly than main-targeting PRs.

G2a — makes the noncomputational MVP usable from Python. The oracle validation and sqale-style examples are the G2b follow-up.

## C++
`NonComputationalModel::from_spec(...)` builds every `TransitionInstrument` and the classifier against one `LevelSet` from raw matrices, so callers never construct those objects or touch level fingerprints/ids (design §3.3).

## Python (`clifft.noncomp`)
```python
from clifft import noncomp
model = noncomp.Model(initial_state=[1,0,0,0,0], transitions={"S": T},
                      classifier=noncomp.Classifier(["0","1"], P), reset_restores_lost=False)
r = noncomp.sample("H 0\nS 0\nM 0\n", model, shots=1000, seed=7)
# r.measurements / .detectors / .observables / .final_status  -> np.uint8
```
- Fixed built-in five-level set, named by `noncomp.Level` / `noncomp.LEVELS`; matrix rows/columns are indexed by `Level` (no level-id objects, no bare indices). There is no generic/custom-level path in Python.
- `final_status` is coarse: `{Computational, Leaked, Lost}` (`QubitStatusKind`); the specific leaked/lost level is not exposed.
- `reset_restores_lost` is the only exposed policy flag; `unknown_source_policy` stays internal (single value today).
- Errors surface only where reachable: bad shapes/probabilities, unknown gate keys, unknown-source transitions, and substochastic / non-binary classifiers.

## Deferred (per discussion)
Custom level sets, an exact `final_level` sidecar (specific leaked/lost level), the independent density-matrix oracle, and sqale-style examples → follow-up. The oracle will validate **supported output distributions**, not trace-out isolation.

## Tests
- C++: `from_spec` construction + validation/error propagation (7 cases).
- Python: model construction, end-to-end sampling, classifier/record semantics, record-layout invariants (24 cases).
- Full C++ suite green (914 cases); Python suite green (stim-dependent modules need the optional `stim` test dep).

### DIFF
--- a/src/clifft/noncomp/model.cc
+++ b/src/clifft/noncomp/model.cc
@@ -151,6 +151,28 @@ NonComputationalModel::NonComputationalModel(
     }
 }
 
+NonComputationalModel NonComputationalModel::from_spec(
+    LevelSet levels, std::vector<double> initial_state,
+    const std::map<std::string, std::vector<std::vector<double>>>& transition_matrices,
+    std::optional<ClassifierSpec> classifier_spec, NonComputationalPolicy policy) {
+    // Build each component against the single `levels` table, so every
+    // fingerprint matches by construction and the constructor's cross-object
+    // checks pass without the caller ever touching a fingerprint or level id.
+    std::map<std::string, TransitionInstrument> transitions;
+    for (const auto& [gate, matrix] : transition_matrices) {
+        transitions.emplace(gate, TransitionInstrument::from_matrix(matrix, levels));
+    }
+
+    std::optional<MeasurementClassifier> classifier;
+    if (classifier_spec.has_value()) {
+        classifier = MeasurementClassifier::from_matrix(std::move(classifier_spec->symbols),
+                                                        std::move(classifier_spec->matrix), levels);
+    }
+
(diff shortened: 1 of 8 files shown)


kawacukennedy (approved):
Clean implementation. The `from_spec` factory neatly hides LevelSet internals; Python wrapper is ergonomic and well-tested. CI already green — good to merge.

bachase:
@kawacukennedy thanks for your interest in clifft! Note that this is just a prototype and not committing to main, so I'm not looking for a review at this time while the approach iterates.

Also more generally on PRs in clifft, please be sure any comments and feedback are more high value. The ones I see from you so far are not given meaningful/actionable advice back to the submitters.

If you are interested in engaging in clifft, I'd recommend looking for an issue and trying to contribute to it rather than working to review. It will help you get more familiar with the codebase, which would lead to better reviews.

PR #131, opened 2026-06-08
build: hide vendored Stim symbols

### LINKED ISSUE #109: Hide vendored Stim symbols from the Python extension's exported symbol table
## Summary

`_clifft_core.so` currently exports all of the statically-linked Stim symbols in its dynamic symbol table. We should hide them so the extension exports only the nanobind entry point.

## Evidence

On the 0.4.1 PyPI wheel (manylinux_2_28_x86_64):

```
$ nm -D --defined-only _clifft_core.abi3.so | grep -c stim
260
$ nm -D --defined-only _clifft_core.abi3.so | wc -l
678
```

So ~38% of the extension's exported dynamic symbols are Stim internals (e.g. `stim::GateTarget::x`, `stim::Circuit` helpers). The same is observed in a local Spack source build (258 symbols). We statically link libstim (Stim's default `BUILD_SHARED_LIBS=OFF`) into the single `_clifft_core.so` that ships in the wheel, but the symbols remain default-visibility and are therefore exported.

## Why this matters

- **Smaller wheel + faster import**: a smaller dynamic symbol table means less work for the loader at import time.
- **Defense-in-depth against ODR**: today, cross-extension isolation relies entirely on CPython loading extensions with `RTLD_LOCAL`. That is sufficient for normal `pip`/Spack installs, but it breaks if anything in the process sets `RTLD_GLOBAL` (e.g. `sys.setdlopenflags(os.RTLD_GLOBAL)`, which some MPI/OpenMP setups do). If another component in the same process also vendors a different Stim version, the loader could then resolve across the two copies and cause segfaults/UB. Hiding the symbols removes this failure mode entirely.
- **Stop being an accidental Stim ABI surface**: nothing should be reaching into our `.so` via `dlsym` to call Stim internals.

This is purely hygiene for the extension — it does not change the public Python API.

## Proposed approach

Restrict exported symbols on the extension target to just the nanobind init symbol via a linker version script (Linux) / exported symbols list (macOS). Sketch:

```cmake
# in src/python/CMakeLists.txt, on the _clifft_core target
if(UNIX AND NOT APPLE)
    target_link_options(_clifft_core PRIVATE
        "LINKER:--version-script=${CMAKE_CURRENT_SOURCE_DIR}/clifft.lds")
elseif(APPLE)
    target_link_options(_clifft_core PRIVATE
        "LINKER:-exported_symbols_list,${CMAKE_CURRENT_SOURCE_DIR}/clifft.exp")
endif()
```

with a version script exporting only the module init symbol (`PyInit__clifft_core` for non-limited-API, or the appropriate symbol for the stable ABI build) and localizing everything else.

Note we build with `STABLE_ABI`, so the export list needs to cover the correct init symbol for the limited-API module.

## Acceptance criteria

- `nm -D --defined-only _clifft_core*.so | grep -c stim` returns 0
- `import clifft` works and the full test suite passes on Linux, macOS, and Windows
- nanobind stub generation (`nanobind_add_stub`) still succeeds
- Wheel size does not regress (should shrink)

## Out of scope

Externalizing Stim (shared/system libstim). We intentionally vendor a pinned Stim version because Stim only guarantees Python-API stability, not C++ ABI stability, and we rely on internal C++ types plus `SIMD_WIDTH`/ISA-baseline coupling between clifft and libstim.

bachase:
## Summary
- restrict the Python extension export table to the module init symbol on Linux and macOS
- add a reusable CI audit that imports clifft._clifft_core and checks dynamic exports for leaked Stim symbols
- run the audit in CI Python installs and cibuildwheel release smoke tests

Closes #109

## Verification
- uv run python tools/ci/audit_python_exports.py
- nm -gU .venv/lib/python3.14/site-packages/clifft/_clifft_core.abi3.so
- uv run pytest tests/python/ -v

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,6 +184,9 @@ jobs:
       - name: Install Python package
         run: uv pip install -e ".[dev]"
 
+      - name: Audit Python extension exports
+        run: uv run python tools/ci/audit_python_exports.py
+
       - name: Run Python tests (default ISA)
         run: uv run pytest tests/python/ -v
 
@@ -252,6 +255,9 @@ jobs:
         run: |
           uv venv .smoke-venv
           uv pip install --python .smoke-venv -e ".[dev]"
+
+      - name: Audit Python extension exports
+        run: .smoke-venv/bin/python tools/ci/audit_python_exports.py
 
       - name: Install qemu-user
         run: sudo apt-get install -y qemu-user
@@ -335,6 +341,9 @@ jobs:
 
       - name: Install Python package
         run: uv pip install -e ".[dev]"
+
+      - name: Audit Python extension exports
+        run: uv run python tools/ci/audit_python_exports.py
 
       - name: Run Python tests
         run: uv run pytest tests/python/ -v
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,8 +91,11 @@ jobs:
           CIBW_BEFORE_BUILD: pip install setuptools-scm
           # Install libomp on macOS for OpenMP support in wheels
           CIBW_BEFORE_ALL_MACOS: brew install libomp
-          # Smoke-test: import the built package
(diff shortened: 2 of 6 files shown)


bachase (commented):
👍

PR #140, opened 2026-06-11
fix: preserve global phase across tableau rewrites

### LINKED ISSUE #139: PeepholeFusionPass changes get_statevector global phase when absorbing S rotations
Found during discussion/review of the current PR #132.

`PeepholeFusionPass` can change the API-visible global phase of `get_statevector()` when it fuses two T gates or S-angle `PHASE_ROTATION`s and absorbs the resulting S/S_dag into the Clifford frame via `apply_virtual_s_downstream`.

The relative amplitudes are preserved, so fidelity-based tests pass, but componentwise statevectors differ by a global factor. Since `PeepholeFusionPass` is in the default pipeline, this affects `clifft.compile(...)`.

Minimal repro:

```stim
H 0
T 0
T 0
H 0
```

Expected/unoptimized statevector:

```text
[0.5+0.5j, 0.5-0.5j]
```

Default optimized statevector:

```text
[0.70710677+0j, 0-0.70710677j]
```

The optimized result is `exp(-i*pi/4)` times the expected result.

This appears to come from replacing the physical fused S operation with a post-hoc tableau update. The tableau fixes a Clifford only up to global phase, but the pass does not compensate `global_weight` for the phase convention chosen later by `to_flat_unitary_matrix()`.

This matters because Clifft otherwise tracks global phase explicitly, and docs describe `get_statevector()` as exact state inspection.

Likely fix options:

1. Keep the fused S/S_dag as a runtime `PHASE_ROTATION` instead of absorbing it into the tableau.
2. Or compute the canonical tableau phase delta when applying `U_C := U_C S_P` and multiply it into `global_weight`.

The test suite should add componentwise optimized-vs-unoptimized statevector checks for S absorption paths, without global-phase alignment.

### LINKED ISSUE #141: Backend virtual-frame composition changes get_statevector global phase
Found while validating the fix for #139 (PR #140).

`lower()` composes the accumulated virtual Clifford frame into the constant-pool tableau:

```cpp
// backend.cc
stim::Tableau<kStimWidth> v_cum_inv = final_v_cum.inverse();
ctx.constant_pool.final_tableau = v_cum_inv.then(*hir.final_tableau);
```

Like the peephole absorption fixed in #140, this replaces an exact operator factorization with a tableau composition, and `to_flat_unitary_matrix()`'s canonical phase for the composed tableau is unrelated to `canonical(hir.final_tableau)` times the frame gates' unitaries. No compensation is applied to `global_weight`, so the API-visible global phase of `get_statevector()` shifts whenever runtime T/rotation ops localize through virtual frame gates.

Minimal repro (after #140, so the optimizer side is exact):

```stim
CX 0 1
S_DAG 0
H 0
T 0
T 0
X 1
```

- HIR value (dense `global_weight * canonical(final_tableau) * op-stream`): identical before/after `PeepholeFusionPass` — the optimizer is not at fault.
- Optimized pipeline (no runtime ops left, trivial frame): statevector equals the HIR value exactly: `[0, 0, 0.7071i, -0.7071]`.
- Unoptimized pipeline (two runtime T ops, virtual H in the frame): `[0, 0, 0.7071, 0.7071i]` — the HIR value times `-i`.

Consequences:

- optimized-vs-unoptimized componentwise statevector equality fails on ~5% of random 4-5 qubit, depth-30 Clifford+T circuits (the two pipelines accumulate different virtual frames), even though both fidelities are 1;
- unoptimized statevectors deviate from the exact circuit unitary by a unit phase on T-circuits whose Clifford part is trivial.

Likely fix: the same canonical-phase machinery from #140 generalizes — when composing, chain per-gate phase deltas `canonical(A) * u_gate == phase * canonical(A * u_gate)` over the frame gate log (every `PendingGateType` has a sparse canonical matrix with at most two entries per column) and fold the product into `constant_pool.global_weight`. That requires keeping a log of appended frame gates (the materialized tableau alone is not enough: the effective phase convention comes from the concrete gate sequence) and costs O(n^3/64) bit-ops per frame gate, so it likely needs gating, e.g. only for circuits where the phase is observable, or an incremental formulation.

Until then, exact componentwise statevector tests should avoid circuits where leftover runtime rotations localize through frame gates (see the test selection note in PR #140).

bachase:
Fixes #139. Fixes #141.

## Problem

Two compiler stages replace exact unitary factors with stim tableau updates, and a tableau fixes its Clifford only up to global phase — `to_flat_unitary_matrix()` resolves the ambiguity by making the first nonzero entry of the row-major flat matrix real positive. Neither stage compensated `global_weight` for that convention:

1. **PeepholeFusionPass** (#139): fusing two T gates (or S-angle rotations) and absorbing the resulting S/S_dag into the final tableau dropped an eighth root of unity (the `H T T H` repro is off by `exp(-i*pi/4)`).
2. **Backend lowering** (#141): `lower()` composes the deferred virtual frame into the constant-pool tableau (`v_cum_inv.then(final_tableau)`), re-anchoring the phase whenever leftover rotations localize through frame routing.

The two errors compose multiplicatively, so they cancel on some circuit classes and compound on others — fixing either alone flips previously exact circuits to phase-wrong, which is why both fixes land together.

## Fix

Both corrections compute exact canonical-phase deltas via the Choi state of the tableau (a 2n-qubit stabilizer state whose minimum-integer support index is stim's canonical anchor), shared in the new `util/canonical_phase.{h,cc}` — Gauss-Jordan support reduction plus anchored amplitude walks, polynomial in n, no dense matrices:

- The peephole folds `canonical(U_C) * R` evaluated at the new canonical anchor into `hir.global_weight` on every S absorption, where `R = Pi_+ +- i*Pi_-` is the exact projector-form rotation removed from the op stream.
- `VirtualFrame` keeps a log of every appended frame gate, and `lower()` chains per-gate deltas `canonical(B) * u == phase * canonical(tableau(B * u))` from the composed tableau back to the HIR tableau, folding the conjugate product into `global_weight`. Every pending-gate matrix has at most two entries per column, so each link is one support reduction plus at most two amplitude walks.
- Programs with measurements or noise skip the backend correction: their outputs are insensitive to global phase and their frame logs grow with every localized measurement, so sampling/QEC compiles are unaffected.

The tableau update and both phase helpers are exposed under `clifft::internal` so tests can fuzz them directly against stim's dense matrices.

## Validation

- `s_absorption_phase` fuzzed against dense stim matrices: 256 random tableau/Pauli/sign/dagger combinations, componentwise (`test_optimizer.cc`).
- `frame_composition_phase` fuzzed through the production `VirtualFrame` composition path: 128 random tableau/gate-log combinations, componentwise (`test_backend.cc`).
- Pass-level value preservation: 200 random Clifford+T+R_Z circuits, comparing `global_weight * canonical(final_tableau) * (op stream)` densely before/after the pass, with an aggregate fusion-count vacuity guard (`test_optimizer.cc`).
- Embedding-invariance oracle pinning multi-word Choi indices (72-qubit frames straddling bit 64) to the dense-validated small-frame value.
- End-to-end componentwise optimized-vs-unoptimized statevector checks in C++ and Python, including absorptions that leave rotations needing frame routing, plus 20 random circuits with **no global-phase alignment** — the test #139 originally asked for. The Python oracle's `optimize=False` baseline now genuinely disables both optimization stages.
- 192-circuit componentwise stress suite: 61 mismatches on main, 0 with this PR.
- Full suites: 769/769 C++ (Debug), 712/712 Python; pre-commit clean.

## Performance

- Measured/noisy programs (sampling, QEC — all repo fixtures): zero overhead; compile times unchanged.
- Peephole fusions: ~30 us/fusion at n=20 up to ~8 ms at n=400 (two Choi reductions per absorption). Stabilizer-only circuits never fuse.
- Unitary measurement-free programs with frame routing: one Choi reduction per frame gate at composition; worst engineered case (n=200, wide multi-qubit T axes) 3.4 ms -> 98 ms compile. The SVM rank cap (k < 63) bounds the frame log for any compilable unitary circuit, and these are exactly the statevector-inspection workloads where the 2^k simulation dominates compile time. Known follow-up optimizations (support caching across fusions, incremental support updates) fit inside `util/canonical_phase` without interface changes.


src/clifft/optimizer/peephole.h
```diff
@@ -1,11 +1,33 @@
...
 #include "clifft/optimizer/hir_pass.h"
+#include "clifft/util/mask_view.h"
+#include "clifft/util/stim_mask.h"
 
+#include <complex>
 #include <cstddef>
 
 namespace clifft {
 
+namespace internal {
```
bachase, line 12: Should these just be in the cc file?

bachase: They're declared here only so the test oracles can call them directly: `test_optimizer.cc` fuzzes `s_absorption_phase` against dense `to_flat_unitary_matrix()` ground truth (256 cases) and pins the multi-word path via the embedding-invariance check — both need the helper symbols, and file-local functions in the .cc wouldn't be reachable from the test TU. Testing only through `PeepholeFusionPass::run` would limit us to n <= 10-ish e2e checks and lose the direct oracle.

Same pattern as `internal::` in `compiler_context.h`. Happy to move them to a separate `peephole_internal.h` if you'd rather keep the public header minimal.

src/clifft/backend/backend.cc
```diff
@@ -516,6 +517,112 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
...
+            ChoiIndex idx = anchor;
+            if (choi_index_bit(idx, g.axis_1) != choi_index_bit(idx, g.axis_2)) {
+                choi_index_flip_bit(idx, g.axis_1);
+                choi_index_flip_bit(idx, g.axis_2);
+            }
+            w = choi_amplitude(old_support, idx);
+            break;
+        }
+
+        case PendingGateType::H: {
```
bachase, line 566: What happens if we hadd new pending gate types? Should we add a default/error case?

src/clifft/backend/backend.cc
```diff
@@ -516,6 +517,112 @@ LocalizationResult localize_pauli(CompilerContext& ctx,
...
+                break;
+            case PendingGateType::S:
+                composed.prepend_SQRT_Z(it->axis_1);
+                break;
+            case PendingGateType::H:
+                composed.prepend_H_XZ(it->axis_1);
+                break;
+            case PendingGateType::SWAP:
+                composed.prepend_SWAP(it->axis_1, it->axis_2);
+                break;
```
bachase, line 621: Same question as above on non-exhaustive switch

src/clifft/backend/backend.cc
```diff
@@ -843,7 +950,31 @@ CompiledModule lower(const HirModule& hir, std::span<const uint8_t> postselectio
...
+
+        // The runtime ops were rewritten through the virtual frame gates,
+        // but the composed tableau realizes those gates only up to stim's
+        // canonical matrix phase. Fold the per-gate canonical deltas into
+        // global_weight so the API-visible global phase survives the
+        // composition. Programs with measurements or noise skip this: their
+        // observable outputs are insensitive to a global phase, and their
+        // frame gate logs grow with every localized measurement.
+        const auto& gate_log = ctx.virtual_frame.gate_log();
+        bool deterministic = true;
```
bachase: Is this beter extracted into a "hir.is_derministic" or similar function?

src/clifft/util/canonical_phase.h
bachase, line 1: Since this is now its own file, should the tests also be in their own standalone cc file?

src/clifft/backend/backend.cc
bachase: Kept the switch exhaustive with no `default` so `-Wswitch` flags a new `PendingGateType` at compile time, and promoted the runtime guard from a Debug assert to an unconditional throw (c563e9e): an unhandled kind now leaves `w == 0` and fails loudly with `std::logic_error` in every build type, instead of silently poisoning `global_weight` with NaN in Release.

src/clifft/backend/backend.cc
bachase: Same treatment: no `default` so the compiler warns on a new gate kind, and the chain's end-to-end consistency check (`composed == target`) is now an unconditional throw rather than an assert — a gate the prepend switch missed cannot reproduce the target tableau, so it fails loudly in all build types.

src/clifft/backend/backend.cc
bachase: Done — extracted to `HirModule::is_deterministic()` (c563e9e), with a doc comment defining it as a fixed-unitary evolution (no measurements, noise, readout noise, or measurement-conditioned Paulis).

src/clifft/util/canonical_phase.h
bachase: Done — the canonical-phase oracle tests (S absorption dense fuzz, embedding invariance, frame composition dense fuzz) now live in `tests/test_canonical_phase.cc`, mirroring the util TU, with the shared dense helpers consolidated into `test_helpers.h` (c563e9e).

PR #143, opened 2026-06-11
fix(backend): preserve global phase across virtual-frame composition

### LINKED ISSUE #141: Backend virtual-frame composition changes get_statevector global phase

bachase:
Fixes #141. Stacked on #140 — review and merge that first; this PR's base is its branch.

## Problem

`lower()` defers virtual frame gates symplectically and composes them into the constant-pool tableau (`v_cum_inv.then(*hir.final_tableau)`). The composed tableau realizes those gates only up to stim's canonical matrix phase, so `get_statevector()`'s global phase shifted whenever leftover rotations localized through frame routing. This error composed with the peephole absorption phase: on `H 0; T 0; T 0; T 0; H 0; T 0` the two bugs cancelled on main, #140 alone exposed the backend error (`e^{i*pi/4}`), and on most random circuits the composition broke agreement in the other direction.

## Fix

Track the phase explicitly instead of letting the composition re-anchor it:

- `VirtualFrame` keeps a log of every appended gate (flushing still materializes into the tableau; the log preserves the concrete sequence, which the materialized tableau alone cannot supply).
- `lower()` chains per-gate canonical deltas `canonical(B) * u == phase * canonical(tableau(B * u))` from the composed tableau back to the HIR tableau, walking the log in reverse application order, and folds the conjugate product into `global_weight`. Each pending-gate matrix (H, S, CNOT, CZ, SWAP) has at most two entries per column, so every link costs one Choi support reduction plus at most two amplitude walks.
- The Choi machinery moves from `peephole.cc` to a shared `util/canonical_phase.{h,cc}` (both `src/clifft` and `src/wasm` source lists updated); the peephole's `s_absorption_phase` now uses it.
- Programs with measurements or noise skip the correction: every observable output there is insensitive to global phase, and their frame logs grow with every localized measurement (QEC-scale compiles are unaffected — fixture compile times unchanged).

## Validation

- `frame_composition_phase` fuzzed against dense stim matrices through the production `VirtualFrame` composition path: 128 random tableau/gate-log combinations, componentwise (`test_backend.cc`).
- The #141 repro class is exact again: `H T T T H T` gives optimized == unoptimized == truth; the directed corpus in C++/Python now includes leftover-rotation circuits that were previously impossible to test componentwise.
- New randomized componentwise test: 20 random 5q/depth-30 Clifford+T circuits, optimized vs unoptimized with **no global-phase alignment** (`test_peephole_oracle.py`) — the test issue #139 originally asked for, now achievable.
- 192-circuit stress suite: 61 componentwise mismatches before #140, 10 after #140 alone, **0 with this PR**.
- Full suites: 769/769 C++ (Debug), 712/712 Python; pre-commit clean.

## Cost

Only measurement/noise-free programs with frame gates pay: one Choi reduction per frame gate at composition time (~0.35 ms/gate at n=100). The SVM rank cap (k < 63) bounds leftover activations, so the chain is bounded for any compilable unitary circuit; `qv10`/`cultivation_d5`/`target_qec` fixtures (all measured) are byte-identical in compile time.

### DIFF
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -26,6 +26,7 @@ add_library(clifft_core STATIC
     svm/svm_forced_kernels.cc
     svm/svm_scalar.cc
     svm/svm_state.cc
+    util/canonical_phase.cc
     util/introspection.cc
 )
 
--- a/src/clifft/backend/backend.cc
+++ b/src/clifft/backend/backend.cc
@@ -1,6 +1,7 @@
 #include "clifft/backend/backend.h"
 
 #include "clifft/backend/compiler_context.h"
+#include "clifft/util/canonical_phase.h"
 #include "clifft/util/mask_view.h"
 
 #include <algorithm>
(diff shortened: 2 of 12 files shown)


bachase:
Note: GitHub shows this as merged because its commits were fast-forwarded into the #140 branch — the two fixes can only land together (each alone flips a circuit class from exact to phase-wrong), so #140 now carries both as one atomic PR. Nothing landed on main yet; #141 stays open until #140 merges.

PR #144, opened 2026-06-11
feat(python): Pauli-twirl helpers for coherent control errors

bachase:
> [!NOTE]
> Prototype work on the `codex/noncomputational-structural-mvp` feature branch — less-strict review bar than `main`.

Work item 4 of the checkpoint plan: coherent-error compatibility. Python-only.

### `clifft.twirl`

Two helpers, deliberately kept **outside** the noncomputational model surface — coherent control errors are unitaries, not status changes, and they enter circuits as ordinary noise instructions:

- `pauli_probabilities(unitary)` — the standard twirl `p_P = |tr(U·P)|²/4` of a 2×2 unitary into `(p_x, p_y, p_z)`, global-phase-invariant, validated (shape/unitarity). Maps directly onto `PAULI_CHANNEL_1(px, py, pz)` or a single `*_ERROR` when one component is nonzero.
- `rotation(axis, radians)` — the closed form `sin²(θ/2)` on the rotation axis, for over-rotation parameters.

The module docstring states the contract precisely: the twirl preserves the diagonal of the unitary's Pauli transfer matrix and discards the coherences between Pauli components — circuit statistics generally differ from the exact unitary even for a single application.

### Tests (8 new)

- Helper math vs closed forms (rotations, X/H/S known cases, validation errors).
- **Exact-vs-twirl through clifft itself**: one *selected* matching case (one S between Hadamards, where the discarded coherence never reaches the measurement) and two counterexamples — a **single** twirled Hadamard (exact H·H is the identity, the twirled channel gives a coin flip) and coherent accumulation of two S gates (1.0 vs 0.5) — all asserted, so the contract cannot regress into "twirl equals exact unless repeated".
- Passthrough: `Z_ERROR` and `PAULI_CHANNEL_1` are statistically unchanged through the noncomputational path on computational qubits, and Pauli noise on a lost operand is dropped.

Python suite: 592 green (583 + 9 new; qiskit-optional modules excluded as usual). Pre-commit clean. No C++ changes.

### Review round

Reworded the twirl contract per review (it preserves the Pauli-transfer-matrix diagonal, not single-use circuit statistics — the H·H counterexample is now a committed test) and reframed the S-between-Hadamards test as a selected matching case.

### DIFF
--- a/src/python/clifft/twirl.py
+++ b/src/python/clifft/twirl.py
@@ -20,6 +20,12 @@ These helpers only compute probabilities. Insert them into circuits as
 ordinary noise instructions (``Z_ERROR(p)``, ``PAULI_CHANNEL_1(px, py, pz)``),
 which both the plain sampler and the noncomputational path treat like any
 other Pauli noise.
+
+This module is experimental. It operates on one unitary at a time and leaves
+the circuit editing to the caller; a more robust interface -- a circuit
+annotated with the sites to twirl, with the noise nodes substituted
+internally -- may supersede these free functions, changing or removing this
+surface.
 """
 
 from __future__ import annotations
--- a/src/python/clifft/twirl.py
+++ b/src/python/clifft/twirl.py
@@ -1,0 +1,73 @@
+"""Pauli-twirling helpers for coherent control errors.
+
+A coherent control error -- an over-rotation, a phase miscalibration -- is a
+unitary, not a noncomputational status change. Clifft can simulate a coherent
+error exactly at the usual active-rank cost; the standard fast-path
+alternative is to *twirl* it: replace the error unitary ``U`` with the Pauli
+channel whose probabilities are
+
+    p_P = |tr(U P)|^2 / 4,    P in {I, X, Y, Z}.
+
+A single twirled error reproduces the exact computational-basis statistics of
+one application; errors that interfere across several gates (echo sequences,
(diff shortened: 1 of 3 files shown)


src/python/clifft/twirl.py
```diff
@@ -0,0 +1,77 @@
...
+
+The twirled channel preserves the diagonal of ``U``'s Pauli transfer matrix
+and discards the coherences between Pauli components. That is an
+approximation, not an equivalence: circuit statistics under the twirled
+channel generally differ from the exact unitary even for a single
+application (twirling one Hadamard turns a deterministic H-H identity into a
+coin flip), and they coincide only in selected configurations where the
+discarded coherences never reach a measurement.
+
+These helpers only compute probabilities. Insert them into circuits as
```
bachase, line 19: I think this is fine for our MVP, but if we want these longer term, I think we'd want a more robust interface for a user to take a circuit, somehow annotated with sites to twirl, and then these get applied internally and replace nodes.

We might want to consider framing this as experimental, or have some type of TODO comment to that effect.

bachase: Agreed — added an experimental note to the module docstring in 00b3594: the surface operates on one unitary at a time and leaves circuit editing to the caller, and a circuit-annotation interface with internal node substitution may supersede it.

PR #134, opened 2026-06-09
feat: add experimental global T-count pass for issue #40

rosspeili:
Adds opt in McrReorderPass, ToddPhasePass, and GlobalTcountPass for global T count reduction on Clifft HIR. Uses bounded MCR reordering and size capped TODD on commuting phase polynomial clusters. Includes evaluation harness, equivalence tests, and a short theory note. Not in the default pipeline.

Closes #40

## Summary

Prototype for issue #40. Checks if global T count optimization can work at the HIR level without changing the parser, bytecode, or VM. Use PeepholeFusionPass, then GlobalTcountPass, then PeepholeFusionPass.

## Test plan

- [x] Python tests pass
- [x] Doc codeblocks pass for docs/
- [x] Evaluation harness runs
- [x] Pre commit formatting and lint on our files
- [ ] C++ tests not run locally on Windows

## Contributor agreement

- [x] I confirm this is my original work and I license it under the project Apache 2.0 license.
- [ ] AI assisted trailer not included.

PR #124, opened 2026-06-07
feat: add experimental global T-count pass for issue #40

luffy-orf:
## Summary

Prototype for [#40](https://github.com/unitaryfoundation/clifft/issues/40): an opt-in HIR pass that explores whether **global** T-count reduction works at the Heisenberg-IR level.
Adds `ExperimentalGlobalTcountPass`, a two-phase optimizer:
1. **MCR phase** — bounded multiplicative-commutator-relation reordering on contiguous T-gate windows to expose same-axis fusion opportunities for peephole.
2. **TODD phase** — size-capped TODD on commuting phase-polynomial clusters within T windows (maintainer-suggested direction from the issue thread).
Also extracts shared `virtual_clifford` logic from peephole for reuse by MCR/TODD re-emission.
**Not in the default pipeline.** Intended usage:
```python
pm = clifft.HirPassManager()
pm.add(clifft.PeepholeFusionPass())
pm.add(clifft.ExperimentalGlobalTcountPass())
pm.add(clifft.PeepholeFusionPass())
```

## Test plan

- [x] C++ tests pass (`ctest --test-dir build -E Bench --output-on-failure`)
- [x] Python tests pass (`uv run pytest tests/python/ -v`)
- [x] Doc examples pass (`uv run pytest --codeblocks docs/ README.md -v`)
- [x] Pre-commit checks pass (`uv run pre-commit run --all-files`)

## Contributor agreement

- [x] I confirm this contribution is my original work (or I have the right to
  submit it) and I license it under this project's
  [Apache-2.0 license](../LICENSE).
- [x] If this contribution was AI-assisted, I have reviewed the generated code
  and included an `Assisted-by:` git trailer in the commit message.

PR #154, opened 2026-06-15
feat: add exact phase-polynomial T-count pass

### LINKED ISSUE #40: Explore HIR-level global T-count optimization

DennisWayo:
## Summary

Closes #40.

Adds `ExactPhasePolynomialTCountPass`, an opt-in HIR pass for bounded exact T-count minimization on small commuting T-gate blocks.

This targets #40, maintainer guidance that the prototype should:

- work with Clifft's C++/HIR internals rather than as an external Python script:
  https://github.com/unitaryfoundation/clifft/issues/40#issuecomment-4624736921
- clearly state the theory/citations and include broader before/after evidence:
  https://github.com/unitaryfoundation/clifft/issues/40#issuecomment-4650387827
- be different or improved relative to the current PRs:
  https://github.com/unitaryfoundation/clifft/issues/40#issuecomment-4670807317

This new pass is registered with `default_enabled = false`, exposed through the Python binding, included in the Wasm source list, and documented under `docs/theory/exact_phase_polynomial_t_count.md`.

## What the pass does

For each contiguous `T_GATE` block:

1. Require pairwise commuting Pauli axes.
2. Build a GF(2) coordinate basis from the HIR `(x, z)` masks, capped at rank 4.
3. Convert signed HIR Paulis and signed Pauli products into an affine
   phase-polynomial model, tracking constants in `global_weight`.
4. Exhaustively search smaller odd-parity representatives, trying both T and
   T-dagger signs.
5. Accept a candidate only when the exact residual truth table is Clifford:
   no degree >= 3 ANF terms, even linear coefficients, and quadratic
   coefficients in `{0, 4}` over `Z_8`.
6. Emit the remaining odd terms as `T_GATE` / `T_DAG` and the Clifford residual
   as `PHASE_ROTATION`; absorb constant residual phase into `global_weight`.

The pass skips a rewrite if the total emitted HIR op count would exceed the original block length.

## How this differs from the other PRs

| PR | Approach there | Difference here |
|---|---|---|
| #123 | Experimental MCR pass using multiproduct commutator reordering. | This PR does no MCR reordering and does not cross noncommuting blockers. It only decodes already-commuting HIR T blocks exactly. |
| #124 | Combined MCR phase plus size-capped TODD-style phase. | This PR deliberately avoids mixing two algorithms. It is a single exact bounded decoder with an exhaustive acceptance check and no shared virtual-Clifford refactor. |
| #132 | TOHPE-style phase-polynomial pass with a broader heuristic/scoring implementation. | This PR is smaller and less broad, but every accepted rewrite is found by exhaustive search over the full rank-capped truth table. It is positioned as an exact verifier/small-block cleanup, not a TOHPE implementation. |
| #134 | MCR + TODD global T-count prototype, now closed. | This PR has a distinct implementation, narrower scope, explicit sign/global-phase dense-matrix tests, and cites the specific theory it relies on. |

The tradeoff is intentional: this is not expected to beat TOHPE/TODD on large benchmark corpora. It provides a conservative exact HIR-native reduction that can be reviewed independently and used as a correctness target for future larger reducers.

## Theory and citations

The phase-polynomial model and Reed-Muller framing follow:

- Matthew Amy and Michele Mosca, "T-count optimization and Reed-Muller codes":
  https://arxiv.org/abs/1601.07363
- Luke Heyfron and Earl Campbell, "An Efficient Quantum Compiler that reduces
  T count":
  https://arxiv.org/abs/1712.01557
- Vivien Vandaele, "Lower T-count with faster algorithms":
  https://arxiv.org/abs/2407.08695

MCR is cited for contrast, not used here:

- Yusei Mori, Ryusei Hakoshima, and Keisuke Fujii, "Nontrivial multi-product
  commutation relation toward reducing T-count in sequential Pauli-based
  computation":
  https://arxiv.org/abs/2509.20052

The docs also link the op-T-mize dataset history as the benchmark family this
line of work should eventually target once Clifft has an importer:

- https://pennylane.ai/blog/2025/01/optimizing-with-op-t-mize-dataset

## Evaluation

The reproducible evaluator is `tools/bench/exact_phasepoly_tcount.py`. It runs
`trace -> PeepholeFusionPass -> ExactPhasePolynomialTCountPass` and prints a
Markdown table:

```bash
python tools/bench/exact_phasepoly_tcount.py
python tools/bench/exact_phasepoly_tcount.py --qc-dir path/to/qc_corpus
python tools/bench/exact_phasepoly_tcount.py --qasm-dir path/to/qasm_corpus \
  --skip-unsupported --max-imported-gates 5000
```

The `.qc` importer is intentionally conservative. It supports Clifford+T primitives plus one-, two-, and three-argument `tof` / `toffoli`; wider multi-control Toffoli gates are rejected instead of silently decomposed with implicit ancillas.

The QASM importer is similarly bounded. It supports OpenQASM 2.0 declarations plus common Clifford+T gate names, ignores barriers and measurements for T-count evaluation, and rejects custom or parameterized gates instead of attempting synthesis.

Positive cases are direct HIR unit fixtures, because they let the tests specify
the exact phase-polynomial axes without depending on a source-circuit importer.

| Case | T before | T after | Notes |
|---|---:|---:|---|
| Rank-4 Reed-Muller zero word, all 15 nonzero Z parities | 15 | 0 | Exact zero phase function. |
| Rank-4 zero word plus one repeated parity | 16 | 1 | Removes the 15-term zero word. |
| Rank-4 unsigned product word with affine Pauli signs | 15 | 0 | Emits six Clifford residual rotations; dense matrix equivalence checks `global_weight`. |
| Rank-4 signed product zero word plus one signed odd term | 16 | 1 | Normalizes to unsigned `T_DAG` plus `exp(i*pi/4)` global phase; dense matrix equivalence checked. |
| Rank-5 all nonzero Z parities | 31 | 31 | Skipped by rank cap. |
| Noncommuting T block | 3 | 3 | Skipped by commuting guard. |

Source-level integration evidence using gates Clifft already parses:

| Source circuit | Traced T | After peephole | After exact pass | Notes |
|---|---:|---:|---:|---|
| Complete `CCZ` hypergraph on 4 qubits, `CCZ 012`, `013`, `023`, `123` | 28 | 8 | 7 | Parsed from source text, traced to HIR, peephole-optimized, then reduced by this pass. |

Related source-level boundary checks:

| Source family | Result after peephole and exact pass | Reason |
|---|---|---|
| Complete `CCZ` hypergraph on 3 qubits | 7 -> 7 | A single CCZ is already optimal for this pass. |
| Complete `CCZ` hypergraphs on 5, 6, and 7 qubits | 20 -> 20, 20 -> 20, 63 -> 63 | The exact pass is rank-capped at four. |
| `CCX` ladders on 3 through 7 qubits | No reduction beyond peephole | The traced phase structure is sparse for this decoder. |

I also ran the pass after `PeepholeFusionPass` on existing repository fixture and tutorial circuits:

| Circuit | Traced T | After peephole | After exact pass | Blocks optimized |
|---|---:|---:|---:|---:|
| `tests/fixtures/cultivation_d5.stim` | 72 | 72 | 72 | 0 |
| `tests/fixtures/qv10.stim` | 0 | 0 | 0 | 0 |
| `tests/fixtures/target_qec.stim` | 0 | 0 | 0 | 0 |
| `docs/guide/circuits/circuit_d3_s_gate_p0.001.stim` | 0 | 0 | 0 | 0 |
| `docs/guide/circuits/circuit_d3_t_gate_p0.001.stim` | 29 | 29 | 29 | 0 |

External QASM smoke benchmark: I also ran the importer/evaluator on a local checkout of Qiskit's `test/benchmarks/qasm` fixtures, skipping unsupported parameterized/custom-gate files and capping imported circuits at 5000 gate lines:

```bash
python tools/bench/exact_phasepoly_tcount.py \
  --qasm-dir ../qiskit/test/benchmarks/qasm \
  --skip-unsupported --max-imported-gates 5000
```

Supported Clifford+T/QASM cases under that cap produced:

| Circuit | Traced T | After peephole | After exact pass | Blocks optimized |
|---|---:|---:|---:|---:|
| `20QBT_45CYC_.0D1_.1D2_3.qasm` | 0 | 0 | 0 | 0 |
| `53QBT_100CYC_QSE_3.qasm` | 0 | 0 | 0 | 0 |
| `54QBT_25CYC_QSE_3.qasm` | 0 | 0 | 0 | 0 |
| `depth_4gt10-v1_81.qasm` | 63 | 39 | 39 | 0 |
| `depth_4mod5-v0_19.qasm` | 14 | 8 | 7 | 1 |
| `depth_mod8-10_178.qasm` | 147 | 87 | 87 | 0 |
| `time_cnt3-5_179.qasm` | 70 | 60 | 60 | 0 |
| `time_cnt3-5_180.qasm` | 210 | 136 | 136 | 0 |

`hwb12.qasm` imported to 171482 gate lines and was excluded by the gate cap; a 20-second temporary evaluator run did not complete, so it is not counted as evidence for or against this pass.

The null result on Clifft's existing fixtures is expected for this prototype: those fixtures do not expose rank-capped dense commuting phase-polynomial blocks after tracing and local peephole fusion.

## Reviewer-bar checklist

- [x] C++ implementation against Clifft HIR, not an external Python-only tool.
- [x] Opt-in pass, default-disabled, with no parser, HIR layout, bytecode, or VM
  changes.
- [x] Clear theory note with citations for phase-polynomial/Reed-Muller,
  TODD/TOHPE context, and MCR contrast.
- [x] Explicitly different from MCR, TODD, and TOHPE PRs: this is a bounded
  exhaustive decoder for already-commuting blocks.
- [x] Tests cover exact rewrites, skipped blocks, rank cap, noncommuting guard,
  source-level `CCZ_4`, affine Pauli signs, and global phase preservation.
- [x] Evaluation includes existing repository fixtures and documents null
  results.
- [x] Reproducible benchmark harness can evaluate built-in source families,
  existing fixtures, supported external `.qc` files, and supported Clifford+T
  QASM files.
- [x] Bounded external Qiskit QASM smoke results are included, with unsupported
  files and the oversized `hwb12.qasm` case documented.
- [ ] Broader real-world benchmark improvements are not demonstrated here. See detailed explanation below; what do you think @bachase?

## What would make the broader real-world benchmark happen

In my opinion; I do not think this PR should try to claim the broader benchmark result by stretching the scope of this pass. The useful next step is to run the same pipeline on a pinned public corpus and report the full outcome, including null results and skipped files. The most relevant inputs would be op-T-mize-style `.qc` Clifford+T circuits, RevLib-derived reversible circuits after a documented Clifford+T decomposition, or a public QASM set restricted to gates this importer handles exactly. For each corpus, the report should include the source URL, commit or release, license, command line, and the reason for any skipped file.

The benchmark harness added here is meant to make that follow-up mechanical: it already handles `.qc`, bounded Clifford+T QASM, unsupported-file reporting, and a gate-count cap for very large imports. If a suitable external fixture has a compatible license, I would add one small case to the repo as a regression test that asserts the T-count before peephole, after peephole, and after this pass.

If a broader corpus mostly shows no improvement, that is still useful evidence. It would mean this pass is best kept as an exact small-block cleanup or verifier, and that the real production win needs a separate transformation that exposes larger phase-polynomial structure, such as a commuting-window collector or a TODD/TOHPE-style reducer. To make the comparison fair, I would run this PR and the other open PRs on the same corpus and report traced T-count, post-peephole T-count, final T-count, runtime, skipped files, and semantic coverage in the same table. Grammar was refined  by Codex.

## Limitations

- Only contiguous `T_GATE` blocks are considered.
- The pass does not reorder through non-T HIR ops and does not optimize
  noncommuting windows.
- `max_rank` is clamped to 4, because the search is exhaustive.
- Clifford residuals are emitted as `PHASE_ROTATION` ops; a later peephole pass
  may absorb some of them, but this pass does not require that.
- It is default-disabled and should be treated as an experimental prototype,
  not a production optimizer or complete bounty solution.

## Test plan

      - 794/794 passed

Additional checks run:

- [x] `cmake --build build --target clifft_tests -j1`
- [x] `ctest --test-dir build -R "Exact phase-polynomial" --output-on-failure`
      - 7/7 passed
- [x] stdlib metadata check matching `pass_registry.h` and `docs/macros.py`
- [x] `python3 -m py_compile tests/python/test_exact_phase_polynomial_t_count.py docs/macros.py`
- [x] `python3 -m py_compile tools/bench/exact_phasepoly_tcount.py`
- [x] `python3 tools/bench/exact_phasepoly_tcount.py --help`
- [x] Parser smoke checks for the `.qc` importer and Qiskit
      `time_cnt3-5_180.qasm` through the bounded QASM importer.
- [x] Temporary C++ evaluator over supported Qiskit QASM smoke cases under a
      5000 imported-gate cap; `hwb12.qasm` separately timed out under a
      20-second cap and is documented as excluded.

## Contributor agreement

AI Use Disclosure: Code syntax was assisted by Codex model. All handwritten code and accompanying text were thoroughly reviewed and verified by me prior to submission. I retain full responsibility for the final contribution.


tests/test_exact_phase_polynomial_t_count.cc
```diff
@@ -0,0 +1,220 @@
...
+    REQUIRE(hir.destab_mask(hir.ops[0]) == 0b0011);
+    REQUIRE(hir.stab_mask(hir.ops[0]) == 0b0011);
+    REQUIRE(std::abs(hir.global_weight.real() - test::kInvSqrt2) < 1e-12);
+    REQUIRE(std::abs(hir.global_weight.imag() - test::kInvSqrt2) < 1e-12);
+    REQUIRE(pass.blocks_optimized() == 1);
+    REQUIRE(pass.t_removed() == 15);
+    require_dense_equal(dense_hir_ops_value(hir), dense_hir_ops_value(original));
+}
+
+TEST_CASE("Exact phase-polynomial pass reduces source-level complete CCZ_4 after peephole",
```
bachase, line 211: Should tests like this be checking circuit equivalence too?

src/clifft/optimizer/exact_phase_polynomial_t_count_pass.cc
```diff
@@ -0,0 +1,570 @@
...
+            hir.demote_to_phase_rotation(op, static_cast<double>(coeff) / 4.0);
+        }
+        ++write;
+    }
+
+    const int global_delta =
+        mod8(model.constant_phase + candidate.residual_constant + emitted_constant);
+    hir.global_weight *= eighth_root(global_delta);
+
+    const bool has_source_map = hir.source_map.size() == hir.ops.size();
```
bachase, line 497: This is fine for an MVP, but is it possible to retain a source mapping still? Worth a comment to capture that for the future if so.

tools/bench/exact_phasepoly_tcount.py
```diff
@@ -0,0 +1,414 @@
...
+            line = line[:-1].strip()
+
+        lowered = line.lower()
+        if lowered.startswith("openqasm ") or lowered.startswith("include "):
+            continue
+        if lowered.startswith("creg "):
+            continue
+        if lowered.startswith("gate ") or lowered.startswith("opaque "):
+            raise ValueError(f"{name}:{lineno}: custom QASM gates are not imported")
+        if lowered.startswith("measure ") or lowered.startswith("barrier "):
```
bachase: Rather than ignoring, should it raise like below? I'd worry it silently means something different if not.

src/clifft/optimizer/exact_phase_polynomial_t_count_pass.cc
```diff
@@ -0,0 +1,570 @@
...
+[[nodiscard]] uint32_t first_set_bit(const std::vector<uint64_t>& bits) {
+    for (uint32_t w = 0; w < bits.size(); ++w) {
+        if (bits[w] != 0) {
+            return 64U * w + static_cast<uint32_t>(std::countr_zero(bits[w]));
+        }
+    }
+    throw std::logic_error("first_set_bit called on zero vector");
+}
+
+void xor_bits(std::vector<uint64_t>& dst, const std::vector<uint64_t>& src) {
```
bachase: This pass reimplements several bit-vector primitives already available through MaskView / MutableMaskView (bit_get, is_zero, lowest_bit, xor_with). I understand the pass needs a scratch combined (x,z) coordinate vector, but could we wrap that vector in MaskView/MutableMaskView or introduce one small local adapter instead of duplicating the low-level bit operations?

Not a big priority, but just noting

tests/test_exact_phase_polynomial_t_count.cc
bachase, line 1: Here or in python, would it make sense to add a randomized circuit test to stress test edge cases?

src/clifft/optimizer/exact_phase_polynomial_t_count_pass.cc
```diff
@@ -0,0 +1,570 @@
...
+    basis.push_back(BasisRow{std::move(bits), row_coord, pivot});
+    std::sort(basis.begin(), basis.end(),
+              [](const BasisRow& a, const BasisRow& b) { return a.pivot < b.pivot; });
+    ++rank;
+
+    ok = true;
+    return 1u << (rank - 1);
+}
+
+[[nodiscard]] int pauli_product_phase_mod4(const Axis& left, const Axis& right) {
```
bachase, line 151: Could you walk through one concrete signed Pauli-product example, such as XX * ZZ = -YY, showing how pauli_product_phase_mod4, constant_phase, coefficient flipping, and final global_weight combine to preserve the exact unitary?

Just asking to understand, the answer could just be a comment in github

src/clifft/optimizer/exact_phase_polynomial_t_count_pass.cc
```diff
@@ -0,0 +1,570 @@
...
+        }
+    }
+    return values;
+}
+
+[[nodiscard]] uint8_t degree(uint32_t monomial) {
+    return static_cast<uint8_t>(std::popcount(monomial));
+}
+
+[[nodiscard]] bool clifford_residual_from_truth(
```
bachase, line 296: Could you add a short derivation, either in docs or test comments, for how the Clifford residual reconstruction works? In particular, why does an ANF quadratic coefficient 4*x_i*x_j get emitted as a parity PHASE_ROTATION plus adjusted linear terms, and why is that valid for arbitrary commuting Pauli generators rather than only Z-basis parities?

tests/test_exact_phase_polynomial_t_count.cc
```diff
@@ -0,0 +1,220 @@
...
+    REQUIRE(hir.ops.size() == 6);
+    for (const auto& op : hir.ops) {
+        REQUIRE(op.op_type() == OpType::PHASE_ROTATION);
+        REQUIRE(op.alpha() == 1.5);
+    }
+    REQUIRE(std::abs(hir.global_weight.real()) < 1e-12);
+    REQUIRE(std::abs(hir.global_weight.imag() - 1.0) < 1e-12);
+    REQUIRE(pass.blocks_optimized() == 1);
+    REQUIRE(pass.t_removed() == 15);
+    require_dense_equal(dense_hir_ops_value(hir), dense_hir_ops_value(original));
```
bachase, line 181: Could you add a nearby end-to-end test that runs the rewritten HIR through lower -> execute -> get_statevector and compares against the original circuit/state? That would verify the emitted residual PHASE_ROTATIONs are handled correctly by the backend, not only by the local dense helper.

bachase (commented):
Thanks @DennisWayo  -- this is a clear and well-scoped contribution. I like that you have framed this as an exact bounded phase-polynomial decoder rather than overclaiming TODD/TOHPE/MCR, and the benchmarking approach is really nice.

Given that you aren't seeing much improvements from this pass, my current read is that this is promising as a correctness-oriented prototype or small-block cleanup pass, but probably not as something we'd end up merging (which is a separate decision than awarding the bounty). 

On your question - You are welcome to run the same benchmark harness on more PRs or more circuits, but do you expect to see this approach to work better in other cases that you've not already explored?

DennisWayo:
Thanks for the detailed pass, @bachase. I pushed `d02ee81` to address the inline comments.

On the broader benchmark question: I do not expect this exact bounded decoder to suddenly become a broad real-world T-count reducer just by running more circuits. More circuits are still useful for documenting the boundary, but I expect mostly null results unless another pass first exposes dense small commuting phase-polynomial blocks. I think the honest framing is that this PR is a correctness-oriented small-block decoder/verifier, not a replacement for TODD/TOHPE-style high-rank reduction or MCR-style reordering.

What changed in this update:

- Added `lower -> execute -> get_statevector` equivalence checks for the source-level `CCZ_4` case and for the residual `PHASE_ROTATION` rewrite case.
- Added a deterministic randomized signed commuting-block stress test covering signs, daggers, repeated axes, skipped cases, and at least one rewrite, checked against the dense unitary oracle.
- Changed the QASM benchmark importer so measurements now raise instead of being silently ignored. Barriers remain skipped because they are unitary no-ops for this T-count evaluator.
- Replaced the duplicated scratch bit-vector helper functions with a small local `ScratchBits` adapter, since the scratch vector is the concatenated `(x, z)` coordinate vector rather than a normal HIR mask view.
- Added the source-map comment: emitted ops are synthesized from the whole decoded block, so the current code clears one-to-one source mappings rather than preserving misleading mappings. A future improvement could keep aggregate source ranges.
- Added the signed Pauli-product example and the quadratic Clifford-residual derivation to the theory note.

Local verification after the update:

- `ctest --test-dir build -R "Exact phase-polynomial" --output-on-failure`
- `ctest --test-dir build -E Bench --output-on-failure`
- `env -u OMP_NUM_THREADS UV_CACHE_DIR=/private/tmp/clifft-uv-cache /private/tmp/clifft-uv-bootstrap/bin/uv run pytest tests/python/ -v`
- `env -u OMP_NUM_THREADS UV_CACHE_DIR=/private/tmp/clifft-uv-cache /private/tmp/clifft-uv-bootstrap/bin/uv run pytest --codeblocks docs/ README.md -v`
- `env -u OMP_NUM_THREADS UV_CACHE_DIR=/private/tmp/clifft-uv-cache PRE_COMMIT_HOME=/private/tmp/clifft-pre-commit-cache /private/tmp/clifft-uv-bootstrap/bin/uv run pre-commit run --all-files`

DennisWayo:
> Thanks @DennisWayo -- this is a clear and well-scoped contribution. I like that you have framed this as an exact bounded phase-polynomial decoder rather than overclaiming TODD/TOHPE/MCR, and the benchmarking approach is really nice.
> 
> Given that you aren't seeing much improvements from this pass, my current read is that this is promising as a correctness-oriented prototype or small-block cleanup pass, but probably not as something we'd end up merging (which is a separate decision than awarding the bounty).
> 
> On your question - You are welcome to run the same benchmark harness on more PRs or more circuits, but do you expect to see this approach to work better in other cases that you've not already explored?

I understand the concern @bachase. I do not want to overclaim this as a broad real-world reducer. My intent is to make this PR useful as a bounded exact decoder, correctness oracle, and benchmark harness, while documenting where it does and does not improve T-count. If that is not mergeable as a production optimizer, would it still be useful to keep open as a prototype/reference contribution, or should I treat it as groundwork for a follow-up pass that exposes larger real-world improvements, which I am willing to spend the time to do?

DennisWayo:
I pushed `5360688` with one follow-up experiment based on the review question.

The new piece is `TGateBlockCollectionPass`, an opt-in/default-disabled HIR pre-pass for the exact decoder. It does not rewrite gates or change T-count itself. It only uses existing `can_swap()`-approved adjacent swaps to pull a later `T_GATE` into the current T block, and only when that T gate commutes with every T gate already in the block. The point is to test a narrow hypothesis: whether this exact bounded decoder starts helping when rank-capped commuting blocks are present but separated by safely movable HIR ops.

The important result is negative on the current built-in real-world fixtures. With:

```bash
python tools/bench/exact_phasepoly_tcount.py --collect-t-blocks
```

the repo fixtures still show no additional T-count reduction, and the collector reports zero collected blocks / zero moved T gates on those fixtures. I added those counters to the benchmark table, along with adjacent swaps and exact blocks considered, so this is visible rather than implied. For example, `cultivation_d5.stim` has 5 exact candidate blocks and `circuit_d3_t_gate_p0.001.stim` has 4, but both still optimize 0 blocks. I also checked that increasing the exact rank cap up to 8 did not change the `circuit_d3_t_gate_p0.001.stim` result.

The positive example is intentionally bounded: a rank-4 Reed-Muller zero word separated by commuting phase rotations. Exact alone leaves the 15 T gates because the block is not contiguous; the collector moves 14 T gates into one block; the exact pass then removes all 15, with statevector equivalence checked after lowering/execution. I added separate tests for source-map swap preservation, noncommuting barriers, and a mixed noise/observable side-table regression so the preflight checks only the candidate T against intervening ops, not unrelated adjacent pairs.

So my answer to the earlier question is now more concrete: I do not expect this approach to become a broad real-world reducer just by trying more circuits. I expect it to help only in circuits that contain small dense commuting phase-polynomial structure hidden behind locally movable operations. This commit makes that exposure experiment reproducible, and it documents that the current built-in real-world fixtures do not expose that pattern to this conservative collector/benchmark path. That keeps the distinction from TODD/TOHPE/MCR-style work explicit: **let me reiterate** that this PR is still a bounded exact decoder plus a conservative exposure probe, not a high-rank reducer or a broader reordering algorithm.

Local verification after this update:

- `ctest --test-dir build -R "T-gate block collection|Exact phase-polynomial" --output-on-failure`
- `ctest --test-dir build -E Bench --output-on-failure`
- `env -u OMP_NUM_THREADS UV_CACHE_DIR=/private/tmp/clifft-uv-cache /private/tmp/clifft-uv-bootstrap/bin/uv run pytest tests/python/ -v`
- `env -u OMP_NUM_THREADS UV_CACHE_DIR=/private/tmp/clifft-uv-cache /private/tmp/clifft-uv-bootstrap/bin/uv run pytest --codeblocks docs/ README.md -v`
- `env -u OMP_NUM_THREADS UV_CACHE_DIR=/private/tmp/clifft-uv-cache PRE_COMMIT_HOME=/private/tmp/clifft-pre-commit-cache /private/tmp/clifft-uv-bootstrap/bin/uv run pre-commit run --all-files`

bachase:
Thanks for the continued work here @DennisWayo. I’m going to treat this as one of the finalist submissions for the hackathon issue. I’m not planning another detailed technical review round before the bounty decision, which I will make tomorrow.

For clarity, I’ll be judging the bounty based on the scientific-prototype bar from #40: the chosen approach, correctness evidence, representative T-count results, testing and validation where feasible, and an clear conclusion about the promise and limitations of the approach. If you want to make any final changes against those expectations, feel free, but I’d encourage keeping them focused and easy to review rather than broadening the PR further.

The bounty decision is separate from merge readiness. I’m not expecting to merge any of the current submissions as-is without further cleanup and review, which we can discuss after the hackathon.

PR #165, opened 2026-07-02
feat(optimizer): segmented pass execution with fence barriers (fence spike)

bachase:
> **Prototype / less-strict review.** Targets the `codex/noncomputational-structural-mvp` feature branch, not `main`.

Step 2 of the state-dependent-jumps plan (`design/state-dependent-jumps.md` §6): the **fence de-risk spike** — measure what it costs to make instrument sites structural optimization barriers, before building the machinery that depends on them (steps 4–6).

**Mechanism** (real infrastructure, reused verbatim by step 4 with `is_fence = "is an INSTRUMENT"`): `run_segmented(module, is_fence)` on both pass managers. Fence ops are never handed to a pass; each maximal fence-free segment is presented to the passes as if it were the whole module, with arenas/side-tables/counters staying in place (so mask handles and pool indices remain valid, and per-segment pool appends accumulate). Barrier semantics hold **by construction**, not per-pass discipline. No individual pass is modified; default pipelines are untouched.

**Results** (Release, 40 samples, means; two fence densities bracketing reality — instruments attach per (annotated op, operand), so for layered circuits they cluster at gate positions rather than inside noise runs):

| Workload | Runtime, realistic (noise-run starts) | Runtime, atomized (every noise site) | Compile, atomized | peak_rank |
|---|---|---|---|---|
| surface d7r7, p=1e-3 | **+2.9%** | +25.3% | **−16% (faster)** | 0 → 0 |
| surface d5r5, p=0.05 | **−0.6%** | +7.5% | — | 0 → 0 |
| cultivation d5 | **−1.7%** | +6.6% | −0.9% | 10 → 10 |

Module shape at realistic density is essentially unchanged (d7r7: identical instruction counts, noise blocks intact). The atomized upper bound's entire cost is one identified mechanism: noise-block coalescing destroyed (d7r7: 378 loose `OP_NOISE` + 15 blocks → 2673 loose), i.e. the gap-sampler walking sites one at a time — precisely the regime the design's instrument **hazard-pooling** (§4.4) recovers. Notably, `peak_rank` is unchanged at *both* densities on all workloads: the concern that fencing the squeeze pass would inflate rank did not materialize.

**Verdict: GO** (recorded in the design note §4.2, replacing the "to be measured" sentence). Realistic-density fencing is free against the proposed thresholds (≤5% runtime, ≤2× compile); the worst case is bounded, explained, and already has a designed mitigation.

**Tests** (5 new, non-bench): no-fence `run_segmented` ≡ `run` (identical opcode streams + same-seed samples); fences block T-pair fusion across a commuting noise op; fences block noise-block coalescing; **exact** `record_probabilities` equivalence of fenced vs unfenced compiles (measurement-fenced, noise-free); statistical detector equivalence on a noisy rep-code circuit fenced at every noise site. Bench cases print the module-shape deltas and are `[bench]`-tagged (excluded from default runs; benchmark names fit the bench-history parser's column).

Verified: Debug **984** + Release **984** (`~[bench]`) green; Python **732** green (no binding surface changes); pre-commit `--all-files` clean.

### DIFF
--- a/design/state-dependent-jumps.md
+++ b/design/state-dependent-jumps.md
@@ -202,7 +202,14 @@ considerations:
    recompiled continuation's prefix is bit-identical to the code that already
    executed (§4.3).
 
-The performance cost of fencing is measured before the rest is built (§6).
+The fencing cost was measured (§6 step 2) by segmenting the default
+pipelines on representative workloads (surface d7r7 at p=1e-3, d5r5 at
+p=0.05, cultivation d5). At the realistic density — fences clustered at
+gate positions, modeled as noise-run starts — compile time, sampling
+throughput, and peak rank are unchanged within measurement noise. At the
+atomized upper bound (a fence at every noise site) sampling slows 7-25%,
+dominated by lost noise-block coalescing; that is the regime instrument
+hazard-pooling (§4.4) is designed to recover. Verdict: go.
 
 ### 4.3 Backend: lowering, and frame composition across the trap boundary
 
--- a/src/clifft/optimizer/bytecode_pass.cc
+++ b/src/clifft/optimizer/bytecode_pass.cc
@@ -1,5 +1,8 @@
 #include "clifft/optimizer/bytecode_pass.h"
 
+#include <utility>
+#include <vector>
+
 namespace clifft {
 
 void BytecodePassManager::add_pass(std::unique_ptr<BytecodePass> pass) {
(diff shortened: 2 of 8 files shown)


bachase:
Closing unmerged by design discussion: the go/no-go measurement is complete (verdict: go — numbers in the PR description and recorded in the design note by the follow-up doc PR), and the enforcement mechanism is superseded by the simpler positional-barrier treatment the pipeline already gives EXP_VAL (`can_swap` in commutation.cc + peephole's `blocks()`), which `INSTRUMENT` will join in step 4 — making `run_segmented` measurement scaffolding with no remaining consumer. The branch stays for reference; the barrier tests, the exact record_probabilities equivalence test, and the bench harness get rebuilt against the real op type in step 4.

PR #170, opened 2026-07-02
feat(hir): INSTRUMENT sites, fences, lowering, and dispatch

bachase:
> **Prototype note:** this PR targets the `codex/noncomputational-structural-mvp` feature branch (not `main`) and is part of the in-progress noncomputational MVP; a less strict review bar applies.

Step 4 of the exact state-dependent jumps plan (`design/state-dependent-jumps.md` §6): `OpType::INSTRUMENT` + fences + lowering + offset table, plus the dispatch onto the step-3 kernels. Inert by default — plain `trace()` still rejects annotations; only the opt-in path materializes instruments. Three commits:

## C1 — HIR op, materialization, fences (`f926eb8`)

- **`OpType::INSTRUMENT`**: mask-carrying like `MEASURE` — the mask is the rewound source projector `Z_q` — with an `InstrumentSite` side-table (per-source totals split into computational destinations vs. the leaked/lost trap remainder, damp coefficients, and a rewound-X fixup mask).
- **Opt-in materialization**: `trace(circuit, &options)` turns `LEVEL_TRANSITION`/`LOSS` annotations into sites at their positions (zero-rate sites elide). The front-end stays model-free: `instrument_trace_options(model)` in the noncomp layer resolves a model into plain numbers. The policy grows `damping = exact (default) | neglect`.
- **Fences**: `INSTRUMENT` joins `EXP_VAL`'s positional-barrier clause in `can_swap` (which covers the squeeze pass) and the peephole's commute check, and joins the mask-carrying group in the virtual-S absorption. Pinned: no T-pair fusion or measurement bubbling across a site; conjugation matches measurement masks (X→Y under an absorbed S).

## C2 — backend lowering + offset table (`d2bc2bc`)

Measurement-pattern lowering: virtualize through `V_cum`, localize, classify → **`OP_INSTRUMENT_ACTIVE`** (any axis — instruments don't compact, so no swap-to-top; X-basis pivots take an absorbed array Hadamard like the rotation route), **`OP_INSTRUMENT_DORMANT_STATIC`**, **`OP_INSTRUMENT_EXPAND`** (exact damping, +1 to k at the site), **`OP_INSTRUMENT_DORMANT_NEGLECT`** (k stable). `CompiledModule` gains the `site_id → bytecode offset` table for step-5 re-entry, rebuilt by scan after bytecode passes (fusion shifts indices). Pinned: the **prefix-identity contract** — same prefix against two different suffixes compiles bit-identically up to the instrument's offset, through the full default pipeline.

One design-note refinement recorded in §4.4: at a **neglect-mode dormant-random site every fire traps**, computational destinations included — an in-line collapse would re-anchor the frame conditionally, which compiled downstream code cannot account for (a measurement's anchor is unconditional, so its compile-time virtual-H alignment is sound). The trap's suffix rewrite handles it instead; fires are rare.

## C3 — dispatch + execution tests (`1835a51`)

One shared handler composes the draw-free kernels with the fire draws. A `p = 1` source (`r = 0`) takes the documented eval-only + per-branch-collapse route. Leaked/lost fires throw a trap error naming the site and circuit line until step 5's resume protocol lands (pinned).

**The tests caught a real bug**: the C2 fixup virtualized a *raw* `X_q`, skipping the tableau rewind every other mask gets — any Clifford between preparation and site broke the destination flip (certain-relaxation shots measured 1 instead of 0). The fixup is now a second rewound mask materialized by the front-end, conjugated by the peephole alongside the op mask, and mapped through `V_cum` at lowering. This is exactly the class of bug the deterministic certain-fire tests were designed to catch.

## Deferred (flagged in the plan)

Hazard-pooling of source-independent Bernoullis (throughput, step-7 driven), the known-level p = 0 elision, suffix-only compilation.

**Verified**: Debug **1063** + Release **1063** (`~[bench]`) green; Python **505** green (docs tripwire drove `docs/opcodes.json` entries for the new op type and opcodes); wasm module builds clean in the emsdk image; pre-commit `--all-files` clean.

**Test round** (`20274a8`): four coverage additions from the robustness review. (1) An **end-to-end no-fire oracle**: all fire mass routed to leaked/lost makes every completed shot a certified no-fire shot, and an `EXP_VAL` probe after the site must read the closed form **⟨Z⟩ = 2r/(1+r²)** exactly per shot, through real compilation, on both the active and expand forms — the survivorship-tilt physics validated through the whole pipeline, not just at kernel level. (2) The **fixup-conjugation blind spot** closed from both sides: a structural pin that virtual-S absorption rotates the side-table fixup (in a circuit where the op's own mask commutes and stays put — the side-table twin of the C2 bug), plus a certain-fire execution test through the full default pipeline with a fusable T pair ahead of the site. (3) **Sign, multi-site, and entanglement** dispatch pins: a provably `FLAG_SIGN`ed active site, sequential sites where a drained source must never fire (exercising the dust clamp in production), and a genuinely multi-axis fixup landing on the right qubit. (4) Noise blocks pinned to never coalesce across a site; `record_probabilities` rejects instrument programs. Debug **1071** + Release **1071** green.

**Review fix** (`12cbf11`): the rejection test's name claimed both exact-probability entry points but only exercised `record_probabilities()`; `basis_probabilities()` now gets its own measurement-free instrument program and assertion.


src/clifft/backend/backend.h
```diff
@@ -73,6 +73,16 @@ enum class Opcode : uint8_t {
     OP_MEAS_ACTIVE_INTERFERE_FORCED,
     OP_SWAP_MEAS_INTERFERE_FORCED,
 
+    // Instrument sites (state-dependent jumps). Each variant evaluates the
```
bachase: Should the comment help the reader understand this relates to modeling non-computational levels classically? For someone just scanning this file, it might be  unclear what "state-dependent jumps" means etc?

src/clifft/frontend/hir.h
```diff
@@ -89,6 +89,62 @@ struct ReadoutNoiseEntry {
 /// Index into HirModule::readout_noise side-table
 enum class ReadoutNoiseIdx : uint32_t {};
 
+/// Index into HirModule::instrument_sites side-table. The index doubles as
+/// the site's stable id: instruments are optimization fences, so no pass
```
bachase: Can you confirm if we use fences or barriers in the existing codebase? I want to be consistent.

src/clifft/frontend/hir.h
```diff
@@ -89,6 +89,62 @@ struct ReadoutNoiseEntry {
...
+    // computational level d. The trap (Leaked/Lost) remainder per source
+    // is p_total[s] - p_dest[s][0] - p_dest[s][1].
+    double p_dest[2][2] = {{0.0, 0.0}, {0.0, 0.0}};
+
+    // No-fire damp coefficients sqrt(1 - p_total[s]).
+    double damp[2] = {1.0, 1.0};
+
+    // Under damping="neglect", a dormant-random site skips the expansion
+    // and applies no no-fire back-action (design/state-dependent-jumps.md
+    // section 4.6). Baked in at materialization from the model policy.
```
bachase: Don't have comments that reference design docs, which might get stale or be removed. Have the comments self-contained, or at least only reference other parts of the codebase.

src/clifft/noncomp/policy.h
```diff
@@ -34,6 +34,20 @@ enum class LostLeakedOpsPolicy : uint8_t {
     Drop = 1,
 };
 
+// Damping policy for exact-mode compilation at sites where the no-fire
+// back-action is genuinely non-Clifford (a source-dependent transition on
+// a dormant qubit with a random outcome). Exact expands the qubit into
+// the amplitude array (+1 to the circuit's k at that site) and applies
+// the damp there. Neglect keeps only the site's exact ingredients -- the
+// state-independent fire draw and the on-fire collapse -- and omits the
```
bachase, line 42: This neglect comment might be clearer in the language of kraus operators or more physics. But I want it concise. I think can keep as is and future documentation can spell this out clearly.

src/clifft/svm/svm_kernels.inl
```diff
@@ -1962,6 +1964,166 @@ static inline void exec_readout_noise(SchrodingerState& state, const ConstantPoo
...
+                                                uint8_t source) {
+    throw std::runtime_error("instrument site " + std::to_string(site.site_id) + " (circuit line " +
+                             std::to_string(site.source_line) + ") fired from source " +
+                             std::to_string(source) +
+                             " to a leaked/lost destination; resumable traps are not "
+                             "implemented yet");
+}
+
+// Destination fixup for an in-line computational fire whose destination
+// differs from its source: XOR the site's virtualized X_q into the frame.
```
bachase: Asking to understand -- in this case source and dest are still computational levels right? so this is like 0 jumps to 1 or 1 jumps to 0?

src/clifft/backend/backend.h
bachase: Agreed — reworded in `c15c476`: the group comment now says these are transitions from the noncomputational trajectory model (leakage, loss, relaxation) whose firing probability depends on which computational level the qubit occupies, which is why the fire branch is evaluated at runtime. Same orientation added to the `InstrumentSite` banner in hir.h.

src/clifft/frontend/hir.h
bachase: The codebase precedent is **barrier** (`commutation.cc` "dataflow barrier", `statevector_squeeze_pass.cc` "EXP_VAL acts as a hard barrier", and the clauses this PR added say "positional barrier"). The one stray "fences" was mine, in this line — switched to "barriers" in `c15c476`. "Fence" now only lives in the design doc as the feature name.

src/clifft/frontend/hir.h
bachase: Fixed in `c15c476` — the comment now points at `DampingPolicy` in `noncomp/policy.h` (which carries the full self-contained description) instead of the design doc.

src/clifft/noncomp/policy.h
bachase: Keeping as is per your call. The Kraus-language version lives in `svm_instrument_kernels.h` (the K_stay/K_jump block), so the mkdocs pass can lift from there when we document the knob for users.

src/clifft/svm/svm_kernels.inl
bachase: Yes, exactly: here `dest` is always a computational level — `0` and `1` are the |0>/|1> jumps (so `dest != source` is a 0→1 or 1→0 jump, handled in-line via the frame fixup), and `dest == source` is a jump onto its own level, which is a real event in the model (a pure carrier collapse, e.g. the diagonal of a dephasing-like column) needing no fixup. The `-1` return is the remainder of the column: mass going to leaked/lost levels, which is the resumable-trap path. Sharpened the comment to say this in `c15c476`.

PR #173, opened 2026-07-03
feat(noncomp): exact-mode driver, continuation cache, and correlated neglect

bachase:
> **Prototype note:** this PR targets the `codex/noncomputational-structural-mvp` feature branch (not `main`); a less strict review bar applies.

Step 6 of the exact state-dependent jumps plan (`design/state-dependent-jumps.md` §4.5/§6): **exact mode becomes usable end-to-end, C++ and Python.** `unknown_source_policy="exact"` routes `sample_noncomputational` to the new driver. Three commits.

## C1 — the continuation rewrite (`0b77fe5`)

`rewrite_continuation()` rebuilds the full annotated circuit under a shot's resolved events: prefix verbatim (bit-identical compilation is what re-entry relies on), annotations kept wherever the qubit is still computational (including after a recapture), classical-source consults consumed with host-pre-drawn outcomes, the trapped jump's carrier edits at the suffix start. The **main line is this rewrite with empty events** — which is what carries readout confusion onto kept measurements — and the ordinary-op path is one helper shared with the AOT rewrite, so the two cannot drift. The forced trace-out's hidden **record slot** is reported by mirroring `trace()`'s per-reset-target numbering: slot indices ride in instruction payloads and survive every bytecode pass, so the slot (not an instruction position) is the durable identity, per the review discussion.

## C2 — driver, cache, Python (`718639e`)

The §4.5 loop as amended: shared main line; per-shot host-drawn initials with `|1⟩` preloaded as Pauli-frame bits (no distinct modules; rare noncomputational initials get their own continuation-from-the-top); traps resolve their destination branching on `destination_pending`, extend the events with pre-drawn classical follow-ons, and resume past the site. **Two-level cache**: the rewrite per canonicalized status-delta (computational initial levels canonicalize to zero — the node stream provably depends only on status kinds), then one compiled module per herald-flag assignment (option (d): ternary classifiers ride the key, reusing the existing patch pass). Every host draw precedes cache lookup, so hits and misses sample identically. `max_rank` rejects over-budget compiles naming the first offending circuit line, plumbed to Python along with `unknown_source_policy="exact"` and the `damping` kwarg. Debug builds byte-compare every continuation prefix against the executed module on cache insertion.

## C3 — correlated neglect (`108cac8`)

The forced trace-out, as designed in review: at cache-compile time the trace-out's hidden measurement swaps to its **forced opcode twin** at the reported slot (`record_probabilities` precedent), and the driver threads the reported source through `state.forced_record` per resume — per-shot runtime data, so the module stays source-independent and the key needs nothing new. The prefix-identity comparison treats the sampling↔forced twin as equal, the one sanctioned prefix difference in a chain.

**Activating it surfaced a real ordering hazard, caught deterministically by the new correlation test**: the squeeze pass reorders *commuting* measurements — exchangeable under sampling semantics, wrong once one is forced. An entangled partner slid between the site and the trace-out and measured the uncollapsed carrier. Exact-mode modules therefore compile with the default pipeline **minus `StatevectorSqueezePass`**, uniformly (prefix identity preserved); the rank-compaction loss is step 7's performance table to quantify, and restoring squeeze via a pinned-measurement barrier flag is a known follow-up. The decisive pin: a Bell-entangled dormant-random qubit with certain source-dependent leak destinations — the classified slot reads back the source, and the partner must equal it on **every** shot (independent redraw would mismatch half the time).

## Tests

17 new cases across three files: continuation-rewrite units (verbatim main line, no X-prep, trace-out placement, classical-consult validation, forced-slot numbering); driver end-to-end (certain-leak trap+classify+sidecar, frame preload, noncomputational initials, seepage recapture, exact-vs-AOT agreement at source-independent rates, seed reproducibility, `max_rank` with line naming, ternary heralds through the cache key, the neglect correlation pin, neglect-vs-exact rank contrast under `max_rank=0`).

**Verified**: Debug **1101** + Release **1101** (`~[bench]`) green; Python **505** green plus an exact-mode smoke through the real wheel; pre-commit clean. (No wasm-list change: the noncomp sources aren't in the wasm build.)

**Deferred, flagged**: hazard-pooling, suffix-only compilation, LRU capping, trapped-shot batching (all step-7-arbitrated); squeeze restoration for exact mode via a pinned-measurement flag; runtime herald state as the clean successor to flags-in-key.

**Review fix** (see latest commit): **[P1]** the classical-outcome stream is now rebuilt by annotation target on every extension instead of appended — a trap can turn the trapped qubit's later consults classical *between* previously recorded positions, and the old cursor replay fed outcomes to the wrong targets. Reused outcomes keep the executed prefix stable; first-seen consults draw fresh (unbiased: they live only in the never-executed suffix). The reviewer's repro is the regression test, plus two same-class chain pins: a **two-forced-trap chain** (two forced slots per shot; the first forced instruction sits in the second continuation's prefix, exercising the sampling/forced mask) and a **neglect fire onto a computational destination** (forced materialization anti-correlates the trapped qubit with its Bell partner under a certain source-swap channel). Debug + Release **1104** green.

**Audit round** (see latest commit): the four remaining cases from the trap-chain consistency audit, all deterministic — **herald-flag reuse** across continuations (p_herald = ½ paired with a deterministic not-heralded bit makes an independent redraw break a per-shot invariant), **exactly-once noise** between two traps across both re-anchors, a **detector + observable spanning the trap boundary** (main-line Born bit vs continuation classified bit), and a **hand-written multi-target annotation** trapping on one target with the sibling's instrument surviving the split. Debug + Release **1108** green.

**Fast-math round** (see latest commits): **[P1]** the Release-only NaN/Inf rejection failures on macOS were tracked to the test *inputs*, not the validation: `-ffast-math` makes a nonfinite floating-point constant in an expression undefined behavior (Clang's `-Wnan-infinity-disabled`), and AppleClang 17 folds such constants to arbitrary finite values, so the tests handed `from_matrix` a perfectly valid matrix. GCC 13 / Clang 18 on Linux preserve the bits, which is why CI and local Release runs stayed green. New `opaque_nan()`/`opaque_infinity()` helpers in `test_helpers.h` assemble the IEEE 754 bit pattern behind a `volatile` at runtime — the write-side mirror of `is_finite_robust()` — and every nonfinite construction site now uses them, including the model initial-state NaN test that passed on AppleClang only by accident (a folded entry still trips the sum-to-1 check). Library validation is untouched: runtime nonfinite values, Python's included, were always rejected. Out of scope but noted: `test_svm.cc` has 20 `CHECK(std::isfinite(...))` assertions that fold to vacuously-true under any fast-math Release (they can never go red, and Debug CI checks them for real; filed as #174), and vendored stim triggers the same warning class internally. **[P3]** the `forced_traceout_slot` comment now says what the slot points at for both destination kinds — trace-out R or materializing R — with SIZE_MAX meaning forcing wasn't requested. Debug + Release **1108** green; the five hardened tests also pass a clang-18 `-ffast-math` Release build, which no longer emits the warning for them.

**Review round 2** (see latest commits): the inline-comment round, applied. **Structural**: `draw_from_column` now sums the admitted entries' mass itself — a caller can no longer hand in a mismatched normalizer that would silently bias the draw, and the remainder-restricted call site loses its duplicated loop. The exact pipelines filter on a new `RecordOrder` declaration every registered pass must carry (`PassInfo` field with no default constructor — an unclassified new pass is a compile error, not a silent exact-mode breakage), with a Debug tripwire asserting the measurement-record sequence survives the HIR pipeline of every compiled module, so even a *misdeclared* pass fails loudly. Seed derivation and all five domain tags now live in one `seed.h` (the AOT/exact disjointness is visible in one place; the duplicated SplitMix64 finalizer is gone). **Naming**: "host" → "driver" everywhere (stream, tag, RNG) — "host" is reserved territory once a GPU backend introduces a host/device split. **Docs**: a vocabulary block at the top of `exact_driver.cc` defines the driver's working terms once (annotation/target/consult/site/trap/carrier/continuation/main line/forced twin/sidecar/driver draw); the header now leads with what the driver is for and says explicitly that neglect damping also runs through it; `sampler.h`'s misplaced doc block is fixed. Debug + Release + clang-18 Release all **1108** green, Python **734** green.

**Audit fixes** (see latest commits): the design-audit findings, applied. **[1]** `ClassicalOutcome` now records the source level it was drawn at, validated on every reuse (driver) and consumption (`rewrite_continuation`) — reuse-by-target is sound today only because a qubit's noncomputational level moves solely through its own consults, and this makes that invariant explicit and loud instead of implicit; a future cross-qubit transition that breaks it throws rather than silently replaying outcomes from the wrong column. **[2]** cache keys count-prefix their variable-length sections: equal keys now mean equal events by construction, not by the improbability of marker-aligned collisions. **[3]** `draw_from_column` throws on an empty admitted column instead of falling through an assert to a Release draw of level 255. **[4]** malformed `LOSS` nodes (wrong arity, nonfinite or out-of-range probability) are rejected by a shared `loss_probability()` helper in both sampling paths instead of silently defaulting to a no-op — the parser guarantees the shape, this covers hand-built circuits. **[5]** the herald convention has one home: `classifier.h` states the positional contract, `MeasurementClassifier::kHeraldSymbol`/`has_herald()` replace the hardcoded `prob(2, …)` and `num_symbols() == 3` at all six call sites. The `memcmp`-padding observation from the audit stands as a note (no change; `Instruction`'s explicit pads + factory zero-init carry it). New tests: source-level-mismatch reject, LOSS arity + range rejects. Debug + Release + clang-18 Release **1110** green, Python **734** green.

**Review round 3** (see latest commit): **[P2]** closed — the previous round's LOSS validation covered classical consults and traps but not *live* quantum-source sites, which ride through the continuation rewrite verbatim into `trace()`'s silent default-to-zero. The exact driver now resolves every annotation up front (tag resolution + LOSS shape), so malformed nodes fail deterministically at entry instead of on whichever shot first consults them — this also makes unknown-tag errors deterministic rather than shot-dependent. `trace()` itself now rejects a LOSS without exactly one argument (minimal arity check; the frontend can't reach the shared noncomp helper without inverting layering). Negative tests land where the reviewer scoped them: the exact public entry point (missing arg + `{7.0}`, two sections of one case) and direct instrument tracing; the AOT public entry is covered transitively (it routes through `sample_history`, which validates via the shared helper). **Terminology**: the vocabulary block shrinks to the invariant-encoding terms (annotation, annotation target, consult, fire, jump, site, trap, carrier, continuation) — "annotation target" spelled out (bare `Target` is the AST type), "fire" = the stochastic event vs "jump" = the stored `ResolvedJump`, `forced_twin()` renamed `forced_opcode()`, and main line / sidecar / driver-draw slots dropped (defined inline where used). The `ExactShotEvents` comment now states the key deliberately omits the derived `source_level`. Debug + Release + clang-18 Release **1112** green, Python **734** green.

### DIFF
--- a/src/clifft/noncomp/exact_driver.cc
+++ b/src/clifft/noncomp/exact_driver.cc
@@ -2,47 +2,44 @@
 // design/state-dependent-jumps.md holds the full design. The vocabulary
 // used throughout this file:
 //
-//   annotation   A LEVEL_TRANSITION[name] or LOSS(p) node in the user's
-//                circuit: the level-transition channel a physical
-//                operation applies to each of its qubits.
-//   target       One (op_index, qubit) pair of an annotation; a node
-//                with n qubit operands is n targets.
-//   consult      One target's per-shot draw against its channel: did it
-//                fire, and to which destination level. A *classical*
-//                consult is one whose qubit's level is definite at that
-//                point (known computational, leaked, or lost), so the
-//                driver draws it without touching quantum state.
-//   fire / jump  A consult that moved the qubit to another level.
-//   site         A target kept as a runtime INSTRUMENT barrier in the
-//                compiled module, consulted in-line by the VM;
-//                trace()'s materialization order assigns site ids.
-//   trap         A fire the VM cannot resolve in-line -- the qubit's
-//                level is not definite there -- so execution stops at
-//                the site and control returns to the driver.
-//   carrier      The simulated qubit cell that keeps holding a
-//                noncomputational population's correlations while the
-//                status ledger tracks its level classically.
-//   continuation The circuit rewritten under a shot's events (initial
-//                statuses, jumps so far, pre-drawn classical outcomes)
-//                and recompiled whole; execution resumes past the
-//                trapped site, the prefix guaranteed identical to what
-//                the shot already ran.
-//   main line    The continuation of "nothing happened": every shot
-//                starts in it unless an initial level is
-//                noncomputational.
-//   forced twin  The forced-outcome variant of a sampling measurement
-//                opcode: same record slot, outcome read from
-//                state.forced_record instead of drawn. Collapses a
-//                neglect-form trap's carrier to the source the trap
-//                reported.
-//   sidecar      The per-shot herald record: one flag per visible
-//                measurement slot, set when a ternary classifier drew
-//                its "ambiguous" symbol for that slot.
-//   driver draw  Randomness this file consumes between VM runs
-//                (initial levels, trap destinations, classical
-//                consults, herald flags): one stream per shot, every
-//                draw ordered before any cache lookup. The VM's own
-//                Born randomness is the separate SVM stream (seed.h).
+//   annotation         A LEVEL_TRANSITION[name] or LOSS(p) node in the
+//                      user's circuit: the level-transition channel a
+//                      physical operation applies to each of its qubits.
+//   annotation target  One (op_index, qubit) pair of an annotation; a
+//                      node with n qubit operands is n annotation
+//                      targets. (Bare "Target" is the AST operand type.)
+//   consult            One annotation target's per-shot draw against its
+//                      channel: did it fire, and to which destination
+//                      level. A *classical* consult is one whose qubit's
+//                      level is definite at that point (known
+//                      computational, leaked, or lost), so the driver
+//                      draws it without touching quantum state.
+//   fire               A consult that moves the qubit to another level:
+//                      the stochastic event itself.
+//   jump               A recorded fire -- the ResolvedJump entry in a
+//                      shot's events.
+//   site               An annotation target kept as a runtime INSTRUMENT
+//                      barrier in the compiled module, consulted in-line
+//                      by the VM; trace()'s materialization order
+//                      assigns site ids.
+//   trap               A fire the VM cannot resolve in-line -- the
+//                      qubit's level is not definite there -- so
+//                      execution stops at the site and control returns
+//                      to the driver.
+//   carrier            The simulated qubit cell that keeps holding a
+//                      noncomputational population's correlations while
+//                      the status ledger tracks its level classically.
+//   continuation       The circuit rewritten under a shot's events
+//                      (initial statuses, jumps so far, pre-drawn
+//                      classical outcomes) and recompiled whole;
+//                      execution resumes past the trapped site, the
+//                      prefix guaranteed identical to what the shot
+//                      already ran.
+//
+// The driver's own randomness (initial levels, trap destinations,
+// classical consults, herald flags) is one per-shot stream, every draw
+// ordered before any cache lookup; the VM's Born randomness is the
+// separate SVM stream (seed.h).
 
 #include "clifft/noncomp/exact_driver.h"
 
@@ -392,9 +389,10 @@ std::vector<uint32_t> record_sequence(const HirModule& hir) {
 }
 #endif
 
-// The forced-outcome twin of a sampling measurement opcode, or
-// NUM_OPCODES when the opcode is not a sampling measurement.
-Opcode forced_twin(Opcode op) {
+// The forced-outcome variant of a sampling measurement opcode -- same
+// record slot, outcome read from state.forced_record instead of drawn
+// -- or NUM_OPCODES when the opcode is not a sampling measurement.
+Opcode forced_opcode(Opcode op) {
     switch (op) {
         case Opcode::OP_MEAS_DORMANT_STATIC:
             return Opcode::OP_MEAS_DORMANT_STATIC_FORCED;
@@ -411,7 +409,8 @@ Opcode forced_twin(Opcode op) {
     }
 }
 
-// Swap the trace-out's hidden measurement at `slot` to its forced twin,
+// Swap the trace-out's hidden measurement at `slot` to its forced
+// opcode,
 // so resume() collapses the trapped carrier to the source the trap
 // reported (read from state.forced_record[slot]) instead of redrawing.
 // Slot indices ride inside instruction payloads and are never renumbered
@@ -419,12 +418,12 @@ Opcode forced_twin(Opcode op) {
 void swap_traceout_to_forced(CompiledModule& module, size_t slot) {
     size_t found = 0;
     for (Instruction& instr : module.bytecode) {
-        const Opcode twin = forced_twin(instr.opcode);
-        if (twin == Opcode::NUM_OPCODES) {
+        const Opcode forced = forced_opcode(instr.opcode);
+        if (forced == Opcode::NUM_OPCODES) {
             continue;
         }
         if (instr.classical.classical_idx == slot) {
-            instr.opcode = twin;
+            instr.opcode = forced;
             ++found;
         }
     }
@@ -442,7 +441,7 @@ bool equal_modulo_forced_swap(const Instruction& fresh, const Instruction& execu
     if (std::memcmp(&fresh, &executed, sizeof(Instruction)) == 0) {
         return true;
     }
-    if (forced_twin(fresh.opcode) != executed.opcode) {
+    if (forced_opcode(fresh.opcode) != executed.opcode) {
         return false;
     }
     Instruction swapped = fresh;
@@ -471,6 +470,20 @@ NonComputationalSample sample_noncomputational_exact(const Circuit& circuit,
     const bool ternary = classifier != nullptr && classifier->has_herald();
 
     const Circuit annotated = annotate(circuit, model);
+
+    // Validate every annotation up front -- tag resolution, LOSS shape --
+    // so a malformed node fails here, deterministically, rather than on
+    // whichever shot first traps or consults it classically: a live site
+    // rides through the rewrite verbatim, and trace() must never be the
+    // first to look at its arguments.
+    for (uint32_t op_index = 0; op_index < static_cast<uint32_t>(annotated.nodes.size());
+         ++op_index) {
+        const AstNode& node = annotated.nodes[op_index];
+        if (node.gate == GateType::LEVEL_TRANSITION || node.gate == GateType::LOSS) {
+            resolve_annotation(node, model, op_index);
+        }
+    }
+
     const InstrumentTraceOptions instrument_options = instrument_trace_options(model);
 
     std::map<std::string, ContinuationEntry> cache;
--- a/src/clifft/noncomp/exact_driver.cc
+++ b/src/clifft/noncomp/exact_driver.cc
@@ -10,6 +10,7 @@
 #include "clifft/noncomp/status_step.h"
 #include "clifft/noncomp/transition_instrument.h"
 #include "clifft/optimizer/pass_factory.h"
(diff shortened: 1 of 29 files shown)


src/clifft/noncomp/exact_driver.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+// Exact-mode driver: the runtime half of exact state-dependent jumps.
```
bachase: I'd have this comment be a bit more standalone-self-contained using 

So more like

"Driver for exact-mode simulation for simulating leak/loss semi-classically. Transitions outside the computational basis must be handled at sample time as some jump transitions will be state (ie. shot) dependnet). .. Under Unkoinw ... (etc.)

src/clifft/noncomp/exact_driver.h
```diff
@@ -0,0 +1,29 @@
...
+//
+// Under UnknownSourcePolicy::Exact, transition firing moves into the VM:
+// the annotated circuit compiles once as a shared main line with every
+// annotation materialized as an instrument site, each shot preloads its
+// sampled initial levels into the Pauli frame and executes, and a fire
+// that cannot resolve in-line traps back here. The driver resolves the
+// destination, extends the shot's event record, fetches or compiles the
+// continuation (cached by the status-outcome delta plus herald flags),
+// and resumes past the site. sample_noncomputational() routes here; this
+// header is internal.
```
bachase: Is this driver still used with the neglect approximation?

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+namespace clifft {
+
+namespace {
+
+// Domain tags for the per-shot sub-seeds, disjoint from the AOT
+// orchestrator's history/classifier/SVM tags so the two modes never
+// correlate. One host stream serves every host-side draw in a shot
+// (initial levels, trap destinations, classical follow-ons, herald
+// flags), in a deterministic order that never depends on cache state.
+constexpr uint64_t kExactHostDomain = 0x11;
```
bachase: What does host versus svm domain mean here? Are these new terms?

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace clifft {
+
+namespace {
+
+// Domain tags for the per-shot sub-seeds, disjoint from the AOT
+// orchestrator's history/classifier/SVM tags so the two modes never
```
bachase: Asking to understand - which AOT orchestrator tags is this comment refererencing?

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+// correlate. One host stream serves every host-side draw in a shot
+// (initial levels, trap destinations, classical follow-ons, herald
+// flags), in a deterministic order that never depends on cache state.
+constexpr uint64_t kExactHostDomain = 0x11;
+constexpr uint64_t kExactSvmDomain = 0x12;
+
+uint64_t derive_exact_seed(uint64_t global, uint64_t shot, uint64_t domain) {
```
bachase: 1. To confirm, this is purely for reproducibility & testing, in most cases, users would not be using a seed?
2. Does this impact the randomness properties of RNG by seeding in this way?

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+                                    " does not name a transition in the model");
+    }
+    return channel;
+}
+
+// Draw a destination level from `column` restricted to `mass` (the sum of
+// the admitted entries), with the measurement kernels' last-positive
+// fallback for the floating-point tail. `admit` filters levels.
+template <typename Admit>
+uint8_t draw_from_column(const TransitionInstrument& instrument, uint8_t source, double mass,
```
bachase: How doe we know mass properly sums to admitted entries? I don't want to overly complicated, but is this a bad design in that it allows impermissible states to be expressed in some calls?

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+                                    "] at op " + std::to_string(op_index) +
+                                    " does not name a transition in the model");
+    }
+    return channel;
+}
+
+// Draw a destination level from `column` restricted to `mass` (the sum of
+// the admitted entries), with the measurement kernels' last-positive
+// fallback for the floating-point tail. `admit` filters levels.
+template <typename Admit>
```
bachase, line 113: The comments should briefly mention the contract/signature of `admit`

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+        acc += p;
+        if (u < acc) {
+            return to;
+        }
+    }
+    assert(last_positive >= 0 && "destination draw over an empty column");
+    return static_cast<uint8_t>(last_positive);
+}
+
+// Walk the annotated circuit's statuses under `events` and rebuild the
```
bachase, line 150: This comment again introduces some new terminology -- "consult" "annotation target" -- is there a single comment spot that gives an overview of these terms, definitions and how they relate in this driver?

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+            std::string_view(info.name) != "StatevectorSqueezePass") {
+            pm.add_pass(info.make_hir());
+        }
+    }
+    return pm;
+}
+
+// The forced-outcome twin of a sampling measurement opcode, or
+// NUM_OPCODES when the opcode is not a sampling measurement.
+Opcode forced_twin(Opcode op) {
```
bachase: Similarly -- new concept of "twin"

src/clifft/noncomp/exact_driver.cc
```diff
@@ -0,0 +1,636 @@
...
+// trace-out is forced -- an entangled partner must not measure before
+// the forced collapse it is correlated with. One pipeline serves every
+// module in the mode, so prefix identity is preserved; the squeezed
+// rank compaction is a performance loss the validation campaign's
+// table will quantify.
+HirPassManager exact_hir_pass_manager() {
+    HirPassManager pm;
+    for (const auto& info : kRegisteredPasses) {
+        if (info.kind == PassKind::HIR && info.default_enabled &&
+            std::string_view(info.name) != "StatevectorSqueezePass") {
```
bachase: I think this is ok fo rnow, but this is brittle under change. What if a new default pass is created that is also incompatible with this mode? Wouldn't it silently break things?

Is there a better classification strategy that will be more robust to evolving passes?

src/clifft/noncomp/sampler.h
```diff
@@ -37,6 +38,13 @@ struct HistorySample {
 // transition firing on a ComputationalUnknown qubit, naming the operation,
 // qubit, and gate. Under EqualizeRates that case is sampled with the
 // equalized-rates approximation instead (see policy.h).
+// Draw one qubit's initial level from the model's shared initial-state
```
bachase, line 35: This comment seems to bleed into the existing one above?

PR #190, opened 2026-07-07
feat(noncomp): reject unsupported circuit/model pairs before sampling

bachase:

## Summary

Unsupported circuit/model pairs previously failed **seed-dependently mid-sampling**: an `MX`/`MY`/`MPP` whose operand the model can leak or lose rejects only on whichever shot first realizes the noncomputational status, and a classified measurement without a classifier likewise — at low rates, `sample(shots=N)` succeeded or raised by seed and discarded all completed shots.

- New `validate_static(annotated, model)` (src/clifft/noncomp/static_check.{h,cc}), wired into the driver's existing up-front validation: an abstract interpretation over per-qubit **sets of reachable statuses**. The design rule that keeps it honest: for ordinary nodes it loops each set member through the *same* primitives the concrete walk uses (`qubit_operands`, `operand_action`, `normal_post_op_status`) — there is no third encoding of the policy to drift. Annotations grow the sets by their reachable destinations (a zero-fire channel grows nothing, mirroring trace()'s elision); resets restore exactly the members the policy restores.
- Any member that would reject at runtime now rejects **before shot 0**, with the op index, qubit, gate, and the reachable status named ("can meet a leak_e qubit under this model and is not representable; rejecting before sampling"). The rewriter's lazy checks stay untouched as the guard for direct `rewrite_continuation` callers.
- Precision is pinned as hard as rejection: `S; R; MX` with a certain leak is **accepted** (leaked always restores), `LOSS; R; M` without a classifier rejects or samples exactly by `reset_restores_lost`, zero-fire channels don't poison reachability, and spectator qubits are untouched.
- `sample()`'s docstring stops over-promising: `max_rank` is enforced per continuation compile (not "before any state allocation"); contract violations are what fail before sampling begins.

One deliberate behavior change: pairs that previously *sometimes* sampled on lucky seeds now always reject — a sampler whose success depends on the seed was the footgun this removes.

## Validation

- C++ Debug and Release: full suites (`~[bench]`) pass — 1055 cases (+7: 3 rejection + 4 precision)
- Python: 890 pass on both wheels (+2)
- Red/green: with the driver wiring commented out, the 3 rejection cases and the no-restore sub-case fail (the low-rate `MX` circuit samples cleanly on seed 1 — the exact seed-dependence being removed); precision cases pass either way
- Existing reject-path pins (C++ and Python) pass unchanged — the new messages keep their matched phrases
- Draw-stream fingerprints byte-identical to the base tip across all accepted configs (the pre-scan draws nothing)

## Review round — no-event-branch retirement

P1 (valid): the walk seeded `Computational` unconditionally and never retired a member whose jump away is certain, falsely rejecting a pure-`leak_e` initial whose transition recaptures to `g` with probability 1 before an `MX`. Both legs fixed under one uniform rule — **a member survives a stochastic channel only when its no-event branch is reachable**: the initial set carries `Computational` only when the model gives g/e mass; a transition replaces each member by its destinations, retaining the source only when its column leaves no-fire mass (`column_sum < 1`, exact after construction's clamping); `LOSS(p)` retains computational/leaked members only when `p < 1`. The reviewer's reproducer now samples identically to base, and its over-removal dual (a second `S` re-leaks from the computational columns, so `MX` must still reject) is pinned at both layers. Suites: C++ 1057×2, Python 892×2, streams identical, pre-commit clean.

### DIFF
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -38,6 +38,7 @@ add_library(clifft_core STATIC
     noncomp/status_step.cc
     noncomp/op_role.cc
     noncomp/rewriter.cc
+    noncomp/static_check.cc
     noncomp/exact_driver.cc
     noncomp/orchestrator.cc
 )
--- a/src/clifft/noncomp/exact_driver.cc
+++ b/src/clifft/noncomp/exact_driver.cc
@@ -50,6 +50,7 @@
 #include "clifft/noncomp/op_role.h"
 #include "clifft/noncomp/rewriter.h"
 #include "clifft/noncomp/seed.h"
+#include "clifft/noncomp/static_check.h"
 #include "clifft/noncomp/status_step.h"
 #include "clifft/noncomp/transition_instrument.h"
 #include "clifft/optimizer/hir_pass_manager.h"
@@ -475,6 +476,13 @@ NonComputationalSample sample_noncomputational_exact(const Circuit& circuit,
             resolve_annotation(node, model, op_index);
         }
     }
+
+    // Static pre-scan: reject circuit/model pairs that can never sample
+    // successfully because some reachable qubit status meets an
+    // unrepresentable operation.  This turns seed-dependent mid-sampling
+    // failures (only the shot that first hits the bad status raises) into
+    // a deterministic up-front rejection.
+    validate_static(annotated, model);
 
     const InstrumentTraceOptions instrument_options = instrument_trace_options(model);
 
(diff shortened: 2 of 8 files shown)


bachase:
Closing unmerged, superseded by the contract-gate approach: review established that the reachability scan cannot be both cheap and complete (quantum-state reachability is the true boundary), and the deciding analysis showed the reject class itself was the problem — with single-qubit X/Y readout classified on vacated carriers plus two coarse up-front contract gates (classifier required when a leak/loss-capable model measures; parity measurements unsupported with such models), no seed-dependent reject remains reachable and no reachability machinery is needed. Replacement PR follows.

PR #191, opened 2026-07-07
feat(noncomp)!: classify single-qubit X/Y readout on vacated carriers

bachase:

Supersedes #190 (closed unmerged). Review of that PR established that a static reachability scan cannot be both cheap and complete — and that the reject class it policed was the real problem.

## Summary

- **Non-reset `MX`/`MY` on a vacated (leaked/lost) qubit now classify, exactly like `M`**: record bit from the classifier column, with ternary herald columns and record-target inversion (`MX !0`) via the shared classified-measurement mechanics; status unchanged. The justification is the design's own basis-agnosticity argument, previously applied only to `MRX`/`MRY`: a vacated carrier has no basis for the readout to disagree with. Z-basis computational readout-confusion stays `M`/`MR`-only; computational-qubit `MX` behavior is untouched (pinned).
- This unblocks the motivating category: stim-style memory-X circuits (`RX`/`MX` data readout) now run under leakage models end to end (smoke-tested), where previously they failed on whichever shot first realized a leak.
- **The unsupported remainder becomes two explicit, deliberately coarse contracts** checked up front in the driver (no reachability analysis, no per-qubit tracking):
  - a leak/loss-**capable** model (noncomputational initial mass, any `LOSS(p>0)`, or any transition entry from a computational source into a noncomputational level) that **measures anything** requires a classifier;
  - `MPP` is **not supported** under capable models — the error names the op and points at the supported route: expand the parity readout into an explicit ancilla circuit, whose ladder gates drop on vacated operands and yield the survivors' parity (also the documented future semantics if built-in support is ever wanted).
  - Bluntness is the contract and is pinned by test: a model that leaks only unmeasured qubits still requires a classifier. The lazy rewriter checks remain as defense-in-depth for direct callers.
- With classification plus these gates, **no seed-dependent reject remains reachable through the public API** — the footgun that motivated the reachability scan is gone without the scan.
- Docs: §5.2 table row + notes rewritten (classify rule, `MPP` capability boundary, classifier contract); `sample()`/`Model` docstrings updated, including the honest `max_rank` wording (enforced per continuation compile).

## Validation

- C++ Debug and Release: full suites (`~[bench]`) pass — 1060 cases (+12: MX/MY/inverted/herald/computational-untouched/memory-X pins, both gates positive and negative, bluntness pin)
- Python: 893 pass on both wheels (+5)
- Red/green: with only the `operand_action` change reverted, the new MX/MY tests fail with the old "not representable" rejection; restored, all green
- Draw-stream fingerprints byte-identical to the base tip across all baseline configs (none use `MX`/`MY`/`MPP`)

## Review round

- The contract helper's comment is condensed to the two current restrictions, per review wording.
- `shots=0` previously returned before validation ran, so it accepted pairs that `shots=1` rejects (reviewer's reproducer). The zero-shot early return now sits after the up-front validation — validation is shot-count independent, and `shots=0` doubles as a dry-run contract check returning empty results. Pinned at both layers (the reproducer shape in Python; C++ also pins the valid-pair empty return). Suites: C++ 1061×2, Python 894×2, streams identical, pre-commit clean.

### DIFF
--- a/src/clifft/noncomp/exact_driver.cc
+++ b/src/clifft/noncomp/exact_driver.cc
@@ -442,6 +442,96 @@ bool equal_modulo_forced_swap(const Instruction& fresh, const Instruction& execu
     return std::memcmp(&swapped, &executed, sizeof(Instruction)) == 0;
 }
 
+// Up-front capability contract for the annotated circuit and model.
+//
+// This is a capability boundary, not a reachability analysis: a model that
+// can leak or lose qubits triggers both gate checks regardless of whether
+// any vacated qubit actually reaches a measurement in a given circuit. The
+// deliberate bluntness is a feature — it catches mismatches early and keeps
+// the contract independent of per-shot randomness.
+//
+// Gate A: parity measurements (MPP — MXX/MYY/MZZ desugar to MPP at parse
+// time) span multiple qubits and have no faithful single-bit classifier
+// substitution, so they are rejected up front when the model is capable.
+//
+// Gate B: a model that can leak or lose qubits requires a classifier when
+// the circuit has any measurements, because the classifier is the only
+// defined mapping from a noncomputational level to a record bit.
+void validate_model_contract(const Circuit& annotated, const NonComputationalModel& model) {
+    // Determine whether the model can ever produce a noncomputational qubit.
+    bool noncomp_capable = false;
+    for (const Level l : kAllLevels) {
+        if (category(l) != LevelCategory::Computational && model.initial_probability(l) > 0.0) {
+            noncomp_capable = true;
+            break;
+        }
+    }
+    if (!noncomp_capable) {
+        for (uint32_t i = 0; i < static_cast<uint32_t>(annotated.nodes.size()); ++i) {
+            const AstNode& node = annotated.nodes[i];
+            if (node.gate == GateType::LOSS && !node.args.empty() && node.args[0] > 0.0) {
+                noncomp_capable = true;
+                break;
+            }
+            if (node.gate == GateType::LEVEL_TRANSITION) {
+                const TransitionInstrument* instr = model.transition_named(node.tag);
+                if (instr == nullptr) {
+                    // Unresolvable tags were rejected by the annotation
+                    // validation loop that runs before this helper; the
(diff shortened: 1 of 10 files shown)


src/clifft/noncomp/exact_driver.cc
```diff
@@ -442,6 +442,96 @@ bool equal_modulo_forced_swap(const Instruction& fresh, const Instruction& execu
     return std::memcmp(&swapped, &executed, sizeof(Instruction)) == 0;
 }
 
+// Up-front capability contract for the annotated circuit and model.
```
bachase: This comment is too long and obtuse. You can be much clearer:

" Validate that the circuit is compatible with the current leak/lose noise model and semantics. The curren restrictions are:

-  parity measurements (MPP) and others are not supported and such circuits will be rejected if any MPP target could potentially reach a noncomputational state.
- A classifier is required if the circuit has any measurements

src/clifft/noncomp/exact_driver.cc
```diff
@@ -442,6 +442,96 @@ bool equal_modulo_forced_swap(const Instruction& fresh, const Instruction& execu
...
+// the contract independent of per-shot randomness.
+//
+// Gate A: parity measurements (MPP — MXX/MYY/MZZ desugar to MPP at parse
+// time) span multiple qubits and have no faithful single-bit classifier
+// substitution, so they are rejected up front when the model is capable.
+//
+// Gate B: a model that can leak or lose qubits requires a classifier when
+// the circuit has any measurements, because the classifier is the only
+// defined mapping from a noncomputational level to a record bit.
+void validate_model_contract(const Circuit& annotated, const NonComputationalModel& model) {
```
bachase, line 451: Is this validation run if shots is 0?

Consider
```
from clifft import noncomp
G, E, LG, LE, L = range(5)
m = [[0.0] * 5 for _ in range(5)]
m[L][G] = 1.0
m[L][E] = 1.0
model = noncomp.Model(initial_state=[1,0,0,0,0], transitions={"S": m})
noncomp.sample("S 0\nM 0", model, shots=0, seed=1)  # returns; shots=1 raises
```

PR #197, opened 2026-07-08
fix(noncomp): drop the reset half of a non-restoring lost measure-and-reset

bachase:

## Summary

Second of the two external-review PRs (follows #196; the stream-changing fix, isolated so its evidence stands alone).

The rewriter's classified path emitted a measure-and-reset's reset half unconditionally — contradicting the status stepper's own stated rule that a non-restoring lost-qubit reset acts on a vacated site and drops. The emitted reset was unobservable dead work on `MR` (the carrier is never read while lost), but the X/Y-basis forms (`MRX`/`MRY`) lower through a hidden X/Y measurement that **spends an SVM draw** on the vacated carrier, shifting every later outcome on the shot's stream for a semantically inert node.

The fix reads the verdict off the stepped status itself — the reset is emitted iff the stepper restored the qubit (leaked always; lost only under `reset_restores_lost`) — so emission and status can no longer disagree. This is the same single-sourcing shape as #196's shared status walk.

## Tests

- Rewrite-level: the non-restoring lost `MR` records without a reset node (`R == 0`), and the `MRX` form emits no `RX`; the restoring counterpart (existing test) still pins `R == 1`.
- End to end: `M` and `MRX` on a certainly-lost qubit at the same seed read **identically shot by shot** (records and statuses) — with the old emission, `MRX`'s hidden draw shifts qubit 1's downstream `H`-measure stream. Classified slot pinned to the lost column (all 1s), vacuity guard on the downstream measurement.
- Red/green: with the fix reverted, all three pins fail.

## Validation

- C++ Debug and Release: full suites (`~[bench]`) pass
- Python: full suite passes on both wheels (Release wheel verified by binary size after the uv-cache flush; see #196 discussion of the cache footgun)
- Draw-stream fingerprint harness byte-identical (no harness config exercises a non-restoring lost measure-and-reset; the new e2e pin carries the evidence for the changed path)

## Review round

- The emission-site comment is cut to the invariant (it read like a defense of the fix — reviewer: bachase).
- Every remaining description of the measure-and-reset now states the conditional reset half (rewriter header, Python module docstring, design note): the record is always kept; the reset re-prepares the site only when it restores it — a leaked qubit always, a lost qubit only by policy. (External-review P3.)

### DIFF
--- a/src/clifft/noncomp/rewriter.cc
+++ b/src/clifft/noncomp/rewriter.cc
@@ -147,11 +147,8 @@ void process_ordinary_node(const AstNode& node, uint32_t op_index,
                 out.nodes.push_back(readout_noise_op(slot, flip));
                 noise_node = out.nodes.size() - 1;
             }
-            // The reset half is emitted only when the status stepper restored
-            // the qubit (leaked always; lost only when the policy opts in).
-            // A non-restoring lost qubit keeps its vacated carrier: emitting
-            // the reset would be dead work on a carrier nothing reads, and
-            // the X/Y-basis forms would spend a hidden SVM draw on it.
+            // The reset half runs only when the stepper restored the site;
+            // a non-restoring lost qubit keeps its vacated carrier.
             if (is_measure_reset(gate) && is_computational(status[qubit])) {
                 out.nodes.push_back(single_qubit_op(reset_for(gate), qubit));
             }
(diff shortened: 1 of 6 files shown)


src/clifft/noncomp/rewriter.cc
```diff
@@ -147,7 +147,12 @@ void process_ordinary_node(const AstNode& node, uint32_t op_index,
                 out.nodes.push_back(readout_noise_op(slot, flip));
                 noise_node = out.nodes.size() - 1;
             }
-            if (is_measure_reset(gate)) {
+            // The reset half is emitted only when the status stepper restored
```
bachase: Make this comment more concise and less like a defense of how you fixed a bug/gap in the code.

bachase: Cut to the invariant in 521f740: two lines stating when the reset runs, rationale dropped. The same commit fixes the three remaining docs that still described the reset half as unconditional (rewriter header, module docstring, design note).

PR #200, opened 2026-07-09
fix(backend): equal-rate sites keep stabilizer cost under exact damping

bachase:

## Summary

Fixes the final review's B1: under the default `damping="exact"`, a source-independent dormant-random site — every `LOSS`, by construction — expanded its qubit into the amplitude array, making loss-only models exponential in qubit count (measured: 5 qubits × `LOSS(0.01)` needed peak rank 5 under exact vs 0 under neglect) for an answer available at stabilizer cost.

The expansion bought nothing: the trap-form lowering is **bit-exact** when the two computational columns agree. The skipped no-fire back-action `diag(r, r)` is proportional to identity, the dormant-random populations are exactly ½/½ so the fire rate is exact, the trap's source draw is the Born posterior, and the forced trace-out collapse is the exact Kraus collapse — the same battle-tested mechanism the neglect path always used. The docs already promised this cost model ("a site whose rates differ between `g` and `e` needs its qubit in the active array"; `LOSS` "always qualifies"); the backend was the outlier, and after this one-condition change every doc sentence is true as written.

The classification becomes: trap-form when `neglect_damping` **or** `p_total[0] == p_total[1]`; source-dependent exact sites expand exactly as before (the pre-existing lowering test's fixture is source-dependent and is retitled to say so).

## Tests (all red with the fix reverted)

- **Opcode-level pin**: an equal-rate site under exact damping lowers to `OP_INSTRUMENT_DORMANT_NEGLECT` with `peak_rank == 0` and `neglect_damping == false` on the site record — proving it took the equal-rate path, not the policy path.
- **Physics at rank 0**: `H; LOSS(0.3); H; M` under exact at `max_rank=0` — survivors read 0 deterministically (interference fully preserved through the trap-form site), lost qubits read the classifier's lost column, and the loss rate lands within 4σ of 0.3.
- **No per-qubit accumulation**: five coherent qubits with per-qubit `LOSS` run at `max_rank=0`.

## Validation

- Full C++ suites (Debug + Release, `~[bench]`) and full Python suites on both size-verified wheels
- The sharp probes and oracle cross-checks pass unchanged (the source-dependent exact path is untouched; the damping-null probe now exercises identical lowerings by construction)
- **Fingerprint harness byte-identical** — draw streams change only for exact-mode circuits with an equal-rate site on a coherent qubit, and no harness config has one; the changed path's correctness is carried by the new pins and the closed-form suites
- One stale execute-level fixture pinned the old lowering (an equal-rate certain site expecting expansion + `peak_rank == 1`); retuned to a source-dependent site that still exercises the fused expand+damp path with a deterministic outcome (fire lands in `g` by destination, no-fire by the annihilating filter)

## Review round

- The classification comment is rewritten in plain words (reviewer: bachase) — what the no-expansion form does, why fires trap instead of collapsing in-line, and why equal rates may take it, without the jargon.
- The neglect-only naming debt is swept: comments and test names that described `OP_INSTRUMENT_DORMANT_NEGLECT`'s mechanism as "neglect-form" now say **trap-form** (named for what it does), and the opcode's declaration comment names both selectors (neglect, or equal rates). The opcode identifier itself is unchanged — renaming it is wider churn than a pre-merge round should carry; happy to do it as a follow-up if wanted.

### DIFF
--- a/src/clifft/backend/backend.cc
+++ b/src/clifft/backend/backend.cc
@@ -977,16 +977,21 @@ CompiledModule lower(const HirModule& hir, std::span<const uint8_t> postselectio
                     opcode = Opcode::OP_INSTRUMENT_ACTIVE;
                 } else if (result.basis == LocalizedBasis::Z_BASIS) {
                     opcode = Opcode::OP_INSTRUMENT_DORMANT_STATIC;
-                } else if (site.neglect_damping) {
-                    // Dormant-random under neglect: no expansion and no
-                    // no-fire back-action. Every fire traps: an in-line
-                    // collapse would re-anchor the frame conditionally,
-                    // which compiled downstream code cannot account for
-                    // (a measurement's anchor is unconditional, so its
+                } else if (site.neglect_damping || site.p_total[0] == site.p_total[1]) {
+                    // Dormant-random without expansion: no no-fire
+                    // back-action, and every fire traps (an in-line collapse
+                    // would re-anchor the frame conditionally, which
+                    // compiled downstream code cannot account for -- a
+                    // measurement's anchor is unconditional, so its
                     // compile-time virtual H is sound; a fire's is not).
+                    // Equal per-source rates take this path even under
+                    // exact damping because it is exact there: the skipped
+                    // no-fire back-action is proportional to identity, and
+                    // the deferred trace-out collapse is the Born posterior.
                     opcode = Opcode::OP_INSTRUMENT_DORMANT_NEGLECT;
                 } else {
-                    // Dormant-random under exact damping: activate the
+                    // Dormant-random with source-dependent rates under
+                    // exact damping: activate the
                     // qubit as route_to_active_z does, with the expansion
                     // fused into the instrument (+1 to k at this site).
                     auto next_axis = static_cast<uint16_t>(ctx.reg_manager.active_k());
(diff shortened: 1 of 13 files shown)


src/clifft/backend/backend.cc
```diff
@@ -977,16 +977,21 @@ CompiledModule lower(const HirModule& hir, std::span<const uint8_t> postselectio
                     opcode = Opcode::OP_INSTRUMENT_ACTIVE;
                 } else if (result.basis == LocalizedBasis::Z_BASIS) {
                     opcode = Opcode::OP_INSTRUMENT_DORMANT_STATIC;
-                } else if (site.neglect_damping) {
-                    // Dormant-random under neglect: no expansion and no
-                    // no-fire back-action. Every fire traps: an in-line
-                    // collapse would re-anchor the frame conditionally,
-                    // which compiled downstream code cannot account for
-                    // (a measurement's anchor is unconditional, so its
+                } else if (site.neglect_damping || site.p_total[0] == site.p_total[1]) {
```
bachase, line 980: Please improve the comment to be clearer, and easier to unerstand. Its reads a bit jargon-y and LLM-y

bachase: Rewritten in plain words in 16926e3 — what the form does, why fires trap, and why equal rates get it. The neglect-only naming debt across comments and test names is swept in de7ed9b (neglect-form → trap-form, opcode declaration comment names both selectors); the opcode identifier itself is left for a follow-up if you want the wider rename.

PR #202, opened 2026-07-09
fix(noncomp): reject transition keys that can never fire as hooks

bachase:

## Summary

Fixes the blocker from the post-#201 external review: `transitions={"MXX": ...}` was accepted and registered a gate hook, but the parser desugars MXX/MYY/MZZ into MPP nodes before hook expansion matches `node.gate`, so the hook silently never fired — sampling returned a different model than the one specified. Reproduced at runtime before the fix (final statuses stay computational, records show pure parity physics).

The reproduction surfaced a second consequence: the same physics written as explicit annotations is *rejected* by the driver's parity-measurement gate ("MPP is not supported under a model that can leak or lose qubits"), so the dead hook was also silently bypassing that reject. Preserving gate identity through parsing would therefore mostly manufacture rejects — model construction now refuses these keys and points at explicit `LEVEL_TRANSITION[...]` annotations, the review's other suggested arm.

The rejection also covers the quieter members of the same trap family: `CH`/`CCX`/`CCZ` (parser-only names, decomposed at parse) and the identity no-ops `I`/`II`/`I_ERROR`/`II_ERROR` (parse but never emit AST nodes) were silently accepted as named-only transitions despite the documented "a key that names a gate is a hook" contract.

Also lands the review's simplification finding: key validation now shares the parser's tag grammar via `is_representable_tag()` (exported from `parser.h`, defined next to the tag-parsing code) instead of a partial denylist. Keys containing `#` (stripped as a comment before tag parsing) or edge whitespace (tags are trimmed) were accepted but could never be referenced.

## Tests (+22 C++, +3 Python)

- Rejection contract for all ten dead-hook names, each with its distinct message.
- Tag-grammar rejects (`#`, `]`, newline, edge/only whitespace, empty) and accepts (`a b`, `a[b`, `T1-decay`), plus a parse round-trip for the accepted keys.
- Hook-fires invariant: for every GateType the model registers as a hook, a sample line must parse to a node of that type and expand through `annotate()`, and the sample-line table must cover exactly the hookable set — so a future gate that is parse-rewritten the way MXX was fails this test loudly.

## Docs

`Model` docstring and the guide's gate-hooks bullet now state the rejection and the explicit-annotation alternative.

## CI note

The first push failed CI in ctest discovery, not in test logic: `catch_discover_tests` re-parses Catch2 test names, and a lone `]`, `[`, or `#` in a TEST_CASE name breaks its splitting, collapsing every later test into one semicolon-joined entry that fails immediately. The key-grammar test names now spell those characters in words (`cce68c4`). Verified via `ctest -E Bench` on both configs (1093/1093) — the same harness CI uses; a direct Catch2 run cannot see this failure mode.

## Review round

`is_parser_desugared` is now a `parser_desugared` gate trait declared at the MXX/MYY/MZZ table rows instead of an enum switch, matching the other trait predicates (`ecef4a0`).

Extended in `45f0e96`: CH/CCX/CCZ join the gate table with the same `parser_desugared` trait (they were parser-only strings), the parser dispatches their rewrites on the enum after gate lookup, `is_parse_only_instruction` is deleted, and the model's two rejection branches collapse into one with a single message ("rewritten at parse time and never appears as a circuit node"). The hook-fires invariant test covers them through its enum loop, and the Python `GateType` enum gains the three values (introspection tripwire green).

### DIFF
--- a/src/clifft/circuit/gate_data.h
+++ b/src/clifft/circuit/gate_data.h
@@ -162,6 +162,9 @@ struct GateTraits {
     bool measure_reset = false;
     bool identity_noop = false;
     bool noise = false;
+    // Lowered by the parser into other node kinds, so the gate never
+    // appears as an AST node.
+    bool parser_desugared = false;
     std::string_view name;
 };
 
@@ -240,9 +243,9 @@ inline constexpr GateTraits kGateTraitsData[] = {
     {.arity = S, .measurement = true, .measure_reset = true, .name = "MR"},
     {.arity = S, .measurement = true, .measure_reset = true, .name = "MRX"},
     {.arity = ML, .measurement = true, .name = "MPP"},
-    {.arity = P, .measurement = true, .name = "MXX"},
-    {.arity = P, .measurement = true, .name = "MYY"},
-    {.arity = P, .measurement = true, .name = "MZZ"},
+    {.arity = P, .measurement = true, .parser_desugared = true, .name = "MXX"},
+    {.arity = P, .measurement = true, .parser_desugared = true, .name = "MYY"},
+    {.arity = P, .measurement = true, .parser_desugared = true, .name = "MZZ"},
     // Resets
     {.arity = S, .reset = true, .name = "R"},
     {.arity = S, .reset = true, .name = "RX"},
@@ -311,10 +314,8 @@ inline constexpr bool is_measure_reset(GateType g) {
 inline constexpr bool is_identity_noop(GateType g) {
     return gate_traits(g).identity_noop;
 }
-// MXX/MYY/MZZ parse into MPP nodes with Pauli-tagged targets, so these
-// GateTypes never appear in a parsed circuit.
 inline constexpr bool is_parser_desugared(GateType g) {
-    return g == GateType::MXX || g == GateType::MYY || g == GateType::MZZ;
+    return gate_traits(g).parser_desugared;
 }
 inline constexpr bool is_noise_gate(GateType g) {
     return gate_traits(g).noise;
--- a/src/clifft/circuit/gate_data.h
+++ b/src/clifft/circuit/gate_data.h
@@ -311,6 +311,11 @@ inline constexpr bool is_measure_reset(GateType g) {
 inline constexpr bool is_identity_noop(GateType g) {
     return gate_traits(g).identity_noop;
 }
+// MXX/MYY/MZZ parse into MPP nodes with Pauli-tagged targets, so these
+// GateTypes never appear in a parsed circuit.
+inline constexpr bool is_parser_desugared(GateType g) {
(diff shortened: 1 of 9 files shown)


src/clifft/circuit/gate_data.h
```diff
@@ -311,6 +311,11 @@ inline constexpr bool is_measure_reset(GateType g) {
 inline constexpr bool is_identity_noop(GateType g) {
     return gate_traits(g).identity_noop;
 }
+// MXX/MYY/MZZ parse into MPP nodes with Pauli-tagged targets, so these
+// GateTypes never appear in a parsed circuit.
+inline constexpr bool is_parser_desugared(GateType g) {
+    return g == GateType::MXX || g == GateType::MYY || g == GateType::MZZ;
```
bachase: Should this be a gate trait instead?

bachase: Yes — moved to a `parser_desugared` trait on the MXX/MYY/MZZ table rows in ecef4a0; the predicate now reads the table like its siblings.

bachase: Extended to CH/CCX/CCZ in 45f0e96 — they join the table with the same trait, the parser dispatches their rewrites on the enum, and the string list is gone.

src/clifft/circuit/gate_data.h
```diff
@@ -147,7 +153,7 @@ enum class GateType : uint16_t {
 enum class GateArity : uint8_t {
     SINGLE,      // Single qubit (H, S, X, T, M, etc.)
     PAIR,        // Two qubits consumed in pairs (CX, CY, CZ)
-    TRIPLE,      // Three qubits consumed in triples (3-qubit noise channels)
+    TRIPLE,      // Three qubits consumed in triples (3-qubit noise channels and parse rewrites)
```
bachase: Its not really parse rewrites, its some 3-qubit gates that happen to be rewritten, I'd leave the comment as 3-qubit noise and gates (CCX, CCZ, ...)

src/clifft/circuit/parser.cc
```diff
@@ -71,6 +71,25 @@ std::string_view trim(std::string_view s) {
     return s;
 }
 
+}  // namespace
+
+// Tags are trimmed as they are parsed, and a '#' starts a comment that
+// is stripped before tag parsing. A key with leading/trailing whitespace
+// or a '#' could never be written as a LEVEL_TRANSITION[key] tag and
+// reach the model unchanged.
```
bachase, line 79: Asking to understand -- what does tag mean in the parser in general? How much does this comment need to be specific about LEVEL_TRANSITION/"reach the model"?

src/clifft/circuit/gate_data.h
bachase:
```suggestion
    TRIPLE,      // Three qubits consumed in triples (3-qubit noise channels and gates)
```

Issue #223, opened 2026-07-29
Track continuation-prefix stability in optimization pass metadata


bachase:
`noncomp.sample` recompiles trajectory continuations and resumes an existing VM state after an instrument trap. Its pass pipeline currently selects default passes using `record_order.preserved`. Preserving measurement-record order is necessary for forced hidden measurements, but does not guarantee that recompiling with a different suffix produces the same already-executed bytecode prefix.

There is no known bug in the current passes: instruments act as barriers, prefix identity has a dedicated test, and debug builds compare real continuation prefixes. This issue would make that requirement explicit and harder to violate when adding passes.

Proposed work:

- Add a mandatory `preserves_instrument_prefix` classification to `PassInfo`, with every pass classified explicitly.
- Require both `record_order.preserved` and `preserves_instrument_prefix` in the trajectory pipeline.
- Add registry-driven tests for every pass claiming prefix stability, using curated prefix/suffix pairs plus deterministic randomized suffixes.
- Compare through the corresponding instrument site, including constant-pool data referenced by the prefix where necessary.
- Consider retaining the continuation-prefix check in release builds and failing with `logic_error` instead of relying only on debug assertions.
- Expose the new classification in the optimization-pass documentation.

Future custom pass-manager support is out of scope. Segmented pass execution may provide prefix stability by construction, but needs separate design work around module metadata, stateful passes, and lowering.

Issue #208, opened 2026-07-10
Add a LEAKAGE(p) noncomputational circuit annotation


bachase:
## Problem or motivation

A common source-independent leakage channel currently requires users to build a five-by-five transition matrix and reference it with `LEVEL_TRANSITION[...]`. Clifft already provides concise `LOSS(p)` syntax, while Deltakit-Stim demonstrates the usability of a circuit-level [`LEAKAGE(p)` instruction](https://github.com/Deltakit/deltakit-stim/blob/main/doc/deltakit_stim_gates.md#the-leakage-instruction).

## Proposed solution

Add `LEAKAGE(p) q...` as a noncomputational annotation with explicit Clifft semantics:

- Transition `G -> LEAK_G` and `E -> LEAK_E` with probability `p`.
- Leave already-leaked and lost sources unchanged.
- Apply independently to each target.
- Accept only finite `p` in `[0, 1]` and preserve portable deterministic sampling.
- Reject ordinary `clifft.compile()` with the existing direction to use `clifft.noncomp.sample()`.
- Do not imply partner depolarization or leakage spreading; those are separate model behavior tracked in #207.

This is syntax for an existing exact transition pattern, not a change to the default interaction policy or a claim of full Deltakit-Stim circuit compatibility.

## Alternatives considered

`LEVEL_TRANSITION[tag]` is already fully expressive but verbose for this common case. A Python-only matrix helper would reduce boilerplate but would not provide concise, self-contained circuit syntax.

Issue #207, opened 2026-07-10
Design configurable interaction effects for leaked qubits


bachase:
## Problem or motivation

Clifft currently drops a physical gate that touches a leaked or lost operand, leaving computational partners unchanged. Some QEC studies instead model a leaked carrier as depolarizing an interaction partner and optionally spreading leakage. Deltakit-Stim provides one example of this model through automatic partner depolarization and configurable leakage spreading/mobility on selected two-qubit gates: [overview](https://github.com/Deltakit/deltakit-stim/blob/main/README.md#how-deltakit-stim-works), [gate reference](https://github.com/Deltakit/deltakit-stim/blob/main/doc/deltakit_stim_gates.md).

Clifft should be able to represent both approaches without changing the existing exact default or introducing global topology into the VM.

## Proposed solution

Design an explicit model-level interaction configuration for physical two-qubit gates:

- Keep the current drop behavior as the default.
- Independently configure a Pauli/depolarizing channel on a computational partner when the other operand is leaked.
- Optionally configure directional leakage spreading onto the computational partner.
- Treat leaked and lost operands separately.
- Preserve the distinction between physical gates and record-controlled virtual feedback.

Implement the effects in the continuation rewriter, where entry statuses and ordered local operands are known. It can replace the dropped gate with localized Pauli noise and, for spreading, a generated exact transition site. The VM remains unaware of noncomputational status and global topology.

Correlated leakage mobility/transfer is out of scope for the initial implementation, but the design should not preclude adding it later.

## Alternatives considered

Overloading standard gate arguments with a fixed set of spreading/mobility probabilities, as Deltakit-Stim does, couples circuit syntax to one physical model. Runtime topology-dependent handling would violate Clifft's compiler-localization invariant. Explicit user annotations also cannot currently condition their effect on another operand's trajectory status.

Issue #206, opened 2026-07-10
Add in-record leakage and loss herald operations


bachase:
## Problem or motivation

Clifft can return a classifier's third symbol in the `heralds` sidecar, but it cannot place a non-destructive leakage or loss status probe into the ordinary measurement record. As a result, `DETECTOR`, `OBSERVABLE_INCLUDE`, and classical feedback cannot consume that information in-circuit.

This is influenced by Deltakit-Stim's [`HERALD_LEAKAGE_EVENT`](https://github.com/Deltakit/deltakit-stim/blob/main/doc/deltakit_stim_gates.md#the-herald_leakage_event-instruction), extended to preserve Clifft's distinction between leakage and loss.

## Proposed solution

Add `HERALD_LEAKAGE_EVENT` and `HERALD_LOSS_EVENT` as noncomputational annotations:

- Produce one ordinary record bit per target without changing qubit status or quantum state.
- Leakage heralds `LEAK_G` or `LEAK_E`; loss heralds only `LOST`.
- Use perfect detection by default, with any false-negative probability explicit and optional.
- Do not require a measurement classifier or populate the classifier `heralds` sidecar.
- Support `rec` consumers including detectors, observables, and feedback.
- Reject ordinary `clifft.compile()` with the existing direction to use `clifft.noncomp.sample()`.

The continuation rewriter can lower a known status to `MPAD`, plus `READOUT_NOISE` when requested, without adding a VM opcode or runtime topology dependency.

## Alternatives considered

Inspecting `final_status` or the `heralds` sidecar after sampling cannot drive in-circuit detectors or feedback. A single combined noncomputational herald would also discard the useful leak-versus-loss distinction.

PR #231, opened 2026-07-29
feat(noncomp): add leakage and loss simulation

### LINKED ISSUE #104: Implement noncomputational structural histories MVP

bachase:
## Summary

- add an experimental five-level model for computational, leaked, and lost states
- support inline and model-attached transitions, measurement classification, status and herald outputs
- execute exact state-dependent transitions through SVM instrument traps and trajectory-specific continuations
- expose the Python API and document the model, semantics, limitations, and optimization contract
- add native, end-to-end, and oracle validation across the compiler, SVM, and Python layers

This branch is a single feature commit whose tree exactly matches the reviewed `bc-cleanup-noncomputational-mvp` tip after merging current `main`. The development branch remains available separately for now and we might keep it for posterity. But it will be cleaner to have this squashed commit, since there were various attempts and changes to the approach while it we developed it.

These changes have been reviewed over a long series of PRs to those earlier branches.

## Validation

- `cmake --build build -j4`
- `ctest --test-dir build -E "^Bench:" --output-on-failure -j4` (1100 passed)
- `uv run pytest tests/python/ -q` (858 passed)

Benchmark tests were intentionally excluded. Documentation code blocks, strict documentation build, WASM, and cross-platform coverage run in PR CI.

## Follow-ups

- #206 adds in-record leakage and loss herald operations
- #207 considers configurable leaked-partner interaction effects
- #208 adds concise `LEAKAGE(p)` circuit syntax
- #223 makes continuation-prefix stability explicit in pass metadata

Closes #104


FarLab:
## Review by Claude (Fable 5)

*This review was produced by Claude Fable 5 (Anthropic), at farrokhlabib's request. Method: four parallel review passes (noncomp core, SVM, compiler pipeline, API/tests/docs) over the full diff, with every reported finding verified against the checked-out PR head — quoted line numbers refer to this branch — and the top finding reproduced empirically against the built library.*

**Verdict up front: this is an unusually well-engineered PR, and no blocking correctness defect was found on any path reachable from parsed circuit text.** The findings below are validation gaps and metadata contradictions, not broken math. Recommendation: approve with requested changes (items 1–6).

### Findings (ranked)

**1. [Moderate] Duplicate qubit targets in one hand-built annotation alias the event key → out-of-bounds read mid-sampling.** The trap machinery keys events on `(op_index, qubit)` (`trajectory_driver.cc:84–105`, `:797`), but `validate_annotation` never rejects a repeated qubit within one multi-target node. Two instrument sites then share a key; on a shot where the first site doesn't fire but the second traps, the driver indexes `instrument_offsets[trap.site_id]` past the end of a size-1 vector (formal UB in release) before `resume()`'s guard throws. Reproduced with a hand-built `AstNode{GateType::LOSS, {qubit(0), qubit(0)}, {0.5}}` + `M 0`: ~25% of shots throw mid-sampling. Parsed Stim text is immune (`LOSS`/`LEVEL_TRANSITION` are single-arity), but hand-built `Circuit`s are an explicitly defended surface — `validate_annotation` already rejects rec/pauli/inverted/out-of-range targets, and `tests/test_trajectory_driver.cc:609` tests exactly that surface. Suggested fix: reject duplicates there (or key by per-node target ordinal), plus a regression test.

**2. [Moderate] `READOUT_NOISE` flip probabilities are never range- or NaN-validated.** `parser.cc:359` checks only argument count; `frontend.cc:1041–1046` copies the values raw into the HIR; no later stage checks them. `M 0` + `READOUT_NOISE(1.5, nan) rec[-1]` compiles and silently samples garbage — NaN comparisons are extra hazardous under the project's `-ffast-math`. Pointed because this PR makes the instruction user-writable with a new two-argument form while adding strict validation to every other new surface (`LOSS` at `parser.cc:355`, instruments via `is_probability` at `frontend.cc:417/425`). Suggested fix: `is_probability` on both args at parse or trace time + a rejection test.

**3. [Medium, robustness] `execute()` silently accepts a state with a stale pending trap** (`svm.cc:159–165`): it re-runs from offset 0 on the half-collapsed array, and the leftover trap then masquerades as a fresh trap at the old `site_id`. `resume()` validates six preconditions; `execute()` validates none. All in-tree callers reset first, so this is a robustness gap rather than a live bug — an `assert`/`throw` mirroring the `resume()` check closes it.

**4. [Medium, API footgun] Transition keys naming noise channels are accepted but silently never fire.** `Model(transitions={"DEPOLARIZE1": leak})` builds successfully and does nothing (`model.cc:26–37`, `86–90` — deliberate and C++-tested, but neither `docs/guide/leakage-and-loss.md` nor the `Model` docstring mentions it; both enumerate only the *rejected* keys). "Leakage accompanies depolarization" is a natural intent, so either reject these keys like MXX/CCZ, or document the inert case loudly.

**5. [Low] "WASM support" is link-repair, not support.** Only `svm_instrument_kernels.cc` + `xoshiro.cc` were added to the WASM source list so the build keeps linking; no `noncomp/*.cc` is compiled in, nothing is exposed in `src/wasm/bindings.cc`, and the rejection message points at `clifft.noncomp.sample`, which doesn't exist there. Suggest rewording the claim (and noting there's no flag to strip the instrument kernels).

**6. [Low, latent contradictions with invariants this PR introduces]**
- `RemoveNoisePass` is registered `kPreservesRecordOrder` (`pass_registry.h:36–38`) but erases `READOUT_NOISE`, which `commutation.cc:14–15` treats as a record writer. Masked today only because the pass is default-off; should be `kBreaksRecordOrder` before any pipeline filters on the flag alone.
- `DropNonUnitaryPass` (`drop_non_unitary_pass.cc:31–44`) resets every derived HIR field except the new `forced_traceout_slot`, leaving it pointing at a deleted op.
- A multi-target reset named as `forced_traceout_node` is guarded only by an `assert` (`frontend.cc:773–779`); in release builds it silently forces the wrong slot. A thrown error would be cheap insurance.

**7. [Low, exactness edges]**
- The non-fused p=1 no-fire path can collapse onto a dust-mass survivor (`svm_kernels.inl:2080/2120` with `svm_instrument_kernels.cc:75`): debug-assert abort, or release-mode rescale of ~1e-18 mass into O(1) amplitudes. Probability is bounded by the dust mass itself, but it's inconsistent with the deterministic clamping `sample_branch` applies elsewhere.
- Two per-ISA renormalization expressions (`svm_kernels.inl:2076/2118`) can FMA-contract on the AVX2/AVX-512 TUs but not scalar → ulp-level gamma skew across ISAs. Branch decisions provably cannot diverge (they live in the shared TU), so this is output-skew only; hoisting the two expressions into the shared TU would make it bit-exact.

**8. [Design question] Tags are rejected on everything except `LEVEL_TRANSITION`** (`parser.cc:235–238`), while upstream Stim v1.15+ allows `NAME[tag]` on every instruction — so real tagged circuits (e.g. Crumble output) now fail to parse. The error is loud and located, and it prevents silent tag drops, so this may be intended; flagging it as an explicit Stim-compat decision to confirm.

**9. [Minor]** Exact-query rejection messages (`basis_probabilities.cc:471`, `record_probabilities.cc:93`) don't mention leakage/loss as a possible culprit; no CHANGELOG entry for the feature; AVX-512 instrument dispatch is never deliberately exercised in CI (pre-existing matrix gap the new opcodes inherit); untested combinations: hooks/annotations inside `REPEAT` blocks, the `Circuit`-object input path of `noncomp.sample` (only the string path is in the suite), `MRY` on a noncomputational qubit end-to-end.

### What was verified solid

- **Optimizer soundness / exactness:** instruments act as a hard positional fence in `can_swap`, peephole's `is_blocked`, both fusion scans, and squeeze; all bytecode passes are allowlist run-fusers. The side-table Pauli conjugation of instrument masks (incl. the arena-resident destination flip with sign tracking, `peephole.cc:213–228`) is correct and directly tested in both orientations. No under-conservative reorder found.
- **Trap/continuation integrity:** all mutable VM state persists in the one `SchrodingerState`; `grow_for_continuation` is strong-exception-safe and grow-only; noise re-anchoring is statistically exact; the continuation-prefix stability concern (#223) has a direct bit-identity test plus a debug-build bytecode-prefix comparison in the driver.
- **Invariants:** the static k-trajectory survives (instrument expansion is unconditional per module, `reg_manager.activate()` keeps `peak_rank` accounting); exact-query APIs cleanly reject instrument programs; the no-leakage hot path pays essentially nothing (four jump-table entries, two predictable per-shot branches, one extra load only for circuits already using readout noise).
- **Probability machinery:** RNG domain separation is sound (8 distinct domain labels, independently derived per-shot streams); trap-draw conditioning exactly matches the SVM's two trap forms; herald once-per-slot reuse is unbiased (classified levels provably cannot change between continuations); classifier conditioning `P(1|level)/(1−p_herald)` and the inverted-target `{p01,p10}` swap match the frontend semantics.
- **Validation quality is the PR's standout:** two independently built oracles (a from-first-principles dense Kraus simulator and an exact branching enumerator, each self-checked before being trusted), hand-derived closed forms with exact-zero assertions on impossible outcomes, 1e-12 dense-oracle kernel tests, and `test_rep_code_round_tvd_reaches_shot_noise` matching the exact joint 8-bit record distribution of a d=3 repetition-code round with source-dependent leakage. Edge cases (p=0/1, all-lost, lost-qubit measurement, seeded reproducibility, detector/observable interaction, trap-spanning detectors) are covered.
- **Bonus:** the peak-rank bound tightening 63→60 (`svm_state.cc:23`) fixes a real pre-existing `size_t` overflow (2^60 × 16 B = 2^64 → `malloc(0)`).

*Caveat: this is an AI-generated review. Findings 1–3 were verified by direct code inspection (finding 1 additionally by execution); treat the rest as carefully-checked pointers, not verdicts.*

PR #234, opened 2026-07-31
perf(backend): reuse noise arena when lowering

inmzhang:
## Summary

When I ran simulation for some large-distance QEC circuits using clifft, I noticed the single-thread memory consumption can be very high. After some profiling, I found reusing the existing noise arena during consuming lowering can reduce the peak memory usage by half on some workloads. 

The lowering pass now moves HIR-owned noise masks into the compiled constant pool instead of allocating a second full arena. This reduces peak memory for large noisy circuits while preserving the existing runtime behavior.

## Benchmarks

Measured with GCC 15, `RelWithDebInfo`, native CPU baseline, one pinned CPU core, and the consuming lowering path. Values are median results.

| Workload | Baseline peak RSS | Branch peak RSS | Memory change | Baseline time | Branch time | Time change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| DEPOLARIZE2, 1024 qubits, 5,000 shots | 49.18 MiB | 30.70 MiB | -37.6% | 67.488 ms | 65.274 ms | -3.3% |
| DEPOLARIZE2, 2,048 qubits, 10,000 shots | 171.27 MiB | 98.01 MiB | -42.8% | 395.225 ms | 386.187 ms | -2.3% |

The change lowers peak memory substantially and shows no compile-time regression on either memory-heavy workload.

## Test plan

  - 1,106 non-benchmark tests passed
  - 862 passed using the GCC 15 workaround
  - 30 passed, 7 skipped
  - All hooks passed using the no-project workaround

## Contributor agreement

### DIFF
--- a/src/clifft/backend/backend.cc
+++ b/src/clifft/backend/backend.cc
@@ -654,6 +654,30 @@ std::complex<double> frame_composition_phase(stim::Tableau<kStimWidth> composed,
 
 namespace {
 
+size_t validate_reusable_noise_channel_masks(const HirModule& hir) {
+    std::vector<bool> seen(hir.noise_channel_masks.size());
+    size_t count = 0;
+    for (const auto& op : hir.ops) {
+        if (op.op_type() != OpType::NOISE)
+            continue;
+        const auto site_idx = static_cast<size_t>(op.noise_site_idx());
+        if (site_idx >= hir.noise_sites.size())
+            throw std::runtime_error("noise site index out of bounds during consuming lowering");
+        for (const auto& channel : hir.noise_sites[site_idx].channels) {
+            const auto handle = static_cast<size_t>(channel.mask);
+            if (handle >= seen.size())
+                throw std::runtime_error(
+                    "noise channel mask handle out of bounds during consuming lowering");
+            if (seen[handle])
+                throw std::runtime_error(
+                    "duplicate noise channel mask handle during consuming lowering");
+            seen[handle] = true;
+            ++count;
+        }
+    }
+    return count;
+}
+
 CompiledModule lower_impl(const HirModule& hir,
                           std::optional<PauliMaskArena> reusable_noise_channel_masks,
                           std::span<const uint8_t> postselection_mask,
@@ -1137,8 +1161,10 @@ CompiledModule lower(HirModule&& hir, std::span<const uint8_t> postselection_mas
                      std::span<const uint8_t> expected_detectors,
                      std::span<const uint8_t> expected_observables) {
     // Noise-removing passes release the arena as a zero-width empty object.
-    // Rebuild the correctly sized empty pool when there is nothing to reuse.
-    if (hir.noise_channel_masks.num_words() != (hir.num_qubits + 63) / 64) {
+    // Overallocated frontend arenas are also copied into a compact pool.
+    const size_t actual_channel_count = validate_reusable_noise_channel_masks(hir);
+    if (hir.noise_channel_masks.num_words() != (hir.num_qubits + 63) / 64 ||
+        hir.noise_channel_masks.size() != actual_channel_count) {
         return lower_impl(hir, std::nullopt, postselection_mask, expected_detectors,
                           expected_observables);
     }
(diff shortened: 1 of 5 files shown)


src/clifft/backend/backend.cc
```diff
@@ -857,8 +868,9 @@ CompiledModule lower(const HirModule& hir, std::span<const uint8_t> postselectio
 
                 NoiseSite mapped_site;
                 for (const auto& ch : hir_site.channels) {
-                    auto in_view = hir.noise_channel_masks.at(ch.mask);
-                    auto out_handle = static_cast<PauliMaskHandle>(next_cp_noise++);
+                    auto in_view = source_noise_channel_masks.at(ch.mask);
+                    auto out_handle =
```
bachase, line 896: The in-place path assumes every channel handle is unique. Although that is true in practice, the HIR structure actually allows duplicate references into the arena. If a future version does that, this copy would cause the slot to be transformed repeatedly. 

Since its not really a supported use today, would it work to defensively validate handle bounds and uniqueness before lowering and throw on violation?

src/clifft/backend/backend.cc
```diff
@@ -687,7 +691,14 @@ CompiledModule lower(const HirModule& hir, std::span<const uint8_t> postselectio
         num_noise_channels += site.channels.size();
     ctx.constant_pool.pauli_masks = PauliMaskArena(n, num_apply_paulis);
     ctx.constant_pool.exp_val_masks = PauliMaskArena(n, num_exp_val_masks);
-    ctx.constant_pool.noise_channel_masks = PauliMaskArena(n, num_noise_channels);
+    const bool reuse_noise_masks = reusable_noise_channel_masks.has_value();
+    if (reuse_noise_masks) {
```
bachase, line 719: Could reuse also require hir.noise_channel_masks.size() == actual_channel_count? 

The frontend arena is actually upper bound, such that any 0 probability channels in PAULI_CHANNEL_2/3 don't show up. Moving it unconditionally can retain substantially more storage than the compiled program needs.

src/clifft/backend/compiler_context.h
```diff
@@ -232,17 +230,22 @@ class VirtualFrame {
             }
         };
 
+        // Copy both inputs before clearing the outputs so lowering can map
```
bachase, line 234: Consider also adding this to the function documentation as a new contract requirement -- now inputs may alias outputs. Copy-before-clear is now required for correctness.

bachase (changes requested):
Thanks for this PR @inmzhang and could proposal. My comments are around a few edge cases that this exposes that I think are worth being defensive for. We can also add some tests for them.

You are welcome to make those changes yourself (if you agree they are worth doing). But if not, I'm happy to manage and add a followup to your branch and still attribute you as the contributor/author of the PR.

inmzhang:
Hi Brad, I'm happy to contribute :-)

 I'll try to address your comment at tomorrow.

bachase:
Wonderful! No rush at all, whenever convenient for you.

src/clifft/backend/backend.cc
inmzhang: Consuming lowering now validates every referenced noise-channel handle before moving the arena. out-of-range handles and duplicate references throw std::runtime_error, preventing repeated slot transformations.

src/clifft/backend/backend.cc
inmzhang: Now the arena is reused only when its width and capacity match the actual number of referenced channels. Overallocated frontend arenas, such as those caused by zero-probability PAULI_CHANNEL entries, now fall back to a compact constant-pool arena.

bachase (approved):
Thanks for addressing the feedback—these changes look good to me!

I made one small test-only follow-up that strengthens the in-place reuse test. It uses two fully utilized handles in reverse order, so the test proves the consuming path preserved the original handles rather than silently taking the compact fallback.

If you agree, feel free to cherry-pick from https://github.com/unitaryfoundation/clifft/commit/fbd1635304b90723da898a61894fec016956e484

inmzhang:
Thanks, the cherry-pick done.

PR #243, opened 2026-08-04
fix(optimizer): eliminate terminal phases across resets

### LINKED ISSUE #242: Optimizer: eliminate terminal phases across MR reset corrections
## Problem

The terminal phase-elimination pass added in #239 does not eliminate every removable phase when the matching terminal measurement is an `MR`, `MRX`, or `MRY` layer with multiple targets.

The frontend lowers each measure-reset target into an interleaved pair:

```text
MEASURE(q0)
CONDITIONAL_PAULI(reset correction for q0)
MEASURE(q1)
CONDITIONAL_PAULI(reset correction for q1)
...
```

`eliminate_terminal_measurement_phases()` treats every `CONDITIONAL_PAULI` as a barrier. Consequently, the reset correction for an earlier target prevents a phase on a later target from reaching its otherwise matching terminal measurement.

## Minimal reproduction

```stim
R 0 1
H 0 1
R_Z(0.3) 0 1
X_ERROR(0.01) 0 1
MR 0 1
```

After the default HIR optimization pipeline:

- **Expected:** both terminal rotations are removed and `peak_rank == 0`.
- **Observed:** one rotation remains and `peak_rank == 1`.

Useful controls:

- Replacing `MR 0 1` with `M 0 1` removes both rotations.
- A single-target `MR 0` removes its rotation.
- Reversing the target order changes which rotation survives, but still leaves one.
- Writing the resets as separate `MR 0` and `MR 1` instructions has the same problem.

This can leave one unnecessary active dimension in real circuits. Since dense state work scales as `2^k`, even a one-dimension miss is material.

## Proposed scope

Extend terminal phase elimination so it can look through reset-generated conditional corrections when doing so is proven safe. Two possible implementations are:

1. Recognize the adjacent `MEASURE` + reset-correction pair produced by frontend lowering and treat that pair atomically for candidates that commute with both operations.
2. Relax the unconditional `CONDITIONAL_PAULI` barrier only when the optimizer can prove both quantum commutation and safe classical dataflow.

The first option is narrower. The second may be a useful generalization, but should not allow a phase to cross same-support or anti-commuting feedback merely because the operation is conditional. The pass must also continue to respect `EXP_VAL`, instrument/continuation, measurement-record, and source-map constraints.

This issue is a follow-up to #238 and #239; it does not change the terminal-measurement identity established there.

## Regression tests

Please cover:

- Multi-target and separately written `MR` layers, including target-order permutations.
- The analogous matching-axis cases for `MRX` and `MRY`.
- A candidate belonging only to a later target, so the test cannot pass by removing only the first target.
- Disjoint reset corrections that should be transparent.
- Same-support or anti-commuting conditional feedback that must remain a barrier.
- Conditional feedback whose classical control depends on a relevant measurement record.
- Circuits with downstream detectors/observables and inverted measurement targets.
- Preservation of source maps and noncomputational continuation boundaries.
- Sampled-distribution equivalence before and after optimization, in addition to checking the optimized HIR and `peak_rank`.

## Acceptance criteria

- The reproduction compiles with `peak_rank == 0` and no surviving phase rotations.
- All matching phases in a multi-target measure-reset layer are eliminated when safe, independent of target order or source grouping.
- Unsafe conditional-feedback cases remain unchanged.
- Existing optimizer, measurement/reset, source-map, and noncomputational tests continue to pass.

bachase:
## Summary

- allow terminal measurement phases to cross conditional Paulis when their quantum supports are disjoint and the existing commutation/dataflow proof permits it
- keep same-support and anti-commuting feedback as conservative barriers
- eliminate both rotations in the issue reproduction, reducing optimized `peak_rank` from 2 to 0
- cover MR target ordering/grouping, MRX/MRY, arbitrary disjoint feedback, instruments, source maps, inverted records, and detector/observable outputs

Closes #242.

## Test plan

- [x] C++ tests pass (`ctest --test-dir build-issue242 -E '^Bench:' --output-on-failure -j4`; 1,124 passed)
- [x] Python tests pass (`uv run pytest tests/python/ -q`; 869 passed)
- [ ] Doc examples pass (not run; no documentation changes)

## Contributor agreement

- [ ] I confirm this contribution is my original work (or I have the right to
- [ ] If this contribution was AI-assisted, I have reviewed the generated code

The commit includes an `Assisted-by: ChatGPT (GPT-5.6 Sol)` trailer. The contributor checkboxes are left for the submitter to confirm.

### DIFF
--- a/src/clifft/optimizer/peephole.cc
+++ b/src/clifft/optimizer/peephole.cc
@@ -338,6 +338,18 @@ inline bool is_blocked(const HeisenbergOp& op_i, const HeisenbergOp& op_j, const
     }
 }
 
+/// Return true when two Pauli strings act nontrivially on any common qubit.
+inline bool has_overlapping_support(MaskView x1, MaskView z1, MaskView x2, MaskView z2) {
+    assert(x1.num_words() == z1.num_words() && z1.num_words() == x2.num_words() &&
+           x2.num_words() == z2.num_words());
+    for (uint32_t w = 0; w < x1.num_words(); ++w) {
+        if (((x1.words[w] | z1.words[w]) & (x2.words[w] | z2.words[w])) != 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
 /// Delete phase operations whose only remaining quantum effect is a phase in
 /// the eigenspaces of a later measurement on the same unsigned Pauli axis.
 ///
@@ -345,7 +357,9 @@ inline bool is_blocked(const HeisenbergOp& op_i, const HeisenbergOp& op_j, const
 /// with the phase axis: branch by branch it only reverses the rotation
 /// direction, and the matching measurement discards either direction. Other
 /// phase operations and measurements may be crossed only when they commute
-/// with the candidate. Classical record consumers are inspected but never
+/// with the candidate. A conditional Pauli is transparent only when its
+/// quantum support is disjoint and the ordinary quantum/classical dataflow
+/// proof permits a swap. Classical record consumers are inspected but never
 /// moved. All other operations are conservative barriers.
 size_t eliminate_terminal_measurement_phases(HirModule& hir, std::vector<uint8_t>& deleted) {
     size_t eliminated = 0;
@@ -365,14 +379,21 @@ size_t eliminate_terminal_measurement_phases(HirModule& hir, std::vector<uint8_t
                 continue;
             const HeisenbergOp& next = hir.ops[j];
 
-            // These are the two deliberate differences from ordinary
-            // commutation: terminal phases may cross every Pauli-noise branch,
-            // while conditional Paulis remain conservative barriers.
+            // Terminal phases may cross every Pauli-noise branch. Conditional
+            // Paulis require a stronger proof: ordinary commutation/dataflow
+            // safety plus disjoint quantum support. The disjointness keeps
+            // overlapping feedback conservative even when its total Pauli
+            // happens to commute symplectically with the candidate.
             if (next.op_type() == OpType::NOISE) {
                 continue;
             }
 
             if (next.op_type() == OpType::CONDITIONAL_PAULI) {
+                if (!has_overlapping_support(candidate_x, candidate_z, hir.destab_mask(next),
+                                             hir.stab_mask(next)) &&
+                    can_swap(hir.ops[i], next, hir)) {
+                    continue;
+                }
                 break;
             }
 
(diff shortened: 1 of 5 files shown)


bachase (commented):
*Review by Claude (Fable 5), posted from bachase's account

**Verdict: LGTM — correct, appropriately conservative, and empirically closes everything #242 reported.** I built this branch (`f4cb57f`) and ran an independent verification battery against the real benchmark circuits from the SymFT comparison study, beyond the synthetic cases in the test suite.

### Empirical verification (all pass)

- **Compiled-k closure** — the gap #242 was about: all four coherent-noise surface circuits now compile to SymFT-matched peak rank (**4 / 7 / 12 / 22** for d3r1 / d3r3 / d5r1 / d5r5), with zero residual `OP_EXPAND_ROT → OP_MEAS_ACTIVE_DIAGONAL` transients; MSC d5/d7 unchanged at 10/19.
- **Adversarial distribution checks** not covered by the suite: (a) disjoint feedback controlled by an intermediate measurement *correlated with the candidate qubit through entanglement*; (b) an MR layer whose reset records feed later feedback; (c) reversed + inverted `MR !1 !0`. Baseline-vs-optimized record distributions agree within 6σ at 400k shots; the anticommuting-feedback negative control stays blocked (peak 1→1) with exactly zero probability delta.
- **Benchmark semantics**: postselected discard fractions on d3r1/d3r3 match the previously validated reference values (0.7 SE / 2.6 SE at 200k/100k shots).
- **Throughput**: original-circuit rates recover to the hand-rewritten (`rzfix`) levels measured in the monitor study — expect ~1.2×/1.6×/1.7×/~4.4× on the coherent family, and the d5r1 head-to-head vs SymFT-single flipping to clifft-ahead.

### Code

- The crossing condition — disjoint quantum support **and** `can_swap` — is right. `can_swap` contributes the classical-dataflow check plus the EXP_VAL/INSTRUMENT positional fences on top of symplectic commutation (nice defense-in-depth; the new instrument-fence test pins it). This is deliberately stricter than a commuting-mask rule, and the extra conservatism costs nothing: every elimination the benchmarks need is a disjoint crossing.
- `has_overlapping_support` is clean (word-width asserts, OR-of-X|Z overlap).
- Test coverage is genuinely thorough: MR grouping/ordering/reversal, separate lines, MRX/MRY, inverted targets + readout noise, detector/observable preservation, source-map hygiene, instrument fence, and both negative controls (anticommuting; same-support commuting).

### Optional, non-blocking

1. All elimination tests are ≤3-qubit synthetics; a small surface-round-style fixture asserting compiled peak rank (the regression signature that surfaced #242 at circuit scale) would guard against re-regression.
2. Possible future relaxation: same-support *commuting* conditionals are also provably droppable-safe by the same branch argument (a record-selected commuting Pauli cannot carry the phase off-axis), but no workload measured so far needs it — fine to stay conservative.

Issue #248, opened 2026-08-05
Define SamplingPlan v0 semantics and structural planner


bachase:
Parent tracker: #247
Architecture discussion: #236

## Objective

Introduce a private, backend-independent semantic `SamplingPlan` and a deterministic planner for the first supported optimized-HIR subset. This establishes the compiler/runtime contract without changing normal compilation or SVM execution.

## Scope

- Define plan-level metadata and semantic actions separately from the legacy 32-byte `Instruction`.
- Model affine Boolean expressions with explicit symbol classes and assignment-before-use validation.
- Distinguish active state-touching measurements, deterministic classical measurements, and fresh dormant random measurements.
- Represent recorded measurement bits separately from effective eigenvalue labels used by later symbolic actions.
- Preserve stable visible/hidden record slots plus noise-site and instrument-site identities.
- Represent active-Pauli rotations/measurements, promotions, compaction pivots, and instrument segment boundaries for the initial supported HIR subset.
- Specify phase, global-scalar/deferred-normalization, and numerical dust semantics without forcing executor-specific storage into the plan.
- Provide deterministic validation and text inspection covering operations, active-width history, dependencies, stable site identifiers, and predicted dense-array passes.
- Reject unsupported HIR operations explicitly.
- Add structural tests using hand-built and frontend-produced HIR.

## Non-goals

- No executor or production routing.
- No packed sampling, product components, GPU lowering, alternate layouts, native specialization, or explicit SIMD work.
- No public Python plan API and no change to `Program`.
- No requirement that arbitrary truncated circuits reproduce a byte-identical full-plan prefix. Instrument boundaries require an explicit state-handoff contract instead.

## Acceptance criteria

- The new code is part of normal core and wasm builds.
- Representative rotation/measurement HIR produces a deterministic, valid plan containing no active-coordinate Clifford sweeps.
- Symbolic dependencies are validated as assignment-before-use.
- Instrument boundaries and stable record/site identities are represented even though continuation execution is deferred.
- Unsupported semantics fail with actionable diagnostics.
- The current compiler and SVM behavior remain unchanged.

bachase:
Implementation is split into reviewable PRs.

- #249 establishes the private semantic model, symbolic expression lifetimes, invariant validation, deterministic inspection, stable ids, and instrument-boundary vocabulary.
- The next PR will implement the optimized-HIR to stabilizer-coordinate planner against that model.
[...]

PR #249, opened 2026-08-05
feat(sampling): add semantic plan foundation

bachase:
## Summary

- add a private `SamplingPlan` representation separate from the legacy 32-byte SVM instruction
- define canonical affine Boolean expressions and explicit exogenous, branch, and derived symbol lifetimes
- distinguish effective measurement branch labels from recorded outcomes
- model active rotations and measurements, dormant promotion and measurement, classical records, and instrument boundaries
- validate active-width transitions, Pauli bounds, stable record/site identities, symbol definitions, and assignment-before-use
- add deterministic plan inspection with predicted dense-array passes
- share the existing measurement dust threshold as a backend-independent semantic constant
- include the new source in native and wasm builds

This is the semantic foundation for #248, not the HIR-to-plan algorithm. It deliberately leaves the production compiler and SVM routing unchanged. The next scoped PR can implement stabilizer-coordinate planning against this validated representation.

The design is an independent Clifft implementation informed by the SymFT architecture described in #236.

## Validation

- `cmake --build build -j 4`
- `ctest --test-dir build -R 'Sampling plan' --output-on-failure` — 8 tests passed
- `ctest --test-dir build --output-on-failure -j 4` — 1,141 tests passed

Part of #248.

### DIFF
--- a/src/clifft/sampling/plan.cc
+++ b/src/clifft/sampling/plan.cc
@@ -1,0 +1,477 @@
+#include "clifft/sampling/plan.h"
+
+#include "clifft/util/numeric.h"
+
+#include <algorithm>
+#include <iomanip>
+#include <limits>
+#include <sstream>
+#include <stdexcept>
+#include <string_view>
+#include <type_traits>
+#include <utility>
+
+namespace clifft::sampling {
+
+namespace {
+
+uint32_t index(SymbolId id) {
+    return static_cast<uint32_t>(id);
+}
+
+uint32_t index(RecordSlot slot) {
+    return static_cast<uint32_t>(slot);
+}
+
+uint32_t index(NoiseSiteId site) {
+    return static_cast<uint32_t>(site);
+}
+
+uint32_t index(InstrumentSiteId site) {
+    return static_cast<uint32_t>(site);
+}
+
+std::vector<SymbolId> xor_terms(const std::vector<SymbolId>& left,
+                                const std::vector<SymbolId>& right) {
+    std::vector<SymbolId> result;
+    result.reserve(left.size() + right.size());
+    size_t i = 0;
+    size_t j = 0;
+    while (i < left.size() || j < right.size()) {
+        if (j == right.size() || (i < left.size() && index(left[i]) < index(right[j]))) {
+            result.push_back(left[i++]);
+        } else if (i == left.size() || index(right[j]) < index(left[i])) {
+            result.push_back(right[j++]);
+        } else {
+            ++i;
+            ++j;
+        }
+    }
+    return result;
+}
+
+std::vector<SymbolId> canonicalize_terms(std::vector<SymbolId> terms) {
+    std::sort(terms.begin(), terms.end(),
+              [](SymbolId left, SymbolId right) { return index(left) < index(right); });
+    std::vector<SymbolId> result;
+    for (size_t begin = 0; begin < terms.size();) {
+        size_t end = begin + 1;
+        while (end < terms.size() && terms[end] == terms[begin]) {
+            ++end;
+        }
+        if ((end - begin) % 2 != 0) {
+            result.push_back(terms[begin]);
+        }
+        begin = end;
+    }
+    return result;
+}
+
+std::string format_expression(const AffineBool& expression) {
+    std::ostringstream out;
+    bool wrote = false;
+    if (expression.constant()) {
+        out << '1';
+        wrote = true;
+    }
+    for (SymbolId term : expression.terms()) {
+        if (wrote) {
+            out << " ^ ";
+        }
+        out << 's' << index(term);
+        wrote = true;
+    }
+    if (!wrote) {
+        out << '0';
+    }
+    return out.str();
+}
+
+std::string_view symbol_kind_name(SymbolKind kind) {
+    switch (kind) {
+        case SymbolKind::Exogenous:
+            return "exogenous";
+        case SymbolKind::Derived:
+            return "derived";
+        case SymbolKind::Branch:
+            return "branch";
+    }
+    return "unknown";
+}
+
+std::string format_words(const std::vector<uint64_t>& words) {
+    std::ostringstream out;
+    out << '[';
+    for (size_t i = 0; i < words.size(); ++i) {
+        if (i != 0) {
+            out << ',';
+        }
+        out << "0x" << std::hex << std::setw(16) << std::setfill('0') << words[i];
+    }
+    out << ']';
+    return out.str();
+}
+
+std::string format_pauli(const ActivePauli& pauli) {
+    std::ostringstream out;
+    out << "width=" << pauli.width << " x=" << format_words(pauli.x)
+        << " z=" << format_words(pauli.z);
+    return out.str();
+}
+
+[[noreturn]] void invalid_plan(std::string message) {
+    throw std::invalid_argument("invalid SamplingPlan: " + std::move(message));
+}
+
+void validate_pauli(const ActivePauli& pauli, uint32_t expected_width, uint32_t action_index) {
+    if (pauli.width != expected_width) {
+        invalid_plan("action " + std::to_string(action_index) + " Pauli width " +
+                     std::to_string(pauli.width) + " does not match active width " +
+                     std::to_string(expected_width));
+    }
+    const size_t words = (static_cast<size_t>(pauli.width) + 63) / 64;
+    if (pauli.x.size() != words || pauli.z.size() != words) {
+        invalid_plan("action " + std::to_string(action_index) +
+                     " Pauli word count does not match its width");
+    }
+    if (words != 0 && pauli.width % 64 != 0) {
+        const uint64_t tail_mask = (uint64_t{1} << (pauli.width % 64)) - 1;
+        if ((pauli.x.back() & ~tail_mask) != 0 || (pauli.z.back() & ~tail_mask) != 0) {
+            invalid_plan("action " + std::to_string(action_index) +
+                         " Pauli has bits outside its active width");
+        }
+    }
+}
+
+std::optional<SymbolId> defined_symbol(const SamplingAction& action) {
+    return std::visit(
+        [](const auto& typed) -> std::optional<SymbolId> {
+            using T = std::decay_t<decltype(typed)>;
+            if constexpr (std::is_same_v<T, MeasureActivePauli> ||
+                          std::is_same_v<T, MeasureDormantRandom>) {
+                return typed.branch;
+            } else if constexpr (std::is_same_v<T, DefineSymbol>) {
+                return typed.symbol;
+            } else {
+                return std::nullopt;
+            }
+        },
+        action);
+}
+
+void validate_expression(const SamplingPlan& plan, const AffineBool& expression,
+                         uint32_t action_index, std::optional<SymbolId> current_definition,
+                         bool allow_current_definition) {
+    if (!expression.is_canonical()) {
+        invalid_plan("action " + std::to_string(action_index) +
+                     " has a noncanonical affine expression");
+    }
+    for (SymbolId term : expression.terms()) {
+        const uint32_t symbol_index = index(term);
+        if (symbol_index >= plan.symbols.size()) {
+            invalid_plan("action " + std::to_string(action_index) + " references symbol s" +
+                         std::to_string(symbol_index) + " out of range");
+        }
+        const SymbolInfo& info = plan.symbols[symbol_index];
+        if (info.kind == SymbolKind::Exogenous) {
+            continue;
+        }
+        if (!info.defining_action.has_value()) {
+            invalid_plan("symbol s" + std::to_string(symbol_index) + " has no defining action");
+        }
+        const bool current_is_allowed = allow_current_definition && current_definition == term &&
+                                        *info.defining_action == action_index;
+        if (*info.defining_action >= action_index && !current_is_allowed) {
+            invalid_plan("action " + std::to_string(action_index) + " references symbol s" +
+                         std::to_string(symbol_index) + " before assignment");
+        }
+    }
+}
+
+void validate_record(const SamplingPlan& plan, RecordSlot record, uint32_t action_index) {
+    const uint64_t total = static_cast<uint64_t>(plan.num_visible_records) +
+                           static_cast<uint64_t>(plan.num_hidden_records);
+    if (index(record) >= total) {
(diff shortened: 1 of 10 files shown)


bachase:
Addressed the architectural review in 87ba176:

- Made the semantic/runtime split explicit: `SamplingPlan` is compile-time IR and executors must lower it to a packed, allocation-free per-shot form.
- Replaced arbitrary-length active-coordinate Pauli vectors with `uint64_t` X/Z masks. This constrains active width, not physical circuit width.
- Added the shared `kDenseActiveWidthLimit = 60` invariant and used it in both SamplingPlan validation and legacy dense-state checks.
- Renamed `negative` to `sign` and `Exogenous` to `Presampled`; aligned inspection output on active-width terminology.
- Documented the exact Hermitian Pauli phase convention and the live-state continuation contract.
- Added validation and tests for measurement-pivot support, duplicate record writes, and unsupported dense active widths.

Deliberately unchanged: `InstrumentBoundary` remains width-neutral, `AffineBool` remains a semantic vector representation pending planner profiling, and identity rotations remain semantically valid.

Validation: 11 focused sampling-plan tests, all 1,146 native tests, and all pre-commit checks pass.

bachase:
Fixed the re-review blocker in 5f3f69d. The validator now rejects any action whose before/after width reaches the dense limit before entering action-specific validation, so Pauli and pivot shifts only see supported widths. The regression test builds the formerly unsafe stream through width 70 and verifies rejection at the first transition to 60.

I also tied the legacy error text to `kDenseActiveWidthLimit` and documented the pivot-selection rationale. Focused SamplingPlan, legacy rank-limit, and state-construction tests pass, along with all pre-commit checks.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+// Private semantic plan for the symbolic-coordinate sampling path.
+//
+// This representation deliberately does not reuse the legacy SVM Instruction.
+// It describes sampling semantics; CPU layout, SIMD selection, dispatch, and
+// device lowering belong to later execution-specific layers.
+// Executors must lower it to a packed, allocation-free runtime plan rather
+// than traversing these variants and vectors in per-shot hot loops.
+//
+// The symbolic-expression and stabilizer-coordinate action split is an
+// independent Clifft implementation informed by the published SymFT design.
```
bachase: I'd rather this comment stand alone than be specific about a transition to an alternate design. It should really just frame the purpose of this file and information at a high-level for readers to understand what it does.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+
+namespace clifft::sampling {
+
+enum class SymbolId : uint32_t {};
+enum class RecordSlot : uint32_t {};
+enum class NoiseSiteId : uint32_t {};
+enum class InstrumentSiteId : uint32_t {};
+
+enum class SymbolKind : uint8_t {
+    Presampled,
```
bachase: SHould these have trailing comments to explain what each means?

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+#include <complex>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <variant>
+#include <vector>
+
+namespace clifft::sampling {
+
+enum class SymbolId : uint32_t {};
```
bachase, line 32: Should we define what a symbol is in the context of clifft/these cricuits? I think record, noise and instrument are more self-identifying.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+enum class InstrumentSiteId : uint32_t {};
+
+enum class SymbolKind : uint8_t {
+    Presampled,
+    Derived,
+    Branch,
+};
+
+// An XOR (parity) expression over plan symbols, possibly inverted. Terms are
+// stored in strict ascending order; XOR construction removes duplicates.
```
bachase: Should this have a brief example in the comment. Or something more approachable? E.g. "a parity expression of symbols in the circuit plan. This enables us to symbolically track how combinations of binary symbols correspond to actions in the circuit, e.g. bit or phase errors due to random noise"

bachase:
Fixed the manual-review finding in 18ce188. Both `MeasureActivePauli` and `MeasureDormantRandom` now require their outcome expression to contain the branch defined by that action; deterministic outcomes remain represented by `RecordClassical`. Added negative regression coverage for both measurement variants.

Validation: build succeeds, all 13 focused SamplingPlan tests pass (27 assertions), and all pre-commit checks pass.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+  private:
+    bool constant_ = false;
+    std::vector<SymbolId> terms_;
+};
+
+[[nodiscard]] AffineBool operator^(AffineBool left, const AffineBool& right);
+[[nodiscard]] AffineBool operator^(AffineBool left, bool right);
+[[nodiscard]] AffineBool operator^(bool left, AffineBool right);
+
+// The Hermitian Pauli body for one action, expressed only in the active
```
bachase: The fact that this defines the unsigned operator and then the affinebool shot-dependint sign might warrant a consolidated comment earlier in the file? 

Also, should we remind readers that an active operation is some circuit operation that is not clifford absorbed into the HIR at compile time. And this represents the pauli operator generating that action (as seen in specific operations below)

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+// coordinates visible immediately before that action. The unsigned operator is
+// i^popcount(x & z) X^x Z^z; an action's AffineBool carries its possibly
+// shot-dependent sign. Dense active width is below kDenseActiveWidthLimit, so
```
bachase: Should we add a static assert that kDeniseActiveWidthLimit is <= size of uint64_t in bits?

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+
+    [[nodiscard]] bool is_identity() const;
+};
+
+// Applies exp(-i * pi * half_turns * P / 2) to the current active state.
+struct RotateActivePauli {
+    ActivePauli pauli;
+    double half_turns = 0.0;
+    // True means the Pauli enters the rotation with a minus sign.
+    AffineBool sign;
```
bachase, line 95: True means -- if an assignment of values to symbolic variables evlautes the affine boolean expression to true, then the rotation would have a minus sign to the angle? (Just asking to confirm, but might be worth cleaning up the comment language here and below).

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+
+// Promotes one dormant coordinate and applies the rotation that introduced it.
+struct PromoteDormantRotation {
+    uint32_t dormant_pivot = 0;
+    double half_turns = 0.0;
+    // True means the promoted Pauli enters with a minus sign.
+    AffineBool sign;
+};
+
+// Samples one active Pauli branch and removes the selected active pivot. The
```
bachase: Consider a less jargon summary, e.g. when a measurement in the current basis/ clifford coordinate frame targets an active qubit, we have to sample the state and collapse the state. Not those exact words but just giving a bit more context.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+// outcome expression is the user-visible record bit after symbolic signs.
+struct MeasureActivePauli {
+    ActivePauli pauli;
+    uint32_t active_pivot = 0;
+    SymbolId branch{};
+    AffineBool outcome;
+    RecordSlot record{};
+};
+
+// Introduces an unbiased effective eigenvalue label while replacing a dormant
```
bachase: Similarly to other measurement struct -- add a little more physics/less jargony phrasing.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+};
+
+// Records a deterministic classical expression without touching the state.
+struct RecordClassical {
+    AffineBool outcome;
+    RecordSlot record{};
+};
+
+// Materializes an affine expression as a reusable derived symbol.
+struct DefineSymbol {
```
bachase, line 139: In what context do we need to define symbol vs record to a measured slot? (Asking to understand)

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+    SymbolId symbol{};
+    AffineBool value;
+};
+
+// Divides ahead-of-time execution segments. A continuation must preserve the
+// live coefficients, active-coordinate meaning and order, active width,
+// symbol and record values, and RNG position established here. It need not
+// reproduce an arbitrary prior plan prefix byte-for-byte. The boundary itself
+// is width-neutral; any state transition belongs to a following action.
```
bachase: What does width-neutral have to do in this context?

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+struct PlannedAction {
+    uint32_t active_before = 0;
+    uint32_t active_after = 0;
+    SamplingAction action;
+};
+
+struct SymbolInfo {
+    SymbolKind kind = SymbolKind::Presampled;
+
+    // Presampled symbols have no defining action. Branch and
```
bachase: Should this be phrased more explicitly "when nullopt/not set ... presampled " here and below?

Also, is this a case where illegal states are representable here and only convention stops them? If so, should we enforce this in the type system instead?

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
...
+    // Presampled symbols may identify the stable HIR noise site that supplies
+    // them. Other presampled event kinds can add their own source identity
+    // without changing expression semantics.
+    std::optional<NoiseSiteId> noise_site;
+};
+
+struct SamplingPlan {
+    uint32_t num_qubits = 0;
+
+    // Active width is the new subsystem's name for legacy active_k. The
```
bachase: Change this comments to stand alone.

But why are we using active width instead of active k? I suppose width is a better meaning, but wanting to confirm.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,182 @@
+#pragma once
+
+// Private semantic plan for the symbolic-coordinate sampling path.
```
bachase: I Want to avoid jargon, but "Plan" is also pretty generic. Should this be SamplingPlan? Or since its private, its not that important to have a clearer name?

src/clifft/sampling/plan.h
bachase: I'd state where this relates to HIR. So something like, "After a circuit is compiled to HIR and optimized, this plan captures the instruction sequence to sample from that circuit. It tracks and captures the action of stochastic events like measurement and noise as symbolic symbols, and how the active state grows/contracts. This then isolates the ahead of time/compile once aspects from what is determined at sample time."

We should cite symft paper appropriately in this comment.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,207 @@
...
+    std::vector<SymbolId> terms_;
+};
+
+[[nodiscard]] AffineBool operator^(AffineBool left, const AffineBool& right);
+[[nodiscard]] AffineBool operator^(AffineBool left, bool right);
+[[nodiscard]] AffineBool operator^(bool left, AffineBool right);
+
+// Operations requiring coefficient-state work are expressed over active
+// stabilizer coordinates. An active coordinate is represented explicitly in
+// the dense 2^k coefficient state; it is not an "active physical qubit."
```
bachase: Drop the "not an active pysical qubit" that was just my misphrasing in my PR comment. Again, these comments should stand alone going forward and not reflect our discussion itself directly.

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,207 @@
...
+// ActivePauli stores the unsigned Hermitian operator
+// i^popcount(x & z) X^x Z^z. Each action carries a separate AffineBool that can
+// change its sign for each shot.
+//
+// The dense-width limit is exclusive, so all valid masks fit in one word even
```
bachase: This is a weird phrasing ..  either don't have a comment or just say, we currently assume due to the exponential memory requirements that the dense-width is at most kDense..., which fits into a uint64. Assert notifies us to update this assumption if we ever change that in the future.

src/clifft/sampling/plan.cc
```diff
@@ -0,0 +1,497 @@
...
+            result.push_back(right[j++]);
+        } else {
+            ++i;
+            ++j;
+        }
+    }
+    return result;
+}
+
+std::vector<SymbolId> canonicalize_terms(std::vector<SymbolId> terms) {
```
bachase, line 59: Should we have a comment saying what this is doing (like define what we mean by the cannonical order?)

src/clifft/sampling/plan.cc
```diff
@@ -0,0 +1,497 @@
...
+
+uint32_t index(NoiseSiteId site) {
+    return static_cast<uint32_t>(site);
+}
+
+uint32_t index(InstrumentSiteId site) {
+    return static_cast<uint32_t>(site);
+}
+
+std::vector<SymbolId> xor_terms(const std::vector<SymbolId>& left,
```
bachase, line 38: Should we have a comment on what this function is doing?

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,207 @@
...
+// when the physical circuit contains many qubits.
+static_assert(kDenseActiveWidthLimit <= std::numeric_limits<uint64_t>::digits);
+
+struct ActivePauli {
+    uint64_t x = 0;
```
bachase, line 84: Are there existing Pauli Mask classes we should reuse rather than rewriting this one?

src/clifft/sampling/plan.h
```diff
@@ -0,0 +1,207 @@
...
+    // Throws std::invalid_argument when the plan is structurally inconsistent.
+    // This checks active-width transitions, masks, stable ids, record slots,
+    // symbol definitions, and assignment-before-use.
+    void validate() const;
+
+    // Deterministic, executor-independent inspection for tests and diagnostics.
+    [[nodiscard]] std::string inspect() const;
+};
+
+[[nodiscard]] uint32_t predicted_dense_passes(const SamplingAction& action);
```
bachase, line 207: Document what this function is does

PR #250, opened 2026-08-05
feat(sampling): plan rotations and measurements from HIR

### LINKED ISSUE #248: Define SamplingPlan v0 semantics and structural planner

bachase:
## Summary

- add an internal deterministic optimized-HIR to SamplingPlan planner for T gates, phase rotations, and measurements
- compute stabilizer-coordinate promotions and measurement compaction during planning, then rewrite all future Paulis into the new coordinates
- preserve geometric signs, balanced-rotation global factors, stable record slots, and measurement-branch dependencies
- emit direct multi-coordinate active-Pauli actions without active Clifford sweep actions
- reject every unsupported HIR operation with an actionable operation type and index
- leave the legacy compiler, SVM, public Program API, and production routing unchanged

This completes the structural planner work in #248 on top of the semantic plan foundation from #249. Instrument boundaries remain represented by the plan IR; instrument and continuation planning are deliberately still unsupported by this first planner subset.

Closes #248

## Testing

- `cmake --build build --target clifft_tests -j4`
- `./build/tests/clifft_tests "Sampling planner*" --reporter compact` — 79 assertions in 14 test cases
- `./build/tests/clifft_tests "~[bench]" --reporter compact` — 209008 assertions in 1164 test cases

`just build-wasm` could not run locally because the Docker daemon was unavailable; the planner is included in the normal core source list used by native and WASM builds.

### DIFF
--- a/src/clifft/sampling/planner.cc
+++ b/src/clifft/sampling/planner.cc
@@ -1,5 +1,6 @@
 #include "clifft/sampling/planner.h"
 
+#include "clifft/util/hir_introspection.h"
 #include "clifft/util/numeric.h"
 #include "clifft/util/stim_mask.h"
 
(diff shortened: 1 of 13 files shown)


src/clifft/sampling/planner.cc
```diff
@@ -0,0 +1,485 @@
...
+
+struct PendingMeasurement {
+    Pauli body;
+    AffineBool sign;
+    RecordSlot record{};
+};
+
+using PendingOperation = std::variant<PendingRotation, PendingMeasurement>;
+
+std::string op_type_name(OpType type) {
```
bachase: Isn't this type to string defined elsewhere already for us to reuse?

src/clifft/sampling/planner.cc
```diff
@@ -0,0 +1,485 @@
...
+            return "EXP_VAL";
+        case OpType::INSTRUMENT:
+            return "INSTRUMENT";
+        case OpType::NUM_OP_TYPES:
+            return "NUM_OP_TYPES";
+    }
+    return "UNKNOWN";
+}
+
+[[noreturn]] void unsupported_operation(OpType type, size_t index) {
```
bachase: Why is this a function? Seems like its only used in one spot; cant we just inline it there?

bachase: Or are you trying to avoid bloating that callsite since it should be rare?

PR #256, opened 2026-08-05
feat(sampling): add direct-Pauli state kernels

### LINKED ISSUE #253: Implement one-allocation SoA direct-Pauli kernels
Parent tracker: #247
Architecture discussion: #236
Planner foundation: #248

## Objective

Build the correctness-first CPU execution substrate for `SamplingPlan` without yet adding a plan interpreter or production routing. This increment isolates the state layout and direct active-Pauli mathematics so they can be reviewed and tested independently.

## Scope

- Add an internal one-shot CPU active state with one aligned allocation split into real and imaginary coefficient planes.
- Preallocate from the declared maximum active width; provide reset and move semantics without allocation in ordinary kernel calls.
- Preserve the scalar/global-phase and deferred-normalization information needed for exact semantic parity, including signed identity rotations.
- Derive execution-specific Pauli descriptors before hot kernels, including pairing, phase/parity, and diagonal-versus-general information.
- Implement correctness-first direct kernels for:
  - active-Pauli rotation;
  - dormant promotion with its introducing rotation;
  - active-Pauli branch probabilities, forced-branch collapse, normalization, and width compaction.
- Keep measurement branch selection outside the kernels so forced-outcome replay and normal RNG execution can share the same state transition.
- Reconcile `predicted_dense_passes()` with the state traversals required by the implemented kernels instead of designing kernels around the current estimate.
- Reuse or extract backend-independent test helpers where useful. Use existing Clifft statevector and legacy-kernel tests as secondary differential coverage while keeping the new implementation independent of SVM lowering.

## Non-goals

- No `SamplingPlan` dispatch loop, RNG ownership, symbolic-expression evaluation, record storage, or public sampling API.
- No production routing or change to the legacy SVM.
- No WASM or browser-playground integration; the new path remains native-only until the playground migration is designed.
- No packed shots, product components, GPU lowering, alternate layouts, native scheduling, hand-written SIMD, or multithreaded tuning.
- No planner-scaling restructure or symbolic-relation reduction from the deferred #247 follow-ups.
- No performance claim from isolated kernel timing; representative measurement starts with the scalar vertical slice.

## Acceptance criteria

- Exhaustive small-width tests cover Hermitian active Pauli masks, positive and negative signs, representative angles, identity operations, promotions, and both measurement branches.
- Rotation and measurement results agree with an independent dense linear-algebra oracle, including branch probabilities, normalized post-measurement states, and compaction conventions.
- Existing statevector and legacy-kernel fixtures are reused where they provide meaningful independent or secondary checks.
- Active width and capacity invariants are validated, and kernel calls preserve allocation addresses and capacity.
- No tableau evolution, commutation analysis, localization, dependency discovery, or dynamic allocation occurs inside a kernel.
- The implementation is included in native builds, with the production compiler, SVM, and WASM playground behavior unchanged.

bachase:
Closes #253

## Summary

- add an internal one-allocation, page-aligned active state with stable 64-byte-aligned real, imaginary, and measurement-work arrays;
- share the SVM's page and huge-page allocation policy through a move-only allocation owner used by both dense states;
- preserve a separate finite global scalar while keeping coefficient planes normalized, including exact signed-identity rotation behavior;
- prepare fixed-size Pauli descriptors before execution and add allocation-free direct kernels for rotations, promotions, measurement probabilities, forced collapse, normalization, and width compaction;
- keep per-shot reset and ordinary kernel entry points exception-free, with validated inputs and assertion-only internal preconditions;
- keep branch choice outside measurement kernels so ordinary sampling and forced replay can share the same transition;
- update active-measurement diagnostics to predict the probability and collapse traversals.

The production compiler and legacy SVM execution semantics are unchanged; the SVM allocation ownership is refactored into the shared utility described above. This PR intentionally does not add a SamplingPlan interpreter, RNG ownership, record or symbol storage, or production routing. The new SamplingPlan path is native-only while the browser playground remains on the legacy SVM.

## Validation

- `cmake --build build --target clifft_tests -j 4`
- `./build/tests/clifft_tests "Page-aligned*" --reporter compact` — 14 assertions in 2 test cases
- `./build/tests/clifft_tests "Sampling kernel*" --reporter compact` — 147,387 assertions in 6 test cases
- `./build/tests/clifft_tests "Sampling plan*" --reporter compact` — 153 assertions in 31 test cases
- `./build/tests/clifft_tests "~[bench]" --reporter compact` — 356,425 assertions in 1,176 test cases

The exhaustive small-width coverage reuses Clifft's existing dense statevector rotation oracle and independently checks every Hermitian Pauli, representative positive and negative angles and signs, promotions, both measurement branches, probabilities, normalized states, and compaction conventions. A composed sequence also covers promotion after a collapse with stale data above the reduced active width. True plan-level differential replay against the legacy SVM remains part of the next end-to-end scalar slice, where both paths can execute the same input.

WASM CI remains a regression check for the existing playground. The SamplingPlan planner and execution substrate are intentionally excluded from that target until the playground migration is designed.

### DIFF
--- a/src/clifft/sampling/soa_kernels.cc
+++ b/src/clifft/sampling/soa_kernels.cc
@@ -1,0 +1,230 @@
+#include "clifft/sampling/soa_kernels.h"
+
+#include "clifft/util/numeric.h"
+
+#include <bit>
+#include <cmath>
+#include <complex>
+#include <numbers>
+#include <stdexcept>
+#include <string>
+
+namespace clifft::sampling {
+
+namespace {
+
+constexpr double kInvSqrt2 = 0.707106781186547524400844362104849039;
+
+uint64_t width_mask(uint32_t active_width) {
+    if (active_width >= kDenseActiveWidthLimit) {
+        throw std::invalid_argument("prepared Pauli active width must be below " +
+                                    std::to_string(kDenseActiveWidthLimit));
+    }
+    return active_width == 0 ? 0 : (uint64_t{1} << active_width) - 1;
+}
+
+void validate_descriptor_width(const SoaState& state, const PreparedPauli& pauli) {
+    if (state.active_width() != pauli.active_width) {
+        throw std::invalid_argument("prepared Pauli width does not match the active state");
+    }
+}
+
+std::complex<double> phase_at(const PreparedPauli& pauli, uint64_t basis) {
+    return (std::popcount(basis & pauli.z) & 1U) != 0 ? -pauli.even_phase : pauli.even_phase;
+}
+
+uint64_t insert_zero_bit(uint64_t packed, uint32_t pivot) {
+    const uint64_t lower_mask = (uint64_t{1} << pivot) - 1;
+    return (packed & lower_mask) | ((packed & ~lower_mask) << 1);
+}
+
+uint64_t diagonal_source(const PreparedMeasurement& measurement, uint64_t packed, bool branch) {
+    const uint64_t without_pivot = insert_zero_bit(packed, measurement.pivot);
+    const bool other_parity =
+        (std::popcount(without_pivot & measurement.z_without_pivot) & 1U) != 0;
+    const bool pivot_value = branch != other_parity;
+    return without_pivot | (static_cast<uint64_t>(pivot_value) << measurement.pivot);
+}
+
+std::complex<double> load(const SoaState& state, uint64_t index) {
+    return {state.real_data()[index], state.imag_data()[index]};
+}
+
+void store(SoaState& state, uint64_t index, std::complex<double> value) {
+    state.real_data()[index] = value.real();
+    state.imag_data()[index] = value.imag();
+}
+
+std::complex<double> compact_nondiagonal(const SoaState& state,
+                                         const PreparedMeasurement& measurement, uint64_t packed,
+                                         bool branch) {
+    const uint64_t source0 = insert_zero_bit(packed, measurement.pivot);
+    const uint64_t source1 = source0 ^ measurement.pauli.x;
+    const double eigenvalue = branch ? -1.0 : 1.0;
+    const std::complex<double> coefficient1 =
+        eigenvalue * std::conj(phase_at(measurement.pauli, source0));
+    return kInvSqrt2 * (load(state, source0) + coefficient1 * load(state, source1));
+}
+
+}  // namespace
+
+PreparedPauli prepare_pauli(ActivePauli pauli, uint32_t active_width) {
+    const uint64_t valid_bits = width_mask(active_width);
+    if ((pauli.x & ~valid_bits) != 0 || (pauli.z & ~valid_bits) != 0) {
+        throw std::invalid_argument("prepared Pauli has bits outside its active width");
+    }
+
+    static constexpr std::complex<double> kIPowers[4] = {
+        {1.0, 0.0}, {0.0, 1.0}, {-1.0, 0.0}, {0.0, -1.0}};
+    const uint32_t overlap = std::popcount(pauli.x & pauli.z);
+    return PreparedPauli{active_width, pauli.x, pauli.z,
+                         pauli.x == 0 ? 0 : pauli.x & (~pauli.x + 1), kIPowers[overlap & 3U]};
+}
+
+PreparedRotation prepare_rotation(ActivePauli pauli, uint32_t active_width, double half_turns) {
+    if (!is_finite_robust(half_turns)) {
+        throw std::invalid_argument("prepared rotation angle must be finite");
+    }
+    const double angle = std::numbers::pi * half_turns / 2.0;
+    return PreparedRotation{prepare_pauli(pauli, active_width), std::cos(angle), std::sin(angle)};
+}
+
+PreparedPromotion prepare_promotion(double half_turns) {
+    if (!is_finite_robust(half_turns)) {
+        throw std::invalid_argument("prepared promotion angle must be finite");
+    }
+    const double angle = std::numbers::pi * half_turns / 2.0;
+    return PreparedPromotion{std::cos(angle), std::sin(angle)};
+}
+
+PreparedMeasurement prepare_measurement(ActivePauli pauli, uint32_t active_width, uint32_t pivot) {
+    PreparedPauli prepared = prepare_pauli(pauli, active_width);
+    if (active_width == 0 || prepared.is_identity()) {
+        throw std::invalid_argument("cannot prepare an active identity measurement");
+    }
+    if (pivot >= active_width) {
+        throw std::invalid_argument("prepared measurement pivot is outside its active width");
+    }
+    const uint64_t pivot_bit = uint64_t{1} << pivot;
+    const bool pivot_is_valid =
+        prepared.x != 0 ? (prepared.x & pivot_bit) != 0 : (prepared.z & pivot_bit) != 0;
+    if (!pivot_is_valid) {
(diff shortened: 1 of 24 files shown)


src/clifft/sampling/soa_state.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include <complex>
+#include <cstddef>
+#include <cstdint>
+#include <span>
+
+namespace clifft::sampling {
+
+// One-shot CPU coefficient state for the direct-Pauli executor. The coefficient
+// planes stay normalized; a separate scalar preserves plan-level global factors.
```
bachase: What does "planes" mean? This sounds like new jargon you are introducing. Please avoid jargon

src/clifft/sampling/soa_state.h
```diff
@@ -0,0 +1,83 @@
...
+// One aligned allocation also contains the scratch planes needed to compact a
+// non-diagonal measurement without aliasing input.
+class SoaState {
```
bachase: Stepping back ... SoA is an implementaiton detail. I'm not sure I want to name this class (or the kernels or the files) with the SoA prefix. If in the future we test and want AoS or something else, we can figure out a naming convention then. But for now, I'd rather this just be the State or SampleState or something. 

If we retire the existing SVM backend in the future, this might be the new SVM, in which case we can always rename it.

src/clifft/sampling/soa_state.h
```diff
@@ -0,0 +1,83 @@
...
+  public:
+    explicit SoaState(uint32_t max_active_width, uint32_t initial_active_width = 0,
```
bachase: We should add a communit documenting that max_active_width is immediately allocated.

src/clifft/sampling/soa_kernels.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "clifft/sampling/plan.h"
+#include "clifft/sampling/soa_state.h"
+
+#include <complex>
+#include <cstdint>
+
+namespace clifft::sampling {
+
+// Fixed-size execution descriptor derived before hot dispatch. It contains the
```
bachase: Again this is a little too jargony, maybe say this is "Represents a Pauli operation to be applied to the state vector. It contains the mask relationships and indexing inot that vector, but is agnostic to the lower-level storage and execution details"

src/clifft/sampling/soa_kernels.cc
```diff
@@ -0,0 +1,230 @@
...
+    if (active_width >= kDenseActiveWidthLimit) {
+        throw std::invalid_argument("prepared Pauli active width must be below " +
+                                    std::to_string(kDenseActiveWidthLimit));
+    }
+    return active_width == 0 ? 0 : (uint64_t{1} << active_width) - 1;
+}
+
+void validate_descriptor_width(const SoaState& state, const PreparedPauli& pauli) {
+    if (state.active_width() != pauli.active_width) {
+        throw std::invalid_argument("prepared Pauli width does not match the active state");
```
bachase: is this throw in a hot path at all? Should it be an assert instead?

src/clifft/sampling/state.h
```diff
@@ -0,0 +1,83 @@
...
+    void set_active_width(uint32_t width);
+
+  private:
+    void release() noexcept;
+    void move_from(SoaState&& other) noexcept;
+
+    double* allocation_ = nullptr;
+    double* real_ = nullptr;
+    double* imag_ = nullptr;
+    double* scratch_real_ = nullptr;
```
bachase, line 74: Is scratch necessary or can those operations be done inline? Just would be nice to avoid extra memory overhead and fit more in L1 cache

tests/test_sampling_kernels.cc
```diff
@@ -0,0 +1,395 @@
...
+        std::popcount(x & z) + 2U * (static_cast<uint32_t>(std::popcount(basis & z)) & 1U);
+    return kIPowers[exponent & 3U];
+}
+
+struct ProjectedBranch {
+    double probability = 0.0;
+    std::vector<std::complex<double>> normalized;
+};
+
+ProjectedBranch dense_project(const std::vector<std::complex<double>>& input, uint64_t x,
```
bachase, line 120: seems like a lot of custom code/dense code for just this test.

Are there existing C++ helpers/utilities that we should be reusing? (ok if not)

tests/test_sampling_soa_kernels.cc
```diff
@@ -0,0 +1,395 @@
...
+void require_vectors_close(const std::vector<std::complex<double>>& actual,
+                           const std::vector<std::complex<double>>& expected) {
+    REQUIRE(actual.size() == expected.size());
+    for (size_t i = 0; i < actual.size(); ++i) {
+        CAPTURE(i);
+        check_complex(actual[i], expected[i], kTolerance);
+    }
+}
+
+uint64_t insert_zero_bit(uint64_t packed, uint32_t pivot) {
```
bachase: Isn't this a function i soa_kernels.cc already? Whats the point of a second copy? Are we doing this copy-paste elsewhere?

src/clifft/sampling/soa_kernels.h
bachase: Reworded this as a state-vector Pauli description whose masks and precomputed index values identify affected coefficients without referring to State storage.

src/clifft/sampling/soa_state.h
bachase: Added an explicit constructor comment that all coefficient and temporary storage for max_active_width is allocated immediately, even when the initial width is smaller.

src/clifft/sampling/soa_kernels.cc
bachase: Yes, this is reached once per kernel dispatch. Prepared-descriptor width and promotion-capacity preconditions are now debug assertions, consistent with the legacy hot kernels. Validation of external or compiler-produced values remains in the prepare functions before dispatch.

src/clifft/sampling/soa_state.h
bachase: Reworded this in terms of the real and imaginary coefficient arrays and removed the plane terminology.

src/clifft/sampling/state.h
bachase: Retained the preallocated temporary arrays for now. With an arbitrary non-diagonal measurement pivot, direct writes into the coefficient prefix can form overwrite cycles, so a simple forward or reverse in-place loop is not correct in general. Untouched temporary arrays do not occupy L1, though they do add allocation size and collapse bandwidth; that is worth measuring after the end-to-end path exists before considering a cycle-based in-place implementation.

src/clifft/sampling/soa_state.h
bachase: Agreed. Renamed SoaState to sampling::State, renamed the state and kernel source files without the SoA prefix, and made the test naming implementation-neutral as well.

tests/test_sampling_kernels.cc
bachase: The generic dense matrix-vector multiply is now shared through test_helpers. The projector and compaction code remains local because it is the independent oracle for the new kernels; reusing production transition logic there would weaken the exhaustive comparison.

tests/test_sampling_soa_kernels.cc
bachase: The test now uses a separately expressed expand_index_without_pivot helper and documents why it remains independent of the kernel helper. A similar utility exists under the legacy SVM header, but making the new path or its oracle include that legacy implementation would couple the paths; extracting a shared public utility for this two-line operation did not seem worthwhile.

PR #258, opened 2026-08-06
feat(sampling): add scalar plan executor

### LINKED ISSUE #257: Add internal scalar SamplingPlan interpreter
Parent: #247

## Goal

Land an internal scalar end-to-end vertical slice from `SamplingPlan` through the direct-Pauli kernels for the rotation-and-measurement subset currently emitted by the planner. Keep the legacy SVM as the production backend and differential oracle.

## Scope

- Validate and lower a semantic `SamplingPlan` once into prepared executable actions before shot execution.
- Preallocate coefficient state, symbols, visible and hidden records, RNG state, and required counters before entering per-shot dispatch.
- Evaluate affine signs and outcomes without allocation.
- Execute active rotations, dormant promotions, active and dormant measurements, classical record writes, and derived-symbol definitions.
- Apply the measurement dust and deterministic RNG policy recorded in #247.
- Keep all ordinary dispatch, kernels, and per-shot reset allocation-free and exception-free.
- Provide an internal test-facing entry point; do not route production sampling or change the public API.

## Acceptance

- Fixed-seed differential tests reuse representative legacy HIR-to-SVM measurement scenarios and compare record behavior where both paths support the same input.
- Independent kernel-oracle coverage remains green.
- Tests cover deterministic measurements, dormant random measurements, active collapse, repeated/classical records, affine signs, global scalar propagation, dust handling, RNG consumption, reset, and repeated shots.
- Exhaustive variant visitors fail at compile time when an action type is added without explicit handling.
- Native tests and pre-commit pass; WASM continues using only the legacy path.

## Out of scope

- Forced-outcome replay and log-probability accounting; this is the immediately following increment and must reuse the same branch classification and transition logic.
- Conditional Paulis, symbolic noise, state-dependent instruments, production routing, public APIs, SIMD specialization, planner restructuring, and collapse-storage optimization.

bachase:
Closes #257.

## Summary

- lower a validated SamplingPlan once into prepared kernel descriptors and flattened affine-expression terms
- add a reusable scalar executor with preallocated coefficient, symbol, and visible/hidden record storage
- keep shot reset, dispatch, affine evaluation, measurement sampling, and kernels allocation- and exception-free
- centralize active-measurement dust classification for reuse by forced replay
- leave the production SVM and WASM playground routing unchanged

## Validation

- exact seeded differential records against the legacy SVM for deterministic, dormant, active, repeated, signed, and multi-coordinate circuits
- focused executor tests: 732 assertions across 7 cases
- native correctness suite excluding opt-in benchmarks: 357155 assertions across 1183 cases

### DIFF
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -39,10 +39,12 @@ MeasurementBranchClassification classify_measurement_branch(
            "measurement probabilities must be finite, nonnegative, and nonzero");
     const double epsilon = kMeasurementDustEpsilon * total;
     if (probabilities.one <= epsilon) {
-        return {.kind = MeasurementBranchKind::Zero, .clamped_dust = probabilities.one > 0.0};
+        return {.kind = MeasurementBranchKind::DeterministicZero,
+                .clamped_dust = probabilities.one > 0.0};
     }
     if (probabilities.zero <= epsilon) {
-        return {.kind = MeasurementBranchKind::One, .clamped_dust = probabilities.zero > 0.0};
+        return {.kind = MeasurementBranchKind::DeterministicOne,
+                .clamped_dust = probabilities.zero > 0.0};
     }
     return {.kind = MeasurementBranchKind::Random};
 }
@@ -210,10 +212,10 @@ bool Executor::sample_active_branch(MeasurementProbabilities probabilities) noex
     switch (classification.kind) {
         case MeasurementBranchKind::Random:
             return rng_.next_double() * probabilities.total() >= probabilities.zero;
-        case MeasurementBranchKind::Zero:
+        case MeasurementBranchKind::DeterministicZero:
             dust_clamps_ += static_cast<uint64_t>(classification.clamped_dust);
             return false;
-        case MeasurementBranchKind::One:
+        case MeasurementBranchKind::DeterministicOne:
             dust_clamps_ += static_cast<uint64_t>(classification.clamped_dust);
             return true;
     }
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -39,12 +39,10 @@ MeasurementBranchClassification classify_measurement_branch(
            "measurement probabilities must be finite, nonnegative, and nonzero");
     const double epsilon = kMeasurementDustEpsilon * total;
     if (probabilities.one <= epsilon) {
-        return {.kind = MeasurementBranchKind::Zero,
-                .clamped_dust = probabilities.one > 0.0};
+        return {.kind = MeasurementBranchKind::Zero, .clamped_dust = probabilities.one > 0.0};
     }
     if (probabilities.zero <= epsilon) {
-        return {.kind = MeasurementBranchKind::One,
-                .clamped_dust = probabilities.zero > 0.0};
+        return {.kind = MeasurementBranchKind::One, .clamped_dust = probabilities.zero > 0.0};
     }
     return {.kind = MeasurementBranchKind::Random};
 }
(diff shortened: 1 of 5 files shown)


src/clifft/sampling/executor.h
bachase, line 1: Should we add a file level comment explaining the role of the executor and high-level how it works?

src/clifft/sampling/executor.h
```diff
@@ -0,0 +1,139 @@
...
+#include <span>
+#include <variant>
+#include <vector>
+
+namespace clifft::sampling {
+
+// Classifies an active measurement before ordinary sampling or forced replay
+// chooses a branch.
+enum class MeasurementBranchKind : uint8_t {
+    Random,
```
bachase, line 26: Should we add trailing comments (or something above) -- sounds like Zero or One is a deterministic measurement?

src/clifft/sampling/executor.h
```diff
@@ -0,0 +1,139 @@
...
+    uint32_t num_visible_records_ = 0;
+    uint32_t num_hidden_records_ = 0;
+    uint32_t num_symbols_ = 0;
+    std::complex<double> global_weight_ = {1.0, 0.0};
+    std::vector<uint32_t> expression_terms_;
+    std::vector<uint32_t> presampled_symbols_;
+    std::vector<Action> actions_;
+};
+
+// Holds every mutable value needed to execute repeated shots of one prepared
```
bachase: Should we also add a comment that ties this back to the physics, e.g. "every mutable value needed to execut ... " which values are these in the physics language of this simulator?

src/clifft/sampling/executor.h
```diff
@@ -0,0 +1,139 @@
...
+
+struct MeasurementBranchClassification {
+    MeasurementBranchKind kind = MeasurementBranchKind::Random;
+    bool clamped_dust = false;
+};
+
+[[nodiscard]] MeasurementBranchClassification classify_measurement_branch(
+    MeasurementProbabilities probabilities) noexcept;
+
+// Owns the execution-specific lowering of one validated SamplingPlan. Kernel
```
bachase: What does execution-specific mean in this context?

src/clifft/sampling/executor.h
```diff
@@ -0,0 +1,139 @@
...
+
+    [[nodiscard]] std::span<const uint8_t> visible_records() const {
+        return std::span<const uint8_t>(records_).first(plan_.num_visible_records_);
+    }
+    [[nodiscard]] std::span<const uint8_t> hidden_records() const {
+        return std::span<const uint8_t>(records_).subspan(plan_.num_visible_records_);
+    }
+    [[nodiscard]] std::span<const uint8_t> symbols() const { return symbols_; }
+    [[nodiscard]] const State& state() const { return state_; }
+    [[nodiscard]] uint64_t dust_clamps() const { return dust_clamps_; }
```
bachase, line 137: Should we have comments explaining what these are returning? Most are self-explanatory but dust_clumps -- should that say the number of times had to clamp dust level probability mass or similar?

src/clifft/sampling/executor.cc
```diff
@@ -0,0 +1,228 @@
...
+#include <type_traits>
+
+namespace clifft::sampling {
+
+namespace {
+
+template <typename>
+inline constexpr bool kAlwaysFalse = false;
+
+uint32_t index(SymbolId id) {
```
bachase, line 19: I feel like these and kAlwaysFalse are in a few files. Maybe that is ok, but ... is that a sign these should be elsewhere?

src/clifft/sampling/executor.cc
```diff
@@ -0,0 +1,228 @@
...
+    }
+    if (num_expression_terms > std::numeric_limits<uint32_t>::max()) {
+        throw std::length_error("sampling executable expression storage exceeds uint32 range");
+    }
+    expression_terms_.reserve(num_expression_terms);
+    actions_.reserve(plan.actions.size());
+    presampled_symbols_.reserve(plan.symbols.size());
+    for (uint32_t symbol = 0; symbol < plan.symbols.size(); ++symbol) {
+        if (plan.symbols[symbol].kind == SymbolKind::Presampled) {
+            presampled_symbols_.push_back(symbol);
```
bachase, line 94: Asking to understand -- how is the index into presampled_symbols related to plan.symbols? We reserve the whole size above on line 89, but then only some of these get pushed_back

tests/test_sampling_executor.cc
```diff
@@ -0,0 +1,210 @@
...
+                      MeasureActivePauli{
+                          {0, 1}, 0, SymbolId{0}, AffineBool::symbol(SymbolId{0}), RecordSlot{0}}},
+        PlannedAction{
+            0, 0,
+            MeasureDormantRandom{1, SymbolId{1}, AffineBool::symbol(SymbolId{1}), RecordSlot{1}}},
+    };
+    return plan;
+}
+
+void require_matches_legacy(std::string_view circuit_text, uint32_t shots, uint64_t seed) {
```
bachase, line 67: Asking to understand -- what happens if we deprecate and remove the legacy pipelines? What will we test instead?

src/clifft/sampling/executor.h
bachase: Reworded this as the CPU lowering for direct-Pauli kernels and distinguished its one-time preparation from per-shot execution.

src/clifft/sampling/executor.cc
bachase: Clarified that this is a dense input-position-to-SymbolId map in ascending SymbolId order. Reserving num_symbols is only an upper bound on capacity; the constructor pushes only presampled symbols.

src/clifft/sampling/executor.h
bachase: Expanded the Executor comment in physics terms: active-coordinate coefficients and global scalar, stochastic Boolean dependencies, records, RNG, and numerical-dust telemetry.

src/clifft/sampling/executor.h
bachase: Documented that this counts positive branch mass classified as numerical dust and accumulates across shots.

src/clifft/sampling/executor.h
bachase: Added a file-level overview covering the prepared CPU lowering, per-shot symbol evaluation, state evolution, sampling, record writes, and preallocated-storage contract.

src/clifft/sampling/executor.h
bachase: Renamed the values to DeterministicZero and DeterministicOne and clarified that they denote Pauli eigenvalue branches before affine correction produces the physical record.

tests/test_sampling_executor.cc
bachase: The legacy comparison is transitional. Before legacy removal, the replacement should combine forced-replay exact branch probabilities, Qiskit Aer for small non-Clifford cases, Stim for Clifford and stochastic behavior, and randomized or equivalent-circuit tests. This update also adds the missing sampled-branch-to-later-state-action regression; broader Python oracle tests fit once this path is exposed through the public selector or API.

src/clifft/sampling/executor.cc
bachase: I am leaving these helpers local for now. They occur in only three sampling translation units and remain tiny implementation details; extracting them would add shared coupling without yet giving the typed IDs a meaningful common operation set.

Issue #247, opened 2026-08-05
Complete the symbolic sampler migration and remove the legacy SVM


bachase:
## Goal

Make the symbolic sampler Clifft's sole production sampling path, preserve the supported public APIs and result semantics, and remove the legacy SVM once all native and browser consumers have migrated.

Discussion #298 records the production architecture and migration decisions. Issue #301 tracks the remaining documentation and release-facing work.

## Current status

- [x] Symbolic capability coverage is complete for ordinary and survivor sampling, postselection, replay and record probabilities, exact final-state queries, noise, expectation values, importance sampling, instruments, and noncomputational continuations.
- [x] The representative symbolic-versus-legacy performance gate is complete in #280.
- [x] Required x86 AVX2 rotation, active-measurement, and new-X instrument coverage landed through #285, #309, #310, and #311.
- [x] Executable-plan compilation, concrete action specialization, kernel dispatch, and production-quality contract cleanup landed through #316, #319, #320, #323, #324, #325, and #326.
- [x] The public native APIs use the symbolic sampler following #321.
- [x] WASM and the playground use the symbolic pipeline and its plan inspection surfaces following #330, #331, and #336.
- [x] The legacy SVM, bytecode implementation, and production fallback were removed in #334.
- [x] The factored-state theory, symbolic architecture, and active-width terminology were updated in #308, #337, and #338.
- [ ] The documentation, final benchmark presentation, migration article, and release validation tracked by #301 remain.

The representative scalar, AVX2, and AVX-512 release paths are at parity or faster than the removed legacy SVM across the migration corpus apart from the accepted small QV-10 tail recorded in #280. Packed-shot batching, product components, intra-shot parallelism, and additional platform kernels are independent follow-ups rather than migration gates.

## Remaining work

1. Replace obsolete instruction, bytecode-pass, and SVM reference material with current HIR and semantic `SamplingPlan` concepts (#304).
2. Finish the symbolic-only user guides and Python API documentation (#305).
3. Refresh release-target performance material (#302) and publish the migration/performance update (#322).
4. Complete the final legacy-reference, link, example, compatibility, and release-validation sweep (#303), then close #301 and this tracker.

Open focused issues and PRs for these documentation slices and link them here. Discussion #298 remains the architecture and decision record; this issue remains the migration completion gate.

## Architectural requirements

- `SamplingPlan` remains the executor-independent semantic boundary.
- Coordinate changes and symbolic dependency discovery happen during planning, not hot execution.
[...]

PR #261, opened 2026-08-06
feat(sampling): add forced record replay

### LINKED ISSUE #260: Add forced-record replay to the scalar sampling executor
## Context

Parent tracker: #247
Architecture discussion: #236

PR #258 added allocation-free scalar sampling for prepared SamplingPlan actions. The next migration increment needs to replay a specified record and compute its joint log probability without depending on matching RNG consumption with the legacy SVM.

## Scope

- Add an internal ReplayResult and replay_shot API to the scalar executor.
- Force every plan record slot in visible-then-hidden order; probabilities are conditional on supplied presampled symbols.
- Lower measurement outcomes to branch-free affine corrections during ExecutablePlan construction so replay can recover internal branches without runtime dependency analysis.
- Use the same active-measurement dust classification as ordinary sampling: a dust branch is unreachable, the survivor contributes zero log probability, and ordinary reachable branches contribute log(p_branch / total).
- Treat dormant-random branches as probability 1/2 and reject inconsistent deterministic or repeated records.
- Preserve allocation-free, exception-free hot execution and consume no RNG during replay.
- Keep the sampling loop free of a runtime replay-mode branch.

## Validation

- Reuse legacy forced-execution and record-probability cases for deterministic, random, active, impossible, repeated, and dust-clamped outcomes.
- Differentially compare replay reachability and log probabilities with the legacy SVM on arbitrary-angle and multi-coordinate MPP circuits.
- Enumerate small record spaces and verify reachable probabilities sum to one.
- Verify replayed branch symbols continue to control later state actions and replay does not advance the sampling RNG.

## Deferred

- Public backend selection and Python integration.
- Marginalization over unspecified hidden records.
- Noise, feedback, instruments, and public record-probability API routing.
- SIMD and planner-performance work.

bachase:
Closes #260.

Parent tracker: #247
Architecture discussion: #236

## Summary

- lower measurement outcomes to branch-free affine corrections during executable-plan preparation
- add allocation-free, exception-free forced replay for visible and hidden records without consuming RNG
- preserve the active-measurement dust policy and report conditional joint log probabilities
- compile sampling and replay as separate template specializations so ordinary sampling has no runtime replay-mode branch

## Validation

- differential replay probabilities and reachability against the legacy SVM for repeated measurements, arbitrary-angle rotations, multi-coordinate MPP, and entangled records
- active/dormant probability accumulation, affine branch inversion, deterministic mismatch, dust, and RNG-preservation tests
- small record-space normalization check
- full native non-benchmark suite: 1,191 cases and 357,905 assertions
- focused release-build replay tests
- pre-commit hooks

### DIFF
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -6,6 +6,7 @@
 #include <cassert>
 #include <cmath>
 #include <limits>
+#include <numbers>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
@@ -16,6 +17,8 @@ namespace {
 
 template <typename>
 inline constexpr bool kAlwaysFalse = false;
+
+constexpr double kLogHalf = -std::numbers::ln2;
 
 uint32_t index(SymbolId id) {
     return static_cast<uint32_t>(id);
@@ -230,7 +233,7 @@ ReplayResult Executor::execute_actions(std::span<const uint8_t> forced_records) 
                     bool branch = false;
                     if constexpr (ForceRecords) {
                         branch = (forced_records[typed.record] != 0) ^ correction;
-                        result.log_probability += std::log(0.5);
+                        result.log_probability += kLogHalf;
                     } else {
                         branch = sample_dormant_branch();
                     }
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -72,6 +72,8 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
                     num_expression_terms += typed.sign.terms().size();
                 } else if constexpr (std::is_same_v<T, MeasureActivePauli> ||
                                      std::is_same_v<T, MeasureDormantRandom>) {
+                    // The current measurement branch is stored separately so
+                    // replay can solve for it from the requested record.
                     num_expression_terms += typed.outcome.terms().size() - 1;
                 } else if constexpr (std::is_same_v<T, RecordClassical>) {
                     num_expression_terms += typed.outcome.terms().size();
@@ -209,13 +211,13 @@ ReplayResult Executor::execute_actions(std::span<const uint8_t> forced_records) 
                     bool branch = false;
                     if constexpr (ForceRecords) {
                         branch = (forced_records[typed.record] != 0) ^ correction;
-                        const ForcedBranchResult forced =
+                        const std::optional<double> log_increment =
                             force_active_branch(probabilities, branch);
-                        if (!forced.reachable) {
+                        if (!log_increment.has_value()) {
                             result.reachable = false;
                             return;
                         }
-                        result.log_probability += forced.log_increment;
+                        result.log_probability += *log_increment;
                     } else {
                         branch = sample_active_branch(probabilities);
                     }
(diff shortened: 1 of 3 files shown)


src/clifft/sampling/executor.h
```diff
@@ -33,6 +33,12 @@ struct MeasurementBranchClassification {
     bool clamped_dust = false;
 };
 
+struct ReplayResult {
```
bachase, line 39: Let's add a comment above the struct on what this is for/represents/

src/clifft/sampling/executor.h
```diff
@@ -72,13 +78,16 @@ class ExecutablePlan {
 
     struct ExecuteActiveMeasurement {
         PreparedMeasurement measurement;
-        PreparedExpression outcome;
+        // The physical record is branch XOR this expression. Removing the
+        // newly defined branch at preparation time lets replay invert that
```
bachase: This is a bit jargony -- is there a clearer phrasing?

bachase: Something like "The physical record is the branch measurement XOR the correction. The correction encapsulates all the sample time (non-measurement) stochastic events that may cause the error free measurement to flip"

src/clifft/sampling/executor.h
```diff
@@ -124,6 +134,13 @@ class Executor {
     // Values correspond to presampled plan symbols in ascending SymbolId order.
     void run_shot(std::span<const uint8_t> presampled_values = {}) noexcept;
 
+    // Forces one Boolean value for every plan record slot, ordered as visible
```
bachase: Replays or resamples a circuit where each outcome is forced to be a specific value. This enables <list workflows>

src/clifft/sampling/executor.h
```diff
@@ -137,8 +154,20 @@ class Executor {
     [[nodiscard]] uint64_t dust_clamps() const { return dust_clamps_; }
 
   private:
+    struct ForcedBranchResult {
```
bachase: This looks like the ReplayResult above..should we unify?

src/clifft/sampling/executor.cc
```diff
@@ -70,8 +71,9 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
                               std::is_same_v<T, PromoteDormantRotation>) {
                     num_expression_terms += typed.sign.terms().size();
                 } else if constexpr (std::is_same_v<T, MeasureActivePauli> ||
-                                     std::is_same_v<T, MeasureDormantRandom> ||
-                                     std::is_same_v<T, RecordClassical>) {
+                                     std::is_same_v<T, MeasureDormantRandom>) {
+                    num_expression_terms += typed.outcome.terms().size() - 1;
```
bachase, line 80: Why is this one less? Was it wrong before?

src/clifft/sampling/executor.h
bachase: Added a struct-level comment defining reachability and conditional joint log probability.

src/clifft/sampling/executor.h
bachase: Reworded this around the raw measurement branch and known sign flips from earlier stochastic events. That keeps the suggested intuition while also covering corrections from prior measurements.

src/clifft/sampling/executor.h
bachase: Expanded the API comment: replay reconstructs the corresponding branch state and joint log probability, enabling differential validation and exact record-probability queries. It also documents record ordering, conditioning, and RNG behavior.

src/clifft/sampling/executor.cc
bachase: It was not wrong before: the executor previously stored the complete outcome. Replay now stores the current branch separately and keeps only the remaining correction terms, so the reserved term count is one smaller. I added a comment at the subtraction.

src/clifft/sampling/executor.h
bachase: Good catch. I removed the duplicate private result type. I did not reuse ReplayResult because the helper returns a single-branch log-probability increment, not the cumulative replay result; it now returns optional<double>, with nullopt meaning unreachable.

PR #263, opened 2026-08-06
feat(sampling): expose experimental Python backend

### LINKED ISSUE #262: Expose the scalar sampling backend through an experimental Python API
## Context

Parent tracker: #247
Architecture discussion: #236
Depends on: #261

The scalar SamplingPlan executor and forced-record replay now cover the initial noiseless rotation-and-measurement subset. The next increment should expose that path explicitly in Python so it can reuse Clifft’s existing curated correctness corpus before more backend capabilities are added.

## Scope

- Add an explicit public experimental namespace without changing the default backend.
- Add an opaque experimental program type that owns the prepared scalar sampling plan instead of attaching a second backend payload to the legacy bytecode Program.
- Compile Stim text through parse, trace, and the normal HIR optimization pipeline into the experimental program.
- Add batch sample execution that prepares once, preallocates output, and reuses one executor across shots.
- Add experimental record-probability queries backed by forced replay.
- Return the existing Python SampleResult shape for supported circuits.
- Reject unsupported HIR operations clearly at experimental compile time.

## Validation

Reuse and parameterize the applicable existing Python tests instead of creating a new random-circuit generator. Initial conformance should include:

- deterministic measurements, superposition, Bell correlation, reproducibility, result shape, and Y-basis interference;
- Bell, Clifford+T, arbitrary-rotation, unreachable, repeated-measurement, and normalization record probabilities;
- existing Qiskit-backed terminal-measurement cases, including the established dense-entanglement fixtures where practical.

Keep noise, feedback, resets, detectors/observables, instruments, postselection, and physical statevector reconstruction out of this increment. Existing legacy-only tests for those features remain unchanged.

## Deferred

- Changing clifft.compile or clifft.sample defaults.
- A long-term backend selector or common public program-metadata redesign.
- New randomized circuit-generation infrastructure.
- Noise, feedback, resets, detectors/observables, instruments, and get_statevector support.

bachase:
## Summary

- add an explicit opt-in `clifft.experimental` program, compiler, sampler, and record-probability API without changing the default backend
- prepare once and reuse one scalar executor across batch sampling and forced-record replay, with output allocated before hot execution
- reuse the existing Python sampling, Stim, exact-probability, HIR-pass, and Qiskit-backed conformance tests against both compatible backends
- keep the experimental namespace out of the generated API documentation for now and reject noise, feedback, resets, detectors, and other unsupported operations at compile time

## Stacking

This PR is stacked on #261 and intentionally targets `codex/sampling-forced-replay`. Retarget it to `main` after #261 merges.

Closes #262.

## Validation

- 1,194 native non-benchmark tests
- 55 focused native sampling tests
- 952 Python tests
- 223 focused Python sampling and HIR-pass tests
- strict documentation build

## Implementation note

Release builds assume finite arithmetic, so unreachable native replay results use the lowest finite `double`, matching the legacy convention. The Python wrapper translates that sentinel to `-inf` for log-probability output.

### DIFF
--- a/docs/reference/python-api.md
+++ b/docs/reference/python-api.md
@@ -24,6 +24,20 @@
 
 ::: clifft.sample_k_survivors
 
+## Symbolic Sampling Backend (experimental)
+
+This explicit opt-in surface exercises the new scalar symbolic-coordinate
+backend without changing the default `clifft.compile` and `clifft.sample`
+pipeline. It currently accepts noiseless rotations and measurements only.
+
+::: clifft.experimental.compile
+
+::: clifft.experimental.sample
+
+::: clifft.experimental.record_probabilities
+
+::: clifft.experimental.Program
+
 ## Leakage and Loss (experimental)
 
 Sampling under a five-level leakage/loss model. See the
--- a/docs/reference/python-api.md
+++ b/docs/reference/python-api.md
@@ -24,20 +24,6 @@
 
 ::: clifft.sample_k_survivors
 
-## Symbolic Sampling Backend (experimental)
-
-This explicit opt-in surface exercises the new scalar symbolic-coordinate
-backend without changing the default `clifft.compile` and `clifft.sample`
-pipeline. It currently accepts noiseless rotations and measurements only.
-
-::: clifft.experimental.compile
-
-::: clifft.experimental.sample
-
-::: clifft.experimental.record_probabilities
-
-::: clifft.experimental.Program
-
 ## Leakage and Loss (experimental)
 
 Sampling under a five-level leakage/loss model. See the
--- a/src/clifft/sampling/executor.h
+++ b/src/clifft/sampling/executor.h
@@ -185,14 +185,16 @@ class Executor {
 
 // Samples a fixed number of shots into row-major visible-record storage. The
 // plan and executor are prepared once, and all output is allocated before the
-// first shot enters hot execution.
+// first shot enters hot execution. Plans with presampled symbols are rejected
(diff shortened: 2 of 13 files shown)


docs/reference/python-api.md
```diff
@@ -24,6 +24,20 @@
 
 ::: clifft.sample_k_survivors
 
+## Symbolic Sampling Backend (experimental)
```
bachase: Let's actually leave this undocumented for now. We can document once we are really ready to publicize and commit to it.

src/clifft/sampling/executor.h
```diff
@@ -133,6 +135,7 @@ class ExecutablePlan {
 class Executor {
   public:
     explicit Executor(const ExecutablePlan& plan, uint64_t seed = 0);
+    Executor(const ExecutablePlan& plan, std::nullopt_t);
```
bachase: What is this second ctor for? Its confusing with this and the optional seed explicit one above. Do we really need both/

docs/reference/python-api.md
bachase: Removed the generated API-reference section. The opt-in module keeps only its source-level instability disclaimer for now.

src/clifft/sampling/executor.h
bachase: Removed the nullopt constructor. Unseeded batch sampling now constructs the ordinary deterministic executor and calls a named reseed_from_entropy method, matching the existing SchrodingerState pattern.

PR #266, opened 2026-08-06
feat(sampling): complete noisy syndrome lifecycle

### LINKED ISSUE #265: Complete the scalar symbolic sampling lifecycle for noisy syndrome circuits
## Context

Parent tracker: #247
Architecture discussion: #236
Depends on: #263

The experimental scalar backend currently supports noiseless rotations and measurements. Before performance measurement, the next increment should complete the symbolic stochastic and classical-output lifecycle so representative Clifft and SymFT/SOFT benchmark circuits can execute without designing noise, readout, feedback, and syndrome handling in isolation.

### Planning

- Lower each HIR Pauli noise site into an explicit categorical distribution whose mutually exclusive outcomes define presampled Boolean symbols.
- Pull sampled Pauli errors into affine signs on later rotations and measurements instead of emitting runtime Pauli-frame work.
- Lower conditional Paulis by propagating the controlling measurement-record expression into later anticommuting operations. This must cover classical feedback and reset decompositions using visible or hidden measurements.
- Model symmetric and asymmetric readout noise at the record site. The sampled flip may depend on the pre-flip record value, and later feedback, detectors, and observables must consume the post-readout value.
- Lower detectors and observables to affine record parities with explicit expected-parity normalization.
- Mark postselected detectors at their circuit position so execution can reject a shot without runtime topology analysis.
- Reduce expressions through direct substitution and canonical XOR cancellation while constructing the plan; do not add a separate general symbolic optimizer in this increment.

### Scalar execution

- Prepare all noise distributions, expressions, detector/observable writes, and output storage before hot execution.
- Sample categorical Pauli noise and conditional readout flips with the repository deterministic RNG contract.
- Preserve mutually exclusive channel semantics, correlations through feedback and resets, and stable visible/hidden record numbering.
- Produce measurements, normalized detectors, and normalized observables.
- Support survivor/discard and logical-error accounting for postselected programs, including early shot rejection.
- Keep ordinary dispatch and kernels allocation-free and exception-free.

### Experimental Python API

- Accept the existing compile options for postselection masks, expected detector/observable parities, and automatic reference normalization.
- Return the existing SampleResult shape for fixed-shot programs.
- Add the minimum experimental survivor interface needed to exercise postselected fixtures.
- Continue rejecting noisy forced-record probability queries until marginalization over hidden stochastic symbols is designed.

## Validation

Reuse existing tests and fixtures rather than adding a random-circuit corpus:

- reset and measurement-reset behavior, including hidden-record indexing;
- measurement-controlled Pauli feedback before later quantum operations and measurements;
- one-, two-, and three-qubit Pauli channels plus correlated and else-correlated errors;
- symmetric and asymmetric readout noise, including feedback after readout;
- detector and observable parity, reference normalization, multiple observables, postselection, survivor counts, and logical-error accounting;
- independent Stim comparisons for stochastic Clifford behavior;
- legacy differential checks for supported Clifford+T behavior;
- small-shot smoke tests for the existing QEC fixture, cultivation d5 fixture, and a generated surface-code circuit.

The representative fixtures must compile and execute through the experimental backend before the subsequent planner/executor measurement increment begins.

## Non-goals

- Packed cross-shot evaluation or SIMD specialization.
- Performance tuning or a broad instrumentation framework.
- Product-component factorization.
- State-dependent instruments, leakage, or loss.
- EXP_VAL, statevector reconstruction, or making the experimental backend the default.
- Noisy record-probability marginalization.

bachase:
## Summary

- lower Pauli noise, correlated-error chains, feedback, and resets into symbolic affine dependencies while preserving mutually exclusive noise-site distributions
- execute symmetric and asymmetric readout noise, normalized detectors and observables, postselection, and survivor accounting without hot-path allocation
- extend the experimental Python backend and reuse the existing sampling, syndrome-normalization, postselection, and QEC fixture coverage
- keep state-dependent instruments and exact-state probes explicitly unsupported

Closes #265

## Stack

- based on #263, which exposes the experimental Python backend
- #263 is based on #261, which adds forced replay validation

## Validation

- cmake --build build -j4
- ctest --test-dir build -R Sampling --output-on-failure: 60 passed
- uv run pytest tests/python/test_experimental_sampling.py tests/python/test_sample.py -q: 140 passed

### DIFF
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -32,6 +32,14 @@ uint32_t index(InstrumentSiteId site) {
     return static_cast<uint32_t>(site);
 }
 
+uint32_t index(DetectorSlot slot) {
+    return static_cast<uint32_t>(slot);
+}
+
+uint32_t index(ObservableSlot slot) {
+    return static_cast<uint32_t>(slot);
+}
+
 }  // namespace
 
 MeasurementBranchClassification classify_measurement_branch(
@@ -59,6 +67,9 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
       max_active_width_(plan.max_active_width),
       num_visible_records_(plan.num_visible_records),
       num_hidden_records_(plan.num_hidden_records),
+      num_detectors_(plan.num_detectors),
+      num_observables_(plan.num_observables),
+      has_postselection_(plan.has_postselection),
       global_weight_(plan.global_weight) {
     plan.validate();
     if (plan.symbols.size() > std::numeric_limits<uint32_t>::max()) {
@@ -83,6 +94,11 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
                     num_expression_terms += typed.outcome.terms().size();
                 } else if constexpr (std::is_same_v<T, DefineSymbol>) {
                     num_expression_terms += typed.value.terms().size();
+                } else if constexpr (std::is_same_v<T, ApplyReadoutNoise>) {
+                    num_expression_terms += typed.source.terms().size();
+                } else if constexpr (std::is_same_v<T, WriteDetector> ||
+                                     std::is_same_v<T, WriteObservable>) {
+                    num_expression_terms += typed.outcome.terms().size();
                 } else if constexpr (std::is_same_v<T, InstrumentBoundary>) {
                     // Rejected below after all structural validation has run.
                 } else {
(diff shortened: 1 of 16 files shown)


src/clifft/sampling/plan.h
```diff
@@ -141,6 +144,31 @@ struct DefineSymbol {
...
+struct ApplyReadoutNoise {
+    SymbolId flip{};
+    AffineBool source;
+    RecordSlot record{};
+    double prob_zero_to_one = 0.0;
+    double prob_one_to_zero = 0.0;
+};
+
+// Writes a normalized detector parity. A nonzero postselected detector rejects
+// the shot immediately; later actions and outputs are irrelevant for it.
```
bachase: How do we handle the reference parity that this is defined against?

src/clifft/sampling/executor.cc
```diff
@@ -249,6 +329,27 @@ ReplayResult Executor::execute_actions(std::span<const uint8_t> forced_records)
                     }
                 } else if constexpr (std::is_same_v<T, ExecutablePlan::ExecuteSymbolDefinition>) {
                     symbols_[typed.symbol] = static_cast<uint8_t>(evaluate(typed.value));
+                } else if constexpr (std::is_same_v<T, ExecutablePlan::ExecuteReadoutNoise>) {
+                    if constexpr (ForceRecords) {
```
bachase: For this (and similar spots) long if constexpr type setup, where most of the variants are treated separately, is it better to have standalone functions to isolate the logic and make it easier to read? More similar to the per opcode dispatch that we have with the SVM approach?

src/clifft/sampling/executor.h
```diff
@@ -160,13 +205,17 @@ class Executor {
...
+    [[nodiscard]] std::span<const uint8_t> observables() const { return observables_; }
+    [[nodiscard]] bool discarded() const { return discarded_; }
     [[nodiscard]] const State& state() const { return state_; }
     // Counts positive branch probability mass classified as numerical dust.
     // This telemetry accumulates across shots.
     [[nodiscard]] uint64_t dust_clamps() const { return dust_clamps_; }
 
   private:
-    void initialize_shot(std::span<const uint8_t> presampled_values) noexcept;
+    void initialize_shot(std::span<const uint8_t> presampled_values, bool sample_noise) noexcept;
```
bachase: Should we document what sample_noise/presampled noise means in these functions?

bachase (commented):
Overall -- I'm also curious if we should have more of a bytecode type representation in this new pathway that is similar to the SVM approach. That made it easy to review/understand what the ultimate post HIR execution would be. What is the trade-off to somthing similar here? Would that also help (or hurt?) with different executions (CPU, CPU batch, GPU etc)?

Issue #269, opened 2026-08-07
Add experimental symbolic-coordinate instrument continuations


bachase:
## Goal

Complete the next CPU-first feature-parity slice by executing state-dependent transition instruments and resumable continuations through the symbolic-coordinate sampling backend, while leaving the legacy SVM as the default production path.

## Scope

- Lower HIR instrument source observables and destination-frame corrections into `SamplingPlan` without runtime tableau work.
- Support definite-source, active/mixed-coordinate, exact dormant activation, and dormant trap-form instrument strategies.
- Precompute direct-Pauli damping, population, and source-collapse descriptors before hot execution.
- Report stable-site traps and preserve coefficients, active-coordinate meaning, symbols, records, outputs, RNG position, and telemetry across a recompiled continuation.
- Permit coefficient, record, symbolic, and scratch storage to grow only at the continuation boundary.
- Expose an explicit experimental noncomputational sampling route without changing the existing public default.
- Reuse the existing trajectory semantics and tests for multiple traps, source-dependent damping, neglect, resets, detector/observable continuity, noise/RNG determinism, and hidden forced trace-out records.

## Constraints

- No runtime localization, commutation analysis, tableau evolution, or dependency discovery.
- No ordinary hot-path allocation or exceptions.
- Preserve stable visible/hidden record and instrument-site identities.
- Keep planner-performance, symbolic-sparsity, packed execution, SIMD, GPU, WASM, and default-backend switching out of scope.

## Acceptance

- Focused direct-Pauli instrument kernel tests.
- Planner and executor validation for every instrument strategy.
- End-to-end experimental trajectory tests differential against the legacy path where deterministic seeded parity applies, plus existing statistical/reference checks where RNG routing differs.
- Full formatting, C++ tests, and Python tests appropriate to the new experimental surface.

bachase:
Implementation is split into two reviewable increments:

1. #270 adds the internal planner, kernels, executor, and continuation contract. It includes stable symbol/noise prefixes, boundary-only storage growth, forced hidden trace-out records, and restoration of the root plan between shots.
2. A stacked follow-up will connect the existing noncomputational rewriter/trajectory lifecycle to this executor and expose the public experimental Python route described by this issue.

The first increment intentionally leaves fixed-plan batch sampling unavailable for instrument plans; this issue remains open until the public trajectory integration and differential coverage land.

bachase:
[...]

PR #270, opened 2026-08-07
feat(sampling): execute transition instruments

bachase:
Part of #269.

This adds the internal sampling-plan and executor support needed for transition instruments:

- plan classical, active, activating, and dormant-trap source handling without runtime topology work
- apply exact no-fire filtering and source collapse for direct active Paulis, while propagating computational destination changes symbolically
- stop on noncomputational outcomes and resume replacement plans at validated instrument boundaries
- preserve stable symbol prefixes and defer suffix noise sampling until the matching continuation is known
- allow coefficient and side storage to grow only at an explicit continuation boundary, including forced hidden trace-out records
- restore the original executable automatically before the next shot after a continuation

The existing fixed-plan batch entry points continue to reject instrument plans. Issue #269 remains open for the stacked public experimental trajectory driver that will connect the noncomputational rewriter to these continuation APIs.

Validation:

- `ctest --test-dir build --output-on-failure` (1,219 tests, including benchmarks)
- post-format `ctest --test-dir build -E '^Bench:' --output-on-failure -j 8` (1,213 tests)

### DIFF
--- a/src/clifft/sampling/plan.h
+++ b/src/clifft/sampling/plan.h
@@ -16,6 +16,7 @@
 
 #include "clifft/util/numeric.h"
 
+#include <array>
 #include <complex>
 #include <cstdint>
 #include <limits>
@@ -58,10 +59,12 @@ enum class ObservableSlot : uint32_t {};
 }
 
 enum class SymbolKind : uint8_t {
+    Unused,      // Stable operation-order slot that this compiled plan does not need.
     Presampled,  // Available before the action stream, such as sampled noise.
     Derived,     // Computed as a parity of previously available symbols.
     Branch,      // Sampled while applying a measurement action.
     Readout,     // Sampled from a record-dependent readout channel.
+    Instrument,  // Records an in-line computational instrument destination flip.
 };
 
 // A parity expression over plan symbols, optionally XORed with true. For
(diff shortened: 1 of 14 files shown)


src/clifft/sampling/plan.h
```diff
@@ -215,15 +243,15 @@ struct PlannedAction {
 // SamplingPlan::validate enforces the legal field combinations and verifies
 // that definition metadata agrees with the referenced action.
 struct SymbolInfo {
-    SymbolKind kind = SymbolKind::Presampled;
+    SymbolKind kind = SymbolKind::Unused;
 
-    // This is nullopt for Presampled. For Branch and Derived it identifies the
-    // unique action that assigns the symbol.
+    // For action-defined kinds, this identifies the unique assigning action.
```
bachase: This comment is confusing, is there a clearer phrasing?

src/clifft/sampling/plan.h
```diff
@@ -192,19 +195,44 @@ struct WriteObservable {
...
+    Classical,    // The source is the Boolean sign; no coefficient work is needed.
+    Active,       // The source Pauli is already supported on active coordinates.
+    Activate,     // Add one |0> coordinate before applying the source Pauli.
+    DormantTrap,  // Sample a dormant-random source and let a continuation collapse it.
+};
+
+struct ApplyInstrument {
+    InstrumentSiteId site{};
+    InstrumentMode mode = InstrumentMode::Classical;
+    // Identity for Classical and DormantTrap. Activate uses the post-activation
```
bachase: To make it read clearer, how about "source is Identity ..."

src/clifft/sampling/planner.cc
```diff
@@ -59,6 +60,18 @@ struct PendingReadoutNoise {
     RecordSlot record{};
     double prob_zero_to_one = 0.0;
     double prob_one_to_zero = 0.0;
+    SymbolId flip{};
+};
+
+struct PendingInstrument {
```
bachase, line 66: For this (and potentially other "Pending" structs", should we add some comments or more detail on what each member correspond to? Some are obvious (say PendingReadoutNoise), but some might be less clear

src/clifft/sampling/planner.cc
```diff
@@ -600,6 +683,71 @@ AffineBool process_measurement(std::vector<PendingOperation>& pending, size_t in
     return outcome;
 }
 
+void process_instrument(std::vector<PendingOperation>& pending, size_t operation_index,
```
bachase, line 693: This function is a bit dense, is it worth some concise comments throughout to explain what is happening?

src/clifft/sampling/state.cc
```diff
@@ -91,6 +91,41 @@ void State::reset() noexcept {
...
+    const uint64_t new_capacity = uint64_t{1} << max_active_width;
+    const uint64_t new_coefficient_stride = round_array_stride(new_capacity);
+    const uint64_t new_scratch_capacity = std::max(uint64_t{1}, new_capacity >> 1);
+    const uint64_t new_scratch_stride = round_array_stride(new_scratch_capacity);
+    const uint64_t allocated_doubles = 2 * new_coefficient_stride + 2 * new_scratch_stride;
+    if (allocated_doubles > std::numeric_limits<size_t>::max() / sizeof(double)) {
+        throw std::length_error("sampling state allocation exceeds addressable memory");
+    }
+
+    PageAlignedAllocation allocation(static_cast<size_t>(allocated_doubles) * sizeof(double));
```
bachase, line 115: I'm still not of fan of having to double allocate transiently, but don't see away around it, right?

src/clifft/sampling/planner.cc
bachase: Added focused documentation for the non-obvious PendingInstrument fields and renamed flip to destination_flip_symbol. The comments now identify the coordinate-dependent Paulis, physical G/E sign mapping, stable site identity, reserved symbol, and continuation prefixes.

src/clifft/sampling/planner.cc
bachase: Added concise comments at the semantic decision points: when dormant exact damping requires activation, why all related Paulis and the suffix share a coordinate change, how Classical versus Active is selected, why the flip symbol exists only in-line, and why the boundary follows the action.

src/clifft/sampling/plan.h
bachase: Rephrased this as the exact per-kind contract: the assigning action index is required for Derived, Branch, Readout, and Instrument symbols, and absent for Presampled and Unused.

src/clifft/sampling/plan.h
bachase: Updated to say explicitly that the source is identity for Classical and DormantTrap.

src/clifft/sampling/state.cc
bachase: Right. Preserving a live contiguous aligned state requires the old block until the coefficients have been copied into the new one. Portable aligned allocation cannot grow in place; avoiding the transient old-plus-new peak would require reserving maximum capacity up front or a platform-specific virtual-memory design. I added a comment recording that tradeoff.

PR #273, opened 2026-08-07
feat(sampling): add expectation value probes

### LINKED ISSUE #272: Add EXP_VAL probes to the symbolic-coordinate sampler
## Goal

Add `EXP_VAL` probes to the experimental symbolic-coordinate CPU backend so ordinary sampling exposes the same per-shot expectation-value columns as the production SVM.

This is the next capability-parity increment under #247 and discussion #236.

## Design

- Carry each HIR expectation probe through planner coordinate rewrites and symbolic Pauli-frame sign propagation.
- Emit a read-only semantic plan action with a stable output slot.
- Classify dormant X/Y support in the planner; such probes write zero without a dense coefficient pass.
- Prepare active Pauli descriptors once in `ExecutablePlan` and evaluate them allocation-free and exception-free during dispatch.
- Preallocate expectation-value output storage in `Executor`.
- Return `exp_vals` from experimental fixed-row and survivor sampling with the existing Python `SampleResult` shape and semantics.

## Validation

- Reuse the existing EXP_VAL oracle cases against both Python backends, including Clifford, non-Clifford, entangled, dormant, signed, measurement-dependent, and multi-probe cases.
- Add focused semantic-plan, kernel, executor, and survivor-output tests.
- Preserve record and state evolution: probes are read-only and do not consume RNG.

## Scope

`record_probabilities()` already supports the experimental backend. `basis_probabilities()` and `get_statevector()` remain a separate exact-state increment: unlike an in-stream EXP_VAL probe, they require retaining and composing the final physical Clifford/coordinate frame and sharing or adapting the batch exact-query machinery. Keeping that work separate avoids adding unused final-frame metadata to this PR before its contract is designed.

Closes when EXP_VAL parity is merged; exact-state API parity remains an explicit #247 replacement gate.

bachase:
## Summary

- lower `EXP_VAL` probes through planner coordinate rewrites and symbolic Pauli-frame sign propagation
- evaluate prepared active-coordinate Paulis without state mutation, RNG consumption, allocation, or exceptions in hot dispatch
- classify dormant X/Y probes and active identity probes so they require no dense coefficient pass
- return preallocated `exp_vals` columns from experimental fixed-row and survivor sampling
- run the existing NumPy, MPP, noise, feedback, and non-disturbance oracle suite against both backends

## Scope

This completes in-stream expectation-probe parity. `record_probabilities()` already supports the experimental backend. Final-state `basis_probabilities()` and `get_statevector()` remain the next exact-query increment because they require a retained and composed physical-coordinate frame rather than the circuit-position Pauli used by `EXP_VAL`.

## Validation

- C++ non-benchmark suite: 1,222 cases / 375,657 assertions
- focused C++ sampling suite: 82 cases / 166,768 assertions
- Python suite: 1,105 tests
- focused EXP_VAL and experimental API suite, including a 129-qubit dormant-mask regression
- continuation persistence and forced-replay expectation regressions

Closes #272

### DIFF
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -51,6 +51,7 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
       num_hidden_records_(plan.num_hidden_records),
       num_detectors_(plan.num_detectors),
       num_observables_(plan.num_observables),
+      num_exp_vals_(plan.num_exp_vals),
       has_postselection_(plan.has_postselection),
       global_weight_(plan.global_weight),
       instrument_distributions_(plan.instrument_distributions) {
@@ -82,6 +83,8 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
                 } else if constexpr (std::is_same_v<T, WriteDetector> ||
                                      std::is_same_v<T, WriteObservable>) {
                     num_expression_terms += typed.outcome.terms().size();
+                } else if constexpr (std::is_same_v<T, WriteExpectationValue>) {
+                    num_expression_terms += typed.sign.terms().size();
                 } else if constexpr (std::is_same_v<T, ApplyInstrument>) {
                     num_expression_terms += typed.sign.terms().size();
                 } else if constexpr (std::is_same_v<T, InstrumentBoundary>) {
@@ -175,6 +178,13 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
                 } else if constexpr (std::is_same_v<T, WriteObservable>) {
                     actions_.emplace_back(ExecuteObservable{prepare_expression(typed.outcome),
                                                             index(typed.observable)});
+                } else if constexpr (std::is_same_v<T, WriteExpectationValue>) {
+                    std::optional<PreparedPauli> pauli;
+                    if (typed.pauli.has_value()) {
+                        pauli = prepare_pauli(*typed.pauli, planned.active_before);
+                    }
+                    actions_.emplace_back(ExecuteExpectation{
+                        std::move(pauli), prepare_expression(typed.sign), index(typed.exp_val)});
                 } else if constexpr (std::is_same_v<T, ApplyInstrument>) {
                     has_instruments_ = true;
                     std::optional<PreparedMeasurement> measurement;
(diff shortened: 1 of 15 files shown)


src/clifft/sampling/plan.h
```diff
@@ -195,6 +199,15 @@ struct WriteObservable {
     ObservableSlot observable{};
 };
 
+// Writes a non-destructive Pauli expectation probe. An absent active Pauli
```
bachase: As absent pauli the best way to represent this case? That seems to overload semantics, where there is an ActivePauli to represent the EXP_VAL observable, but the operator just has dormant support?

src/clifft/sampling/executor.cc
```diff
@@ -536,6 +548,18 @@ void Executor::execute_action(const ExecutablePlan::ExecuteObservable& action,
     observables_[action.observable] = static_cast<uint8_t>(evaluate(action.outcome));
 }
 
+template <bool ForceRecords>
+void Executor::execute_action(const ExecutablePlan::ExecuteExpectation& action,
+                              std::span<const uint8_t>, ReplayResult&) noexcept {
+    assert(action.exp_val < exp_vals_.size() && "expectation slot must be preallocated");
+    if (!action.pauli.has_value()) {
+        exp_vals_[action.exp_val] = 0.0;
```
bachase, line 561: If we pre-allocate exp_vals to be all 0, do we have ot do this at runtime? Maybe not that useful optimization

bachase: The store is required because outputs are not cleared per shot: validation guarantees each completed plan overwrites them. In particular, an instrument continuation can replace a suffix where this slot was nonzero with one where dormant support makes it exact zero. Construction-time initialization would then leak the prior suffix result. a6daf85a documents this and adds a regression that changes the same slot from 1.0 to zero across continuations.

src/clifft/sampling/plan.h
bachase: The full transformed observable has already been classified at this layer; the optional value is the active-coordinate kernel input, with absence representing the distinct exact-zero result. I kept the two-case optional rather than adding another variant or a mode plus a dummy Pauli, but renamed the field to `active_projection` and clarified that contract in a6daf85a so it no longer reads as a missing observable.

PR #275, opened 2026-08-07
feat(sampling): add basis probability queries

### LINKED ISSUE #274: Add basis probability queries to the symbolic-coordinate backend
## Goal

Add exact computational-basis probability queries to the experimental symbolic-coordinate CPU backend without adding work to ordinary shot execution.

This is the next exact-query replacement gate under #247 and discussion #236.

## Design

- Retain an immutable final coordinate-to-physical Clifford frame only for plans whose HIR is eligible for pure-state basis queries.
- Compose that frame during planning, where coordinate changes are already known; runtime dispatch must not evolve or inspect tableaus.
- Reuse the executor existing read-only coefficient-state accessor without copying its split arrays.
- Refactor the existing selected-basis amplitude evaluator so legacy interleaved coefficients and symbolic split real/imaginary coefficients share the same elimination and Gray-code implementation.
- Execute the unitary plan once when the exact query is called, then evaluate only the requested physical basis masks. Do not expand a full statevector.
- Expose `clifft.experimental.basis_probabilities()` with the existing input, bit-order, dtype, shape, and return-log semantics. The production default remains unchanged.

## Hot-path acceptance gates

- No new semantic or executable actions.
- No new ordinary dispatch branches, dense passes, RNG draws, allocations, or per-shot frame updates.
- No final-frame metadata or composition work for noisy, measured, feedback, detector/observable, postselected, or instrument plans.
- Existing action counts and seeded sampling behavior remain unchanged.

## Validation

- Reuse the existing Python basis-query suite against both backends where the public contracts overlap.
- Compare representative Clifford, Clifford+T, continuous-rotation, rank-deficient, full-rank, and multiword-mask cases with the production backend and Qiskit/statevector oracles.
- Add focused planner/executor tests for final-frame composition and nonunitary rejection.

## Scope

`get_statevector()` remains the following increment because its public API exposes an execution-state object; it will reuse this final-frame handoff after the backend-neutral state contract is designed. Importance sampling and performance work remain later #247 gates.

Closes when basis-query parity is merged.

bachase:
Closes #274.

## Summary

- retain the final coordinate-to-physical tableau only for pure-state plans eligible for exact basis queries
- share the existing exact basis-probability evaluator between the legacy interleaved state and symbolic split coefficients
- expose clifft.experimental.basis_probabilities with the same query inputs and results as the public API
- run the existing basis-probability Python oracle suite against both backends, plus focused planner, validation, multiword-mask, and executor differentials

## Sampling-path impact

This adds no sampling actions and changes no ordinary executor dispatch or kernel. The final tableau is composed during planning and retained only for eligible unitary plans; noisy, measured, feedback, detector, and instrument plans do not carry it. Executor state is evaluated only when the exact query is explicitly requested.

Dense statevector expansion remains deferred to a separate slice.

## Validation

- pre-commit: all hooks passed
- C++: 1,225 cases / 375,688 assertions passed
- Python: 1,145 tests passed
- focused basis-probability Python suite: 80 tests passed

### DIFF
--- a/src/clifft/sampling/executor.h
+++ b/src/clifft/sampling/executor.h
@@ -52,6 +52,13 @@ struct ForcedTraceOut {
     uint8_t source = 0;
 };
 
+class ExecutablePlan;
+
+[[nodiscard]] std::vector<double> basis_probabilities(const ExecutablePlan& plan,
+                                                      std::span<const uint64_t> basis_masks,
+                                                      size_t num_basis_masks,
+                                                      size_t words_per_basis_mask);
+
 [[nodiscard]] MeasurementBranchClassification classify_measurement_branch(
     MeasurementProbabilities probabilities) noexcept;
 
@@ -71,6 +78,7 @@ class ExecutablePlan {
     [[nodiscard]] bool has_postselection() const { return has_postselection_; }
     [[nodiscard]] bool has_readout_noise() const { return has_readout_noise_; }
     [[nodiscard]] bool has_instruments() const { return has_instruments_; }
+    [[nodiscard]] bool supports_basis_probabilities() const { return final_tableau_.has_value(); }
     [[nodiscard]] uint32_t num_instrument_sites() const {
         return static_cast<uint32_t>(instrument_distributions_.size());
     }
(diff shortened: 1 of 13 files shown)


src/clifft/sampling/executor.h
```diff
@@ -85,6 +93,10 @@ class ExecutablePlan {
 
   private:
     friend class Executor;
+    friend std::vector<double> basis_probabilities(const ExecutablePlan& plan,
```
bachase: I didn't notice this friends before. I usually think friend is a bad design sign. Can you check this and the existing use of friends in the sampling area to see which are really neecessary?

PR #281, opened 2026-08-07
perf: evaluate symbolic expressions incrementally

bachase:
## Summary

- lower every prepared affine expression to a Boolean register plus a symbol-to-register dependency tape
- update each dependent register only when a symbol is assigned true, making expression reads O(1)
- reconstruct continuation registers from constants and the live true-symbol prefix at the trap boundary
- remove the retained per-expression term lists from `ExecutablePlan`

This is the focused scalar execution optimization selected by the measurements in #280. It does not add packed-shot execution, prototype modes, public instrumentation, or a general graph-optimization framework.

## Release acceptance

Three balanced paired blocks were run from fresh processes on one pinned CPU, using matched circuit/output semantics and paired seeds. `symbolic/legacy` below 1 means symbolic is faster. The baseline is the pre-change run already posted on #280.

| Case | Baseline symbolic/legacy | This PR symbolic/legacy | This PR legacy ms | This PR symbolic ms |
|---|---:|---:|---:|---:|
| surface d7 r7 aggregate | 10.683x | 0.669x | 612.6 | 412.0 |
| surface d7 r7 forced k0 | not previously run | 0.468x | 1526.0 | 717.9 |
| cultivation d3 aggregate | 10.118x | 0.897x | 650.0 | 582.9 |
| cultivation d3 forced k0 | 12.541x | 0.972x | 547.0 | 531.8 |
| distillation aggregate | 0.454x | 0.176x | 2127.8 | 374.5 |
| coherent d3 r3 aggregate | 1.592x | 1.640x | 833.2 | 1389.6 |
| regime k12/L512 | 1.268x | 1.129x | 936.0 | 1057.9 |
| low-leak noncomputational | 9.850x | 9.687x | 405.3 | 3956.5 |

The surface A/A paired B/A control had median 0.998 (range 0.947-1.028); k12/L512 had median 1.007 (range 0.968-1.014). Discard and conditional logical statistics agree within their reported uncertainty. Checksums were used only to prove that outputs were consumed; cross-backend checksum equality was not required because RNG schedules differ.

The result closes the 10-12x surface/cultivation expression-evaluation gap without packed counts. It intentionally leaves the independently measured coherent-kernel gap, noncomputational continuation/suffix-rewrite gap, and compile-time gap for separate work.

## Profile

A Release-optimized, symbolized, frame-pointer build profiled surface d7 r7 after the change. The old per-use affine term scan is absent. Flat cycle shares for the replacement path were:

- direct expression-register lookup: 6.74%
- true-symbol dependency propagation: 4.74%
- expression-register reset: 0.95%
- symbol assignment helper: 0.89%

The larger remaining self-costs are normal shot dispatch/action execution and presampled-noise handling, rather than affine term traversal.

## Correctness

Added focused coverage for:

- resetting true-symbol effects across shots
- noisy postselection and normalized outputs
- continuation reconstruction from true prefix symbols
- the experimental noncomputational Python path across a trap

Existing suites cover forced-k sampling, replay, EXP_VAL, detector/observable normalization, continuation register growth, and the legacy/analytic oracles.

Validation:

- `ctest --test-dir build-production --output-on-failure` - 1,243 passed
- `uv run pytest tests/python/ -q` - 1,253 passed
- `uv run pre-commit run --all-files` - passed

Refs #280

### DIFF
--- a/src/clifft/sampling/executor.cc
+++ b/src/clifft/sampling/executor.cc
@@ -100,7 +100,42 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
     if (num_expression_terms > std::numeric_limits<uint32_t>::max()) {
         throw std::length_error("sampling executable expression storage exceeds uint32 range");
     }
-    expression_terms_.reserve(num_expression_terms);
+    std::vector<uint32_t> expression_terms;
+    expression_terms.reserve(num_expression_terms);
+    std::vector<uint32_t> expression_term_begins;
+    expression_term_begins.reserve(plan.actions.size());
+    expression_register_constants_.reserve(plan.actions.size());
+
+    auto prepare_expression = [&](const AffineBool& expression) {
+        if (expression_register_constants_.size() >= std::numeric_limits<uint32_t>::max()) {
+            throw std::length_error("sampling executable expression count exceeds uint32 range");
+        }
+        const uint32_t register_id = static_cast<uint32_t>(expression_register_constants_.size());
+        expression_term_begins.push_back(static_cast<uint32_t>(expression_terms.size()));
+        expression_register_constants_.push_back(static_cast<uint8_t>(expression.constant()));
+        for (SymbolId term : expression.terms()) {
+            expression_terms.push_back(index(term));
+        }
+        return PreparedExpression{register_id};
+    };
+    auto prepare_measurement_correction = [&](const AffineBool& outcome, uint32_t branch) {
+        if (expression_register_constants_.size() >= std::numeric_limits<uint32_t>::max()) {
+            throw std::length_error("sampling executable expression count exceeds uint32 range");
+        }
+        const uint32_t register_id = static_cast<uint32_t>(expression_register_constants_.size());
+        const uint32_t begin = static_cast<uint32_t>(expression_terms.size());
+        expression_term_begins.push_back(begin);
+        expression_register_constants_.push_back(static_cast<uint8_t>(outcome.constant()));
+        for (SymbolId term : outcome.terms()) {
+            if (index(term) != branch) {
+                expression_terms.push_back(index(term));
+            }
+        }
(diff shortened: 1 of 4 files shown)


src/clifft/sampling/executor.h
```diff
@@ -107,9 +107,7 @@ class ExecutablePlan {
     friend class Executor;
 
     struct PreparedExpression {
-        uint32_t term_begin = 0;
-        uint32_t term_count = 0;
-        bool constant = false;
+        uint32_t register_id = 0;
```
bachase, line 110: If this is the only member of PreparedExpression, is it worth having the struct still?

src/clifft/sampling/executor.h
```diff
@@ -227,7 +222,9 @@ class ExecutablePlan {
     uint32_t initial_noise_end_ = 0;
     std::complex<double> global_weight_ = {1.0, 0.0};
     std::optional<stim::Tableau<kStimWidth>> final_tableau_;
-    std::vector<uint32_t> expression_terms_;
+    std::vector<uint8_t> expression_register_constants_;
```
bachase, line 228: Should their be a brief block comment here xplaining what these 3 vectors are and how they relate?

src/clifft/sampling/executor.cc
```diff
@@ -100,7 +100,42 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
     if (num_expression_terms > std::numeric_limits<uint32_t>::max()) {
         throw std::length_error("sampling executable expression storage exceeds uint32 range");
     }
```
bachase, line 102: In either ExecutablePlan constructor, is there common code to extract for how these expression terms are created?

That is, I wonder if there is like an ExpressionTerms manager class that isloates the logic and makes it easier to understand and manage.

bachase: Regardless of that decisions, should there be a few concise inline comments sprinkled throughout on how this initialization is happening?

PR #282, opened 2026-08-08
perf: specialize symbolic Pauli rotations

bachase:
## Summary

- enumerate non-diagonal Pauli coefficient pairs directly instead of scanning and skipping half of the state
- precompute the one-bit phase relationship between paired coefficients
- specialize the scalar butterfly into explicit real and imaginary arithmetic for real and imaginary Pauli phases
- keep diagonal rotation, fusion, SIMD, measurement, and packed-shot behavior unchanged

Part of #280.

## Measurement

The descriptor-driven microbenchmark used the actual non-diagonal rotation streams from coherent d3 r3 and k12/L512. Individual speedups versus the current scalar kernel were:

| Variant | coherent d3 r3 | k12/L512 |
|---|---:|---:|
| direct pair enumeration | 1.06x | 1.07x |
| derive partner phase | 1.01x | 0.99x |
| explicit arithmetic | 1.19x | 1.29x |
| all three combined | 1.63x | 1.71x |

Three balanced blocks with six fresh processes per backend gave:

| Case | main symbolic/legacy | PR symbolic/legacy | PR legacy ms | PR symbolic ms |
|---|---:|---:|---:|---:|
| coherent d3 r3 aggregate | 1.640x | 1.189x | 824.2 | 970.2 |
| regime k12/L512 | 1.129x | 0.776x | 918.7 | 713.0 |
| surface d7 r7 aggregate | 0.669x | 0.774x | 542.5 | 414.8 |
| cultivation d3 aggregate | 0.897x | 0.860x | 609.2 | 526.1 |
| distillation aggregate | 0.176x | 0.195x | 1945.4 | 382.6 |

The surface and distillation symbolic absolute times remain stable; ratio movement there is dominated by legacy timing variation. Outcome statistics agree within uncertainty.

QV-10 raw was not in the earlier acceptance subset, so it received a clean main A/B. Symbolic time fell from 6.07 seconds to 4.42 seconds per 10,000 shots, a 1.37x speedup. The remaining 8.64x legacy gap is therefore pre-existing, not a regression from this change.

On coherent d3 r3, versus the merged-main profile, symbolic cycles fell 29.1%, instructions 34.6%, and branches 57.3%. Rotation self share fell from 68.99% to 53.73%; its estimated absolute cycle cost fell about 1.81x. The confirming call graph had zero lost samples. Measurement probability is now 13.59%, executor work 10.17%, and collapse 6.55%.

## Validation

- exhaustive dense rotation oracle: 91,264 assertions
- C++: 1,244/1,244 passed
- Python: 1,253/1,253 passed
- repository pre-commit: passed

### DIFF
--- a/src/clifft/sampling/kernels.cc
+++ b/src/clifft/sampling/kernels.cc
@@ -56,7 +56,7 @@ void apply_nondiagonal_rotation(State& state, const PreparedRotation& rotation,
             const double right_imag = imag[right];
             const bool odd_phase = (std::popcount(left & rotation.pauli.z) & 1U) != 0;
             const double left_sine = odd_phase ? -even_left_sine : even_left_sine;
-            const double right_sine = rotation.pauli.partner_phase_negated ? -left_sine : left_sine;
+            const double right_sine = RealPhase ? left_sine : -left_sine;
 
             if constexpr (RealPhase) {
                 real[left] = rotation.cosine * left_real + right_sine * right_imag;
@@ -117,7 +117,6 @@ PreparedPauli prepare_pauli(ActivePauli pauli, uint32_t active_width) {
         {1.0, 0.0}, {0.0, 1.0}, {-1.0, 0.0}, {0.0, -1.0}};
     const uint32_t overlap = std::popcount(pauli.x & pauli.z);
     return PreparedPauli{.active_width = active_width,
-                         .partner_phase_negated = (overlap & 1U) != 0,
                          .x = pauli.x,
                          .z = pauli.z,
                          .pair_selector = std::bit_floor(pauli.x),
--- a/src/clifft/sampling/kernels.cc
+++ b/src/clifft/sampling/kernels.cc
@@ -34,6 +34,45 @@ std::complex<double> phase_at(const PreparedPauli& pauli, uint64_t basis) {
     return (std::popcount(basis & pauli.z) & 1U) != 0 ? -pauli.even_phase : pauli.even_phase;
 }
 
+template <bool RealPhase>
+void apply_nondiagonal_rotation(State& state, const PreparedRotation& rotation,
+                                double sine) noexcept {
+    double* const real = state.real_data();
+    double* const imag = state.imag_data();
+    const uint64_t size = state.size();
+    const uint64_t pair_stride = rotation.pauli.pair_selector;
+    const uint64_t pair_period = pair_stride << 1;
+    const double base_phase =
+        RealPhase ? rotation.pauli.even_phase.real() : rotation.pauli.even_phase.imag();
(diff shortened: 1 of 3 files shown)


bachase:
One small simplification is worth testing before merge.

`partner_phase_negated` is redundant with the existing real/imaginary kernel
specialization. `even_phase` is `i^popcount(x & z)`, so:

- `RealPhase == true` exactly when `popcount(x & z)` is even, and the partner
  phase is unchanged;
- `RealPhase == false` exactly when it is odd, and the partner phase is
  negated.

Inside the template, this can therefore be written as:

```cpp
const double right_sine = RealPhase ? left_sine : -left_sine;
```

That permits removing `PreparedPauli::partner_phase_negated`, its preparation,
and its focused metadata assertions. It removes a redundant representation
invariant and lets the choice disappear at template instantiation instead of
depending on an additional descriptor field in the pair loop.

The current code is correct, so I do not consider this a correctness blocker.
Please test the simplified form on the Linux box against the combined scalar
microbenchmark and at least coherent d3 r3 and k12/L512. The compiler may
already be loop-unswitching the invariant Boolean, in which case the result
could be neutral; retain the simpler form if performance is neutral or better,
and report back if it produces a measurable regression.

bachase:
I tested the proposed removal of `PreparedPauli::partner_phase_negated` before retaining it. The replacement derives the partner sign from the existing `RealPhase` template specialization:

```cpp
const double right_sine = RealPhase ? left_sine : -left_sine;
```

The focused kernel A/B was noisy: coherent d3 r3 was about 0.9% faster in the balanced repeat, while k12/L512 was about 3.4% slower. I therefore checked the actual sampling path and hardware counters before deciding.

Three balanced end-to-end blocks, each ordered old/new/new/old:

| Case | Old mean | Simplified mean | Change |
|---|---:|---:|---:|
| coherent d3 r3, 300k shots x2 | 2.0107 s | 2.0366 s | +1.3% |
| k12/L512, 20k shots x2 | 1.4807 s | 1.4507 s | -2.0% |

Five-run `perf stat` counters:

| Case | Cycles | Instructions | Branches | Branch misses |
|---|---:|---:|---:|---:|
| coherent d3 r3 | -0.2% | -0.2% | -1.6% | -1.5% |
| k12/L512 | +0.4% | -0.04% | +1.3% | -1.0% |

I consider that neutral overall, with a small coherent instruction/branch improvement and no material k12 regression. The simplification also removes redundant prepared state and the associated runtime load, so I retained it in cd43ecf.

Validation on the retained version:

- exhaustive dense rotation oracle: 91,264 assertions passed;
- C++: 1,244/1,244 passed;
- Python: 1,253/1,253 passed;
- all pre-commit hooks passed.

PR #283, opened 2026-08-08
perf: fuse constant symbolic rotations

bachase:
## Summary

- fuse consecutive constant-sign symbolic rotations whose X masks span at most two active directions
- precompose each run into a bounded selector-indexed U1, U2, or U4 matrix
- execute the group in one scalar coefficient sweep
- retain the existing direct-rotation path for dynamic signs, wider X spans, and large selector spaces

This is the ISA-neutral base of the QV optimization identified in #280. An AVX-512 specialization follows in #284; this PR intentionally contains no architecture-specific representation or dispatch.

## Evidence

Matched Release measurements from the issue #280 campaign:

| Case | Before fusion | Scalar fusion | Speedup |
|---|---:|---:|---:|
| QV-10 raw, 10k shots | 4.513 s | 1.523 s | 2.96x |
| QV-20 raw, 1 shot | 2.452 s | 0.815 s | 3.01x |

The transformation removes about 89% of projected QV-10 rotation coefficient visits and 94% of QV-20 visits. Surface, cultivation, distillation, coherent d3 r3, and k12/L512 guards remained neutral within the campaign noise threshold.

The review-driven physical extraction into `sampling/fused_rotation.{h,cc}` was also neutral: balanced medians changed by +1.7% on QV-10 and -3.9% on QV-20.

## Design constraints

- `PreparedFusedRotation` is ISA-neutral.
- Fusion policy, GF(2) run discovery, matrix composition, and scalar execution are isolated in `sampling/fused_rotation.{h,cc}`.
- The `ExecutablePlan` constructor only consumes a prepared run or lowers its actions individually.
- Matrix tables are bounded to at most five independent representative selectors.
- Fusion requires at least three rotations, matching the legacy fusion threshold.
- Dynamic-sign and signed-identity rotations remain barriers on the direct scalar path.
- No allocations occur in the execution loop.
- No general pass framework is introduced.

## Validation

- 1,243 non-benchmark C++ tests
- 1,253 Python tests
- coefficient-by-coefficient scalar oracles for rank-zero, rank-one, and rank-two groups
- focused dynamic-sign, signed-identity/global-phase, below-threshold, and rank-three fallback tests
- existing sampling and backend parity tests
- pre-commit hooks

Refs #280


src/clifft/sampling/executor.h
```diff
@@ -115,6 +116,10 @@ class ExecutablePlan {
         PreparedExpression sign;
     };
 
+    struct ExecuteFusedRotation {
+        uint32_t rotation = 0;
```
bachase: SHould we add a comment that this is an index into the fused_rotation vector?

src/clifft/sampling/fused_rotation.h
```diff
@@ -0,0 +1,41 @@
...
+#include <complex>
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <span>
+#include <vector>
+
+namespace clifft::sampling {
+
+// Applies one precomposed unitary to every orbit generated by at most two X
```
bachase: This is a bit jargony (orbit) -- is there a way to express this in more standard quantum mechanics notation?

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+#include <array>
+#include <bit>
+#include <cassert>
+#include <complex>
+
+namespace clifft::sampling {
+
+namespace {
+
+// A general matrix sweep costs more than one Pauli butterfly, so match the
```
bachase: Have this be a more durable comment. We plan to retire the legacy pathway (fusion), so this comment should stand alone.

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+// legacy fusion threshold and require it to replace at least three sweeps.
+constexpr size_t kMinFusedRotationActions = 3;
+
+// Each independent representative parity doubles the matrix table. Five
```
bachase: Make this a little less jargony (representative parity?) 

Can reference QV workloads, but this should also stand alone.

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+// covers the measured QV workloads while bounding a descriptor at 32 matrices.
+constexpr uint32_t kMaxFusedRotationSelectors = 5;
+
+class BinaryBasis {
```
bachase, line 28: Add a brief class summary of what this class is representing mathematically.

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+  public:
+    [[nodiscard]] bool insert(uint64_t value, uint32_t max_rank = 64) {
```
bachase, line 33: Add brief documentation for public members.

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+            for (size_t inner = 0; inner < dimension; ++inner) {
+                result[row * dimension + column] +=
+                    left[row * dimension + inner] * matrix[inner * dimension + column];
+            }
+        }
+    }
+    std::copy_n(result.begin(), dimension * dimension, matrix.begin());
+}
+
+std::optional<PreparedFusedRotation> prepare_fused_rotation(std::span<const PlannedAction> actions,
```
bachase, line 110: Have concise comments sprinkled in at key parts of this function so readers know what it is doing.

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+            }
+            multiply_matrix_left(matrix,
+                                 std::span<const std::complex<double>>(unitary).first(matrix_size),
+                                 dimension);
+        }
+    }
+    return result;
+}
+
+uint64_t insert_zero_bit(uint64_t packed, uint32_t pivot) {
```
bachase: Is this function used elsewhere too? Wondering if there are some existing shared bitwise operators to use.

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+    const uint64_t lower_mask = (uint64_t{1} << pivot) - 1;
+    return (packed & lower_mask) | ((packed & ~lower_mask) << 1);
+}
+
+template <size_t Dimension>
+void apply_fused_rotation_orbits(State& state, const PreparedFusedRotation& rotation) noexcept {
```
bachase, line 203: Document this function (what it is doing mathematically)

src/clifft/sampling/fused_rotation.h
bachase: You can use orbit but just define it

src/clifft/sampling/fused_rotation.cc
```diff
@@ -0,0 +1,301 @@
...
+            }
+            real[indices[row]] = output_real;
+            imag[indices[row]] = output_imag;
+        }
+    }
+}
+
+}  // namespace
+
+FusedRotationRun prepare_fused_rotation_run(std::span<const PlannedAction> actions) {
```
bachase, line 273: Add some concise comments within this function too

tests/test_fused_rotation.cc
bachase, line 1: IS this testing sufficient?

src/clifft/sampling/executor.h
bachase: Renamed the field to rotation_index, so its role as an index into the prepared fused-rotation storage is explicit at every use.

src/clifft/sampling/fused_rotation.h
bachase: Defined orbit in the public descriptor comment as the set of computational-basis states obtained by XORing a basis state with the binary span of the X masks.

src/clifft/sampling/fused_rotation.cc
bachase: Replaced the legacy comparison with a standalone cost rationale: three rotations are required to amortize the extra arithmetic of one dense matrix sweep.

src/clifft/sampling/fused_rotation.cc
bachase: Reworded this in terms of Z-parity conditions evaluated on an orbit representative, and explained both the exponential table growth and the 32-matrix cap.

src/clifft/sampling/fused_rotation.cc
bachase: Added a class summary defining BinaryBasis as a row-reduced basis for a vector subspace of GF(2)^64.

src/clifft/sampling/fused_rotation.cc
bachase: Documented insert, rank, and rows, including the dependent-vector success behavior and the rank-cap failure condition.

src/clifft/sampling/fused_rotation.cc
bachase: Added concise mathematical comments for representative construction, selector reduction, matrix variants, and the split between representative and within-orbit phase.

src/clifft/sampling/fused_rotation.cc
bachase: There was a duplicate helper in the symbolic measurement kernels and a separate legacy SVM helper. I extracted a sampling-local indexing.h shared by both symbolic kernels, without introducing a dependency on the legacy SVM.

src/clifft/sampling/fused_rotation.cc
bachase: Documented the kernel as gathering each one-, two-, or four-dimensional coefficient subspace, applying its selected dense unitary, and scattering the result back. The representative and orbit enumeration steps are also annotated.

src/clifft/sampling/fused_rotation.cc
bachase: Added comments explaining the fixed-unitary eligibility rules, the maximal rank-two prefix scan, and why action_count remains useful when selector growth rejects fusion.

tests/test_fused_rotation.cc
bachase: The existing coefficient oracle covers rank-zero, rank-one, rank-two, high-pivot, and wide-selector fallback behavior, while the focused file covers dynamic-sign, identity, threshold, and rank-three barriers. I added the missing exact five-selector-cap case; the oracle now performs 299 assertions on the stacked AVX branch and 169 on the scalar PR.

PR #284, opened 2026-08-08
perf: vectorize fused symbolic rotations

bachase:
## Summary

- add an AVX-512 specialization for rank-two fused symbolic rotations with `pivot_lo >= 3`
- keep scalar fused execution as the fallback and correctness oracle
- prepare optional eight-lane weight sidecars only when the process-wide runtime selector chooses AVX-512
- move CPU feature detection and `CLIFFT_FORCE_ISA` handling into a shared internal utility used by both the legacy SVM and symbolic executor
- exercise symbolic sampling before legacy execution in wheel/QEMU smoke tests so fallback paths cannot accidentally rely on SVM initialization

This is stacked on #283. The scalar descriptor remains ISA-neutral; SIMD types and intrinsics stay in the separately compiled AVX-512 translation unit.

## Measured result

Single-threaded Release-with-symbols profiles on the issue #280 QV controls:

| Case | Scalar fusion | AVX-512 | Speedup | Legacy | Symbolic / legacy |
| --- | ---: | ---: | ---: | ---: | ---: |
| QV-10 | 1.523 s | 0.663 s | 2.30x | 0.499 s | 1.33x |
| QV-20 | 0.815 s | 0.183 s | 4.46x | 0.249 s | 0.73x |

The high-pivot specialization covers about 88% of fused QV-10 coefficient visits and effectively all QV-20 fused work. Post-change profiles attribute 32.2% of QV-10 runtime and 85.2% of QV-20 runtime to the packed U4 kernel. QV-10 remains behind legacy because 33.7% of runtime is unfused direct rotation and 22.8% is low-pivot/ineligible scalar fusion; those shapes are intentionally deferred.

Prepared AVX-512 sidecars add about 1.14 MiB for QV-10 and 7.69 MiB for QV-20. Scalar, AVX2, ARM, and WASM plans do not pay this cost. Each scalar descriptor and optional sidecar now live in one private entry, making their correspondence structural. The review changes were performance-neutral or better in balanced checks: both QV-10 and QV-20 medians were about 4% lower than the pre-review AVX binary.

## Portability

- non-dispatch builds have an explicit scalar runtime-ISA resolver
- PR CI compiles and runs that path without `CLIFFT_ENABLE_RUNTIME_DISPATCH`
- the no-dispatch smoke passes with both GCC and Clang locally
- the symbolic executor has no dependency on `svm_backend()` or SVM initialization
- SIMD types remain absent from common headers

## Validation

- 1,247 non-benchmark C++ tests in Debug and `x86-64-v2` Release builds
- 1,254 Python tests
- targeted symbolic executor tests under auto, forced scalar, and forced AVX2 selection
- QEMU Haswell and Nehalem symbolic execution, selecting AVX2 and scalar fallbacks respectively
- forced unsupported AVX-512/AVX2 requests under QEMU fail with a clean `CLIFFT_FORCE_ISA` exception rather than `SIGILL`
- object disassembly confirms no ZMM instructions in executor, scalar fusion, or shared-selector objects and packed ZMM instructions in the AVX-512 kernel object
- pre-commit hooks

## Deferred follow-ups

- low-pivot fused U4
- remaining direct rotation shapes
- AVX2 symbolic kernels
- OpenMP

These are separate kernel/parallelization decisions and are not included in this stack.

Refs #280

### DIFF
--- a/src/clifft/sampling/executor.h
+++ b/src/clifft/sampling/executor.h
@@ -122,7 +122,9 @@ class ExecutablePlan {
     };
 
     struct FusedRotationEntry {
+        // The architecture-neutral descriptor always supports scalar execution.
         PreparedFusedRotation rotation;
+        // The resolved ISA may attach prepared storage and a specialized kernel.
         FusedRotationSidecar sidecar;
     };
 
--- a/src/clifft/sampling/executor.h
+++ b/src/clifft/sampling/executor.h
@@ -7,6 +7,7 @@
 // writes records using only preallocated storage.
 
 #include "clifft/sampling/fused_rotation.h"
+#include "clifft/sampling/fused_rotation_simd.h"
 #include "clifft/sampling/kernels.h"
 #include "clifft/sampling/plan.h"
 #include "clifft/sampling/state.h"
@@ -243,6 +244,7 @@ class ExecutablePlan {
     std::vector<InstrumentDistribution> instrument_distributions_;
     std::vector<uint32_t> instrument_resume_offsets_;
     std::vector<PreparedFusedRotation> fused_rotations_;
+    std::vector<FusedRotationSidecar> fused_rotation_sidecars_;
     std::vector<Action> actions_;
 };
 
--- a/src/clifft/sampling/executor.h
+++ b/src/clifft/sampling/executor.h
@@ -121,6 +121,11 @@ class ExecutablePlan {
         uint32_t rotation_index = 0;
     };
 
+    struct FusedRotationEntry {
+        PreparedFusedRotation rotation;
+        FusedRotationSidecar sidecar;
+    };
+
     struct ExecutePromotion {
         PreparedPromotion promotion;
         PreparedExpression sign;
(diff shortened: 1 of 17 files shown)


src/clifft/sampling/executor.h
```diff
@@ -120,6 +121,11 @@ class ExecutablePlan {
         uint32_t rotation_index = 0;
     };
 
+    struct FusedRotationEntry {
+        PreparedFusedRotation rotation;
```
bachase, line 126: Should we just put this structure's members directly inside and avoid the nesting/simplify the type systme?

src/clifft/sampling/executor.h
```diff
@@ -120,6 +121,11 @@ class ExecutablePlan {
...
+        FusedRotationSidecar sidecar;
```
bachase, line 128: Should we comments on what these two members encapsulate?

src/clifft/sampling/executor.h
bachase: I would keep this composed. PreparedFusedRotation is the ISA-neutral descriptor shared by fusion construction, scalar execution, and AVX-512 sidecar preparation/execution. Flattening it into the executor-private entry would either couple those APIs to that private type or duplicate the descriptor surface. FusedRotationEntry only adds the structural pairing with an optional host-specific sidecar. I added member comments in 8e68b8e to make that boundary explicit.

src/clifft/sampling/executor.h
bachase: Agreed. Added comments in 8e68b8e: rotation is the always-available architecture-neutral descriptor, while sidecar is optional prepared storage plus a specialized kernel for the resolved ISA.

PR #286, opened 2026-08-09
refactor(sampling): separate executor responsibilities

bachase:
## Summary

- move validated CPU lowering and immutable prepared tables into ExecutablePlan
- keep Executor focused on mutable one-shot state, lifecycle, and action dispatch
- move multi-shot/result and replay orchestration into sampler
- keep exact final-state queries in a separate state-query interface
- hide scalar versus optional ISA-specific fused-rotation preparation behind a portable dispatch object

## Design constraints

This is a behavior-neutral ownership refactor. It does not add a pass manager, change the semantic or executable variants, alter the action tape or data layout, or change kernel behavior. The hot action loop and kernels remain in Executor; sampler crosses that boundary once per shot. Architecture-specific vector types remain confined to the AVX-512 translation unit, and the scalar descriptor remains the fallback and correctness path.

Runtime ISA selection is resolved and validated once per executable plan, then passed to each fused descriptor for optional sidecar preparation.

## Validation

- pre-commit on all tracked files
- 1,247 non-benchmark C++ tests passed
- 1,253 Python tests passed, with the non-runtime-dispatch forced-ISA test correctly skipped on Apple Silicon
- Release build of clifft_core


bachase (commented):
Split immutable plan lowering, per-shot execution, multi-shot orchestration, and exact state queries into focused components. Encapsulate fused-rotation ISA dispatch and resolve ISA selection once per executable plan.

PR #288, opened 2026-08-09
perf: vectorize direct symbolic rotations

bachase:
## Summary

- add separately compiled AVX-512 kernels for direct diagonal and non-diagonal symbolic-coordinate rotations
- select a one-byte kernel during `SamplingPlan` to `ExecutablePlan` lowering, without growing `ExecuteRotation` (72 bytes) or the action variant (112 bytes)
- evaluate dynamic signs once per action and retain scalar handling for identity/global phase, short diagonal states, low pivots, and the measured-unprofitable pivot-four shape
- keep ISA-specific types and data out of common descriptors; lane permutations and parity signs are shared static tables, with no per-action sidecar

The initial eligibility boundary was checked with a Release microbenchmark. Width-three diagonal and pivot-three paired rotations were profitable (about 1.24x and 1.26x respectively), as were pivots five and above. Pivot-four rotations were 8-18% slower than scalar across tested active widths, so they deliberately remain on the scalar path.

## Performance

Same Release build, one thread pinned on an AMD EPYC 9554P, runtime ISA `avx512`. Ratios below are symbolic / legacy, so lower is better.

| Case | Before | This PR |
|---|---:|---:|
| QV-10 seed 42, 5,000 shots | 1.346x | 1.02x |
| QV-10 seed 44, 5,000 shots | 1.597x | 1.12x |
| QV-10 seed 51, 5,000 shots | 0.814x | 0.77x |
| QV-20 seed 47, 3 shots | 0.708x | 0.73x |
| QV-20 seed 49, 3 shots | 0.883x | 0.83x |
| coherent d3 r3, 300,000 shots | 1.133x | 0.92x |
| k12/L512, 20,000 shots | 0.769x | 0.59x |

Surface d7 r7, cultivation d3, and distillation aggregate screens showed no material regression; cultivation also improved. One-shot QV-20 remains a setup guard and passed for both selected seeds.

The refreshed QV-10 seed-44 profile is now 37.1% scalar fused rotation, 34.9% AVX-512 fused U4, 12.2% direct AVX-512 rotation, 1.8% scalar direct rotation, and 9.8% measurement probability/collapse. That makes low-pivot fused U4/U2 the next independent kernel target; this PR intentionally does not broaden into it.

## Correctness and portability

- scalar-oracle checks for both dynamic-sign values, real and imaginary Pauli phase classes, every within-vector lane permutation, and identity/global phase
- explicit diagonal width-two/three and pairing pivot-two/three/four/five boundaries
- complete Release C++ suite under automatic, forced scalar, and forced AVX2 dispatch
- `x86-64-v2` build-flag audit and focused tests under automatic, forced scalar, and forced AVX2 dispatch
- QEMU execution on Haswell/forced-AVX2 and Nehalem/forced-scalar CPUs
- no-runtime-dispatch compile check for the generic executor path
- generated assembly contains packed `zmm`, `vpermpd`, and FMA operations
- pre-commit suite

Refs #280.

### DIFF
--- a/src/clifft/sampling/direct_rotation_dispatch.h
+++ b/src/clifft/sampling/direct_rotation_dispatch.h
@@ -4,11 +4,14 @@
 
 #include <cstdint>
 
+namespace clifft::internal {
+enum class RuntimeIsa;
+}
+
 namespace clifft::sampling {
 
-// The selector fits in ExecuteRotation's existing tail padding. Architecture-
-// specific data stays in its translation unit instead of expanding every
-// direct-rotation descriptor.
+// This selector occupies ExecuteRotation's existing tail padding. It carries
+// no architecture-specific data and does not expand the action descriptor.
 enum class DirectRotationKernel : uint8_t {
     Scalar,
     Avx512Diagonal,
@@ -17,6 +20,8 @@ enum class DirectRotationKernel : uint8_t {
 
 static_assert(sizeof(DirectRotationKernel) == 1);
 
+// Exposed separately from host ISA resolution so portable tests can verify the
+// structural eligibility boundaries even on machines without AVX-512.
 [[nodiscard]] inline DirectRotationKernel select_direct_rotation_kernel(
     const PreparedRotation& rotation, bool use_avx512) noexcept {
     if (!use_avx512 || rotation.pauli.is_identity()) {
@@ -27,16 +32,20 @@ static_assert(sizeof(DirectRotationKernel) == 1);
                                                 : DirectRotationKernel::Scalar;
     }
     const uint64_t pairing_bit = rotation.pauli.pair_selector;
-    // Pivot-four sweeps regress relative to scalar at every measured active
-    // width, unlike both the pivot-three boundary and pivots five and above.
+    // On the current Zen 4 performance host, pivot-four sweeps regress at every
+    // measured active width, unlike both the pivot-three boundary and pivots
+    // five and above.
     return pairing_bit >= (uint64_t{1} << 3) && pairing_bit != (uint64_t{1} << 4)
                ? DirectRotationKernel::Avx512HighPivot
                : DirectRotationKernel::Scalar;
 }
 
-// This function is linked only on x86-64 runtime-dispatch builds and must be
-// called only after the dispatcher has selected AVX-512.
-void apply_direct_rotation_avx512(State& state, const PreparedRotation& rotation,
-                                  DirectRotationKernel kernel, bool sign) noexcept;
+[[nodiscard]] DirectRotationKernel resolve_direct_rotation_kernel(
+    const PreparedRotation& rotation, internal::RuntimeIsa runtime_isa) noexcept;
+
+// Keeps ISA selection out of the executor while retaining the scalar kernel as
+// the portable fallback and correctness implementation.
+void apply_direct_rotation(State& state, const PreparedRotation& rotation,
+                           DirectRotationKernel kernel, bool sign) noexcept;
 
 }  // namespace clifft::sampling
--- a/src/clifft/sampling/direct_rotation_dispatch.h
+++ b/src/clifft/sampling/direct_rotation_dispatch.h
@@ -14,31 +14,11 @@ namespace clifft::sampling {
 // no architecture-specific data and does not expand the action descriptor.
 enum class DirectRotationKernel : uint8_t {
     Scalar,
-    Avx512Diagonal,
-    Avx512HighPivot,
+    Diagonal,
+    HighPivot,
 };
 
 static_assert(sizeof(DirectRotationKernel) == 1);
-
-// Exposed separately from host ISA resolution so portable tests can verify the
-// structural eligibility boundaries even on machines without AVX-512.
-[[nodiscard]] inline DirectRotationKernel select_direct_rotation_kernel(
-    const PreparedRotation& rotation, bool use_avx512) noexcept {
-    if (!use_avx512 || rotation.pauli.is_identity()) {
-        return DirectRotationKernel::Scalar;
-    }
-    if (rotation.pauli.is_diagonal()) {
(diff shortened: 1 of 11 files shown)


bachase:
Addressed the final review feedback in `c705906`:

- Split direct rotation dispatch into `direct_rotation_dispatch.{h,cc}` and the x86-only declaration in `direct_rotation_simd.h`, matching the fused-rotation boundary.
- Moved runtime-ISA resolution and all `CLIFFT_ENABLE_RUNTIME_DISPATCH`/kernel branching out of `Executor`; it now evaluates the sign and calls `action.apply(state_, sign)`.
- Retained the byte-sized selector and added a compile-time guard that `ExecuteRotation` remains 72 bytes. The overall action variant remains 112 bytes.
- Used `action.apply(...)` instead of an owning `action.rotation.apply(...)`: wrapping the 64-byte portable rotation plus the selector as a separate member would round that member to 72 bytes and expand the containing action to 80 bytes once the expression register is added.
- Added scalar-oracle coverage for high-pivot X masks with intermediate high bits, for both dynamic-sign values, plus the laneless diagonal probe.
- Scoped the pivot-four comment explicitly to the current Zen 4 performance host.

Validation:

- Native Release suite: 1,259 cases / 407,885 assertions passed.
- Focused direct-rotation tests passed under automatic AVX-512, forced scalar, and forced AVX2 selection.
- A no-runtime-dispatch core build passed.
- An x86-64-v2 Release build and the same automatic/scalar/AVX2 focused tests passed. Compile commands confirm the generic dispatch TU uses only the v2 baseline while AVX-512 flags remain confined to `kernels_avx512.cc`.
- Required pre-commit checks passed.
- Balanced before/after timing was neutral-to-better: QV-10 seed 44 was 1-3% faster in all three pairs; coherent d5 r1 was within about +0.1% to +0.5% after the first warm-up pair. Checksums matched.

src/clifft/sampling/direct_rotation_dispatch.h
```diff
@@ -0,0 +1,51 @@
...
+// no architecture-specific data and does not expand the action descriptor.
+enum class DirectRotationKernel : uint8_t {
+    Scalar,
+    Avx512Diagonal,
+    Avx512HighPivot,
+};
+
+static_assert(sizeof(DirectRotationKernel) == 1);
+
+// Exposed separately from host ISA resolution so portable tests can verify the
```
bachase: This feels like an awkward design choice more oriented around testing. Maybe its the best approach, but is there a better multi-ISA strategy/ perhaps similarly to the SVM approach?

src/clifft/sampling/executable_plan.h
```diff
@@ -60,8 +61,16 @@ class ExecutablePlan {
     struct ExecuteRotation {
```
bachase, line 61: Consider adding comments to members to make it clear what each is for.

src/clifft/sampling/direct_rotation_dispatch.h
```diff
@@ -0,0 +1,51 @@
...
+    const PreparedRotation& rotation, bool use_avx512) noexcept {
+    if (!use_avx512 || rotation.pauli.is_identity()) {
+        return DirectRotationKernel::Scalar;
+    }
+    if (rotation.pauli.is_diagonal()) {
+        return rotation.pauli.active_width >= 3 ? DirectRotationKernel::Avx512Diagonal
+                                                : DirectRotationKernel::Scalar;
+    }
+    const uint64_t pairing_bit = rotation.pauli.pair_selector;
+    // On the current Zen 4 performance host, pivot-four sweeps regress at every
```
bachase: IS this a sign we are overfitting to Zen 4?

src/clifft/sampling/kernels_avx512.cc
bachase, line 1: Throught this file, is it worth adding a few comments to each function/function section explaining what it is doing?

src/clifft/sampling/direct_rotation_dispatch.h
bachase: Addressed in b8ea9e4. I agreed that the public boolean selector was test-shaped. The action tag is now ISA-neutral (Scalar, Diagonal, HighPivot), and the production resolver takes the shared RuntimeIsa, matching the fused-dispatch boundary and giving AVX2 an explicit extension point. I also benchmarked the closer SVM-style alternative of a process-wide function pointer: it regressed QV-10 seed 44 by 2-5% because this would be one indirect call per rotation, whereas SVM amortizes its indirect call over an entire program. The wrapper therefore keeps one process-wide ISA value plus a predicted shape branch and direct call; paired QV and coherent d5 r1 measurements were neutral versus the prior wrapper.

src/clifft/sampling/executable_plan.h
bachase: Addressed in b8ea9e4. Added concise comments for the portable rotation geometry and weights, dynamic sign register, and host-selected shape tag, including why the tag does not grow the descriptor.

src/clifft/sampling/direct_rotation_dispatch.h
bachase: I kept the pivot-four fallback, but revised the wording in b8ea9e4 so it does not present the Zen 4 result as a universal architectural fact. This is a conservative policy: pivot four regressed 8-18% against scalar at every measured width on the current performance host, while a benefit on Intel is currently hypothetical. Enabling it broadly would exchange a known regression for an unmeasured gain. The new ISA-neutral tags and RuntimeIsa resolver leave room to change this policy per implementation once another host provides evidence.

src/clifft/sampling/kernels_avx512.cc
bachase: Addressed in b8ea9e4. Added why-oriented comments around lane-sign construction, diagonal updates, high-pivot block pairing and lane permutation, fused orbit traversal, real versus imaginary phase specialization, and sidecar preparation.

PR #289, opened 2026-08-10
perf(sampling): make symbolic planning forward

bachase:
## Summary

- replace eager pending-suffix coordinate rewrites with a cumulative coordinate frame that localizes each Pauli when planning reaches it
- accumulate noise, feedback, measurement branches, and instrument destination flips in a packed forward symbolic Pauli frame
- preserve the semantic `SamplingPlan` / `AffineBool` boundary and the existing flat symbol-major executable dependency tape
- add symbolic planning and executable-preparation stages to the native compile profiler

The symbolic frame follows the same broad `p_x` / `p_z` idea as the legacy virtual frame and the symbolic Clifford-Pauli-frame separation described by [SymFT](https://arxiv.org/abs/2607.28600). Runtime topology planning, the legacy instruction ABI, and hot execution are unchanged.

## Planner-stage results

RelWithDebInfo on Apple Silicon; medians compare the same fixtures before and after this branch:

| Fixture | Before | After | Speedup | Frame workspace |
| target QEC | 2.94 ms | 0.65 ms | 4.5x | 7,684 B |
| cultivation d5 | 159.1 ms | 12.24 ms | 13.0x | 172,804 B |
| QV-20 seed 42 | 10.13 ms | 2.23 ms | 4.5x | 368 B |

The remaining cultivation planner profile is dominated by cumulative tableau composition/inversion. Executable preparation and semantic affine-expression ownership are not the current planner bottleneck.

## Batching disposition

This PR does not implement batching. It keeps batching straightforward by leaving `SamplingPlan` lane-independent; `ExecutablePlan` continues to lower affine expressions into its scalar symbol-major dependency tape, while a future batch executor can lower the same semantic expressions into a bit-sliced lane layout.

## Validation

- `ctest --test-dir build-debug --output-on-failure -j4 -E "^Bench:"`: 1,253/1,253 passed
- `uv run --frozen --only-group dev pre-commit run --all-files --show-diff-on-failure`: passed
- focused planner and canonical affine-expression regressions
- symbolic compile-profiler smoke checks for both symbolic and legacy backends

Refs #247. Supports the planner and continuation-compile track in #280.

### DIFF
--- a/src/clifft/sampling/planner.cc
+++ b/src/clifft/sampling/planner.cc
@@ -7,11 +7,16 @@
 #include "stim.h"
 
 #include <algorithm>
+#include <bit>
+#include <cassert>
 #include <cmath>
 #include <complex>
 #include <cstdint>
+#include <limits>
 #include <numbers>
+#include <numeric>
 #include <optional>
+#include <span>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
@@ -78,8 +83,8 @@ struct PendingReadoutNoise {
 };
 
 struct PendingInstrument {
-    // Source observable and computational destination correction, expressed
-    // in the coordinate basis that future planner transformations update.
+    // Source observable and computational destination correction stay in the
+    // initial HIR coordinates used by the cumulative symbolic Pauli frame.
     Pauli body;
     Pauli destination_flip;
     // Maps the source observable's eigenspaces to physical G and E; earlier
@@ -87,7 +92,7 @@ struct PendingInstrument {
     AffineBool sign;
     // Indexes both the plan-owned distribution and its continuation boundary.
     InstrumentSiteId site{};
-    // Reserved in HIR order so a rewritten suffix cannot renumber the prefix.
+    // Reserved in HIR order so coordinate evolution cannot renumber the prefix.
     SymbolId destination_flip_symbol{};
     // Continuations retain only noise and symbols preceding this HIR site.
     uint32_t next_noise_site = 0;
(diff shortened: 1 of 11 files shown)


bachase:
Addressed in f1765480:

- Added structured Activate-instrument coverage after an existing active coordinate, including exact source/sign, destination-flip propagation, and the following measurement correction.
- Removed the local inverse/scatter from Activate by using the promotion-frame invariant directly.
- Made the pending operation stream const.
- Labeled the symbolic-frame workspace metric as an estimate and documented its layout coupling and planner-only scaling boundary.

I intentionally did not add a stored whole-plan digest or land the old/new differential harness. With the legacy path planned for deprecation, exact internal action layout should not become a compatibility contract. The new focused regression preserves the semantic behavior most exposed by this rewrite while allowing later planner representation changes.

src/clifft/sampling/planner.cc
```diff
@@ -162,6 +167,173 @@ bool anticommutes(const Pauli& left, const Pauli& right) {
     return !left.ref().commutes(right.ref());
 }
 
+std::vector<size_t> identity_indices(size_t size) {
+    std::vector<size_t> result(size);
+    std::iota(result.begin(), result.end(), size_t{0});
+    return result;
+}
+
```
bachase: Should we extract these frames to their own .h/.cc that is standalone testable? We don't have another use case yet, but wondering if we are packing too much into planner.cc?

bachase: Agreed. Extracted the coordinate, symbolic, promotion, and measurement frame algebra into an internal planner_frame.h/.cc component in 476182c3. Added direct tests for coordinate composition/round-trips, multiword affine corrections, symbol bounds, promotion localization, and measurement-frame installation. Keeping it in clifft::sampling::internal avoids creating a public API or constraining a future batched representation. The split also lets the profiler query the symbolic frame workspace estimate instead of copying its layout formula.

PR #290, opened 2026-08-10
perf: vectorize lane-paired symbolic rotations

bachase:
## Summary

- add an AVX-512 direct-rotation kernel for pairing pivots 0 through 2, where both members of every coefficient pair fit in one eight-lane vector
- derive partner amplitudes with an in-register XOR permutation and use the existing lane parity tables without adding a per-action sidecar
- retain the compact one-byte kernel selector and scalar fallback for AVX2, scalar-only, narrow-width, identity, and pivot-4 shapes
- compare every low-lane X mask against the scalar oracle for both signs and both real/imaginary phase cases, including imaginary-class shapes whose Z mask extends above the vector lanes
- name the selector's lane-count, width-floor, and pivot-four constants in the shared SIMD header so the portable selector and the AVX-512 translation unit cannot drift

## Evidence

On the issue #280 Release-with-symbols build, one thread pinned to CPU 3 on the AMD EPYC 9554P:

- coherent d5 r1 symbolic median: 11.25 s on merged main, 8.16 s here (1.38x faster); the established legacy median is 6.51 s
- coherent d3 r3: approximately 1.16x faster in the stable paired blocks
- k12/L512: approximately 1.20x faster
- QV-10 and QV-20 guards: neutral to slightly faster, depending on circuit seed
- surface, cultivation, and distillation guards: neutral

The isolated kernel is 1.34x faster than scalar at width 3 and 2.3x-3.8x faster from widths 4 through 12. Generated assembly contains packed `zmm`, `vpermpd`, and FMA operations.

The post-change coherent d5 r1 profile is 48.5% vector direct rotation, 21.6% scalar pivot-4 rotation, and 25.5% measurement probability/collapse. Two pivot-4 prototypes regressed end-to-end and are intentionally excluded.

## Tests

- `ctest --test-dir /tmp/clifft-issue280-fusion-build --output-on-failure -E '^Bench:'`
- `ctest --test-dir /tmp/clifft-issue280-fusion-build --output-on-failure -R '^Bench:'`
- direct-rotation differential tests under native AVX-512, `CLIFFT_FORCE_ISA=avx2`, and `CLIFFT_FORCE_ISA=scalar` (1,311 assertions each)
- scalar-only runtime ISA compile check

### DIFF
--- a/src/clifft/sampling/direct_rotation_dispatch.cc
+++ b/src/clifft/sampling/direct_rotation_dispatch.cc
@@ -18,6 +18,10 @@ DirectRotationKernel select_direct_rotation_avx512(const PreparedRotation& rotat
                                                 : DirectRotationKernel::Scalar;
     }
     const uint64_t pairing_bit = rotation.pauli.pair_selector;
+    if (pairing_bit < (uint64_t{1} << 3)) {
+        return rotation.pauli.active_width >= 3 ? DirectRotationKernel::LanePaired
+                                                : DirectRotationKernel::Scalar;
+    }
     // Pivot four has a distinct stride-16 access pattern that regressed against
     // scalar at every measured width, so it remains on the fallback until a
     // kernel designed for that shape is available.
--- a/src/clifft/sampling/direct_rotation_dispatch.cc
+++ b/src/clifft/sampling/direct_rotation_dispatch.cc
@@ -9,25 +9,31 @@ namespace clifft::sampling {
 
 namespace {
 
+// One full vector block: the dense state must span the vector lanes before a
+// whole-block kernel is profitable or even addressable.
+constexpr uint32_t kMinVectorActiveWidth = 3;
+static_assert(uint64_t{1} << kMinVectorActiveWidth == kAvx512DoubleLanes);
+
+// Stride-16 pairing regressed against scalar at every measured active width,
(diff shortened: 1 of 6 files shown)


src/clifft/sampling/direct_rotation_dispatch.cc
```diff
@@ -18,6 +18,10 @@ DirectRotationKernel select_direct_rotation_avx512(const PreparedRotation& rotat
                                                 : DirectRotationKernel::Scalar;
     }
     const uint64_t pairing_bit = rotation.pauli.pair_selector;
+    if (pairing_bit < (uint64_t{1} << 3)) {
```
bachase: These are some magic numbers here and next line-- worth at least a comment if not a constant?

bachase: Named in 3145c42: the lane count now lives in `direct_rotation_simd.h` as `kAvx512DoubleLanes` (shared with the kernel TU so they cannot drift), the width floor is `kMinVectorActiveWidth` with a `static_assert` tying it to the lane count, and the pivot-four exclusion is `kPivotFourSelector` carrying the measurement rationale. The redundant `pairing_bit >= 8` conjunct fell out in the process.

PR #292, opened 2026-08-10
perf(sampling): vectorize lane-paired active measurements

bachase:
## Summary

- add an AVX-512 probability and collapse kernel for active measurements whose Pauli pairing stays within an eight-double lane block
- select the kernel once while lowering the executable plan, storing the one-byte selector in existing action padding
- compact the selected branch in place using fixed permutations and four-double stores, with the existing scalar implementation as the correctness and portability fallback
- retain scalar execution for diagonal, high-pivot, AVX2/scalar, and one-vector-block width-3 shapes

No allocation, exception, runtime topology planning, architecture-specific descriptor data, or action-size growth is introduced.

## Performance

Paired medians from the same pinned Release build and benchmark harness on the AVX-512 performance host:

| Workload | main symbolic | this PR | change |
| --- | ---: | ---: | ---: |
| coherent d5 r1, 20k shots | 1.645 s | 1.373 s | 16.5% faster |
| coherent d3 r3, 100k shots | 0.236 s | 0.219 s | 7.1% faster |
| k12/L512, 20k shots | 0.463 s | 0.283 s | 39.0% faster |
| noncomputational d17 r5, 500 shots | 1.330 s | 1.054 s | 20.8% faster |

For coherent d5 r1, symbolic is now within about 5% of legacy on this host. Coherent d3 r3 and k12/L512 remain faster than legacy. Noncomputational execution remains about 5.2x slower than legacy because its dominant high-pivot instrument activation path is unchanged by this PR.

QV measurements are diagonal and therefore structurally unaffected. A same-layout width-3 A/B showed a 1.6% regression when vectorizing one-block measurements, so selection starts at active width 4.

## Verification

- full native Release-with-symbols test suite: 1,271 cases, 557,826 assertions
- full suite with `CLIFFT_FORCE_ISA=avx2`: 1,271 cases, 408,546 assertions
- full suite with `CLIFFT_FORCE_ISA=scalar`: 1,271 cases, 408,546 assertions
- no-runtime-dispatch generic build links successfully
- exhaustive scalar-oracle coverage for widths 3 through 6, all low-lane X masks, all Z masks, valid low pivots, both Pauli phase classes, and both measurement branches
- generated assembly contains packed AVX-512 permutations, FMA operations, and compact AVX stores
- required pre-commit checks pass

Checksums match main for every paired benchmark.

Contributes to #280.

### DIFF
--- a/src/clifft/sampling/kernels_avx512.cc
+++ b/src/clifft/sampling/kernels_avx512.cc
@@ -29,10 +29,11 @@ constexpr size_t kDimension = 4;
 constexpr size_t kMatrixSize = kDimension * kDimension;
 constexpr double kInvSqrt2 = 0.707106781186547524400844362104849039;
 
-// Lanes whose measurement pivot is zero are the unique representatives of
-// the four Pauli pairs in a vector block. Fixed permutations place those lanes
-// in the low half for an ordinary four-double store; hardware compress-stores
-// were substantially slower on the performance host.
+// Measuring a Pauli P pairs |b> with |b xor x>. Lanes whose measurement pivot
+// is zero enumerate the four pairs in a vector block exactly once: pivot zero,
+// for example, selects |0>, |2>, |4>, and |6>. Fixed permutations place those
+// representatives in the low half for an ordinary four-double store; hardware
+// compress-stores were substantially slower on the performance host.
 constexpr std::array<__mmask8, 3> kMeasurementSourceMasks = {0x55, 0x33, 0x0f};
 
 using LaneIndices = std::array<uint64_t, kLanes>;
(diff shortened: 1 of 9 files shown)


src/clifft/sampling/kernels_avx512.cc
bachase, line 1: Per my usual feedback, add a few concise comments throughout this new kernels for the reader to better understand what is going on. Tie at least some back to traditional quantum mechanics notation and terminology (or even with a small example), so we understand how lanes relate to e.g. four pauli pairs  depending on measurement pivot.

bachase: Addressed in 96c840b. The kernel comments now connect SIMD lanes to the four |b>, |b xor x> Pauli pairs, state the compacted projector formula, explain the real/imaginary phase split, and describe why removing the pivot permits safe in-place compaction. I also renamed the lane-index bound separately from the measured active-width profitability threshold.

PR #300, opened 2026-08-11
perf(sampling): streamline symbolic frame planning

bachase:
## Summary

- iterate symbolic Pauli corrections through packed nonidentity support instead of querying every physical qubit through `stim::bit_ref`
- preserve affine terms emitted by packed symbolic-frame rows because they are already sorted and unique
- move temporary affine storage into empty XOR accumulators instead of allocating and copying it
- cover mixed X, Y, and Z corrections across packed-word boundaries and the affine move path

This is planner-only work. It does not change the executable plan, RNG schedule, sampling APIs, or hot execution.

## Performance

Measured from the same native Release build on an AMD EPYC 9554P, pinned to CPU 3. Primary cases use five alternating 200-compile blocks; numbers are medians of block medians. The baseline is merged `main` at `7141819`.

| Circuit | Baseline plan | PR plan | Change | Baseline total | PR total | Change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| surface d7 r7 | 15.31 ms | 6.32 ms | -58.7% | 20.06 ms | 10.91 ms | -45.6% |
| cultivation d5 | 9.17 ms | 4.51 ms | -50.8% | 13.47 ms | 8.63 ms | -35.9% |

Three alternating 100-compile guard blocks showed total compile-time improvements of 23.0% on cultivation d3, 23.5% on distillation, 22.1% on target-QEC, 8.3% on coherent d3 r3, and 5.7% on coherent d5 r1.

Five alternating 200-compile QV guard blocks were neutral within noise: QV-10 was +1.3% and QV-20 was +0.2% in total compile time. `SymbolicPauliFrame::apply` was below 0.2% of the QV-20 baseline profile, and QV compile time is dominated by executable preparation, so there is no causal regression signal from these planner changes.

Against current legacy compilation, the remaining absolute excess is approximately 4.36 ms on surface (10.91 versus 6.56 ms) and 4.75 ms on cultivation d5 (8.63 versus 3.88 ms). Post-change profiles are diffuse: affine XOR and allocation, coordinate lookup, plan validation, operation queuing, and executable lowering each account for only a few percent. This PR therefore stops before introducing broader storage or pass machinery.

## Test plan

- [x] C++ tests pass (`ctest --test-dir /tmp/clifft-direct-debug2 -E '^Bench:' --output-on-failure`): 1,276 tests
- [ ] Python tests pass (`uv run pytest tests/python/ -v`)
- [ ] Doc examples pass (`uv run pytest --codeblocks docs/ README.md -v`)

Python and documentation coverage are left to CI because this change is confined to private C++ planner data structures.

## Contributor agreement

- [x] I confirm this contribution is my original work (or I have the right to submit it) and I license it under this project's [Apache-2.0 license](../LICENSE).
- [x] If this contribution was AI-assisted, I have reviewed the generated code and included an `Assisted-by:` git trailer in the commit message.

Closes no issue; contributes to #280.

### DIFF
--- a/src/clifft/sampling/plan.cc
+++ b/src/clifft/sampling/plan.cc
@@ -3,6 +3,7 @@
 #include "clifft/util/numeric.h"
 
 #include <algorithm>
+#include <cassert>
 #include <cmath>
 #include <iomanip>
 #include <limits>
@@ -245,8 +246,17 @@ AffineBool::AffineBool(bool constant) : constant_(constant) {}
 AffineBool::AffineBool(bool constant, std::vector<SymbolId> terms)
     : constant_(constant), terms_(canonicalize_terms(std::move(terms))) {}
 
+AffineBool::AffineBool(bool constant, std::vector<SymbolId> terms, CanonicalTermsTag)
+    : constant_(constant), terms_(std::move(terms)) {
+    assert(is_canonical() && "trusted affine terms must be sorted and unique");
+}
+
 AffineBool AffineBool::symbol(SymbolId id) {
-    return AffineBool(false, {id});
+    return AffineBool(false, {id}, CanonicalTermsTag{});
+}
+
+AffineBool AffineBool::from_canonical_terms(bool constant, std::vector<SymbolId> terms) {
+    return AffineBool(constant, std::move(terms), CanonicalTermsTag{});
 }
 
 bool AffineBool::is_canonical() const {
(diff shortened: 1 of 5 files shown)


bachase:
Addressed the actionable review findings in `95f7708`.

- Non-empty rvalue affine XOR now delegates to the const-reference implementation, restoring its single-symbol in-place fast path while retaining storage transfer into an empty destination.
- Sparse Pauli traversal now masks the final packed word in Release builds, so stray padding cannot produce an out-of-range frame index.
- Added a 65-qubit structural test covering a valid boundary bit together with nonzero padding.

I also prototyped replacing the packed loop with Stim's `for_each_active_pauli`. Although cleaner, its callback must query X and Z through `bit_ref` for every active correction. Five alternating blocks showed roughly 4-6% planner regressions on both surface d7 r7 and cultivation d5, so I did not retain it in this performance PR.

With the shared XOR path and tail mask, five alternating 400-compile blocks improved over the prior PR head:

| Circuit | Plan change | Total change |
| --- | ---: | ---: |
| surface d7 r7 | -4.6% | -3.2% |
| cultivation d5 | -5.6% | -5.0% |

QV-10 and QV-20 remained neutral in total compile time. The full non-benchmark Debug suite passes: 1,277 tests, plus pre-commit checks.

I left `from_canonical_terms` unchanged. `SamplingPlan` is a private source-level IR, the factory has one trusted producer, Debug validates its precondition, and plan-stored expressions receive Release-mode validation. A friend/passkey layer would add machinery without closing a demonstrated failure path.

Issue #301, opened 2026-08-11
Rewrite documentation for the symbolic sampling architecture


bachase:
## Goal

Update the documentation so theory, user guides, references, and performance material describe symbolic sampling and no longer teach the legacy SVM as the current architecture.

This tracker covers documentation work only. Backend migration, playground implementation, and the executable refactor remain tracked separately.

## Terminology

- Use **symbolic-coordinate sampling** for the enduring method and **symbolic sampling pipeline** for its compiler/execution architecture.
- Use **symbolic sampler** when a short distinction from the legacy path is useful.
- After cutover, ordinary user documentation should normally say **the Clifft sampler**, without presenting symbolic as a selectable backend.
- Reserve **backend** for concrete targets such as CPU, WASM, or a future GPU implementation.
- Use **legacy SVM** only in migration notes, historical material, or explicit comparisons.
- Use **active width** $k$ for the number of active stabilizer coordinates, including **peak active width** for the maximum reached by a plan.
- Use **active-state dimension** $2^k$ for the number of amplitudes in the dense active state. Do not use active dimension or active rank as synonyms for $k$.
- When documenting `program.peak_rank`, describe the value it reports as the peak active width.

## Sequencing

- The factored-state theory update can land now.
- Architecture details should follow the stable outcome of #298.
- Symbolic-only user and reference documentation should land with the legacy removal.
- Performance material should use final post-#285 measurements.

Work is divided into the attached sub-issues. Completion requires a successful documentation build, runnable examples, and no unintended current-facing SVM or bytecode documentation.

bachase:
Completed by #349. All seven documentation sub-issues are now closed.

The current documentation describes Clifft's sole sampling pipeline, the v0.8 migration article records the method evolution and API changes, current release-target performance has explicit provenance, and the final sweep validated the strict docs build, executable examples, long-form tutorials, terminology, and external links.

Closing this documentation tracker as complete.

PR #308, opened 2026-08-11
docs: explain symbolic factored-state theory

### LINKED ISSUE #307: Rewrite the factored-state and symbolic-sampling theory
## Goal

Rewrite the theory material around the semantic invariant used by symbolic sampling.

## Scope

- Present the factorization per sampled trajectory.
- Define active and dormant stabilizer coordinates, rather than physical-qubit subsets.
- Explain the compile-time coordinate frame and symbol-dependent Pauli frame.
- Show how Pauli-frame effects become affine signs and outcomes in SamplingPlan.
- Explain promotion, active measurement, and dormant measurement as active-width transitions.
- Correct legacy runtime-frame and normalization descriptions across linked theory pages.

This conceptual work can land before #298. Avoid depending on proposed implementation names that are not yet present.

bachase:
## Summary

- trace the method lineage from the original Clifft factored active-state design through the SymFT refinements adopted back into current Clifft
- name Wang Fang, Huazhe Lou, and Riling Li at first use and add the complete SymFT reference
- reserve active width for $k$ and active-state dimension for $2^k$
- define the state per sampled trajectory using active and dormant stabilizer coordinates
- explain affine Pauli-frame signs and adaptive coordinate transitions with concrete examples
- walk readers from circuit text through `SamplingPlan`, executable preparation, `Executor`, and per-shot results
- add visible method attribution to the README

This is the theory workstream from #301, rebased onto the symbolic-only main branch after the production cutover and legacy SVM removal in #321 and #334. The remaining documentation issues will audit the current user-facing pages and add final performance and release material rather than duplicate this theory page. The long-form migration and performance update is tracked in #322.

Closes #307

## Validation

- [x] `just docs-build`
- [x] `just py-doctest` — 31 passed, 7 skipped
- [x] `uv run --frozen --only-group dev pre-commit run --all-files --show-diff-on-failure`

### DIFF
--- a/docs/theory/overview.md
+++ b/docs/theory/overview.md
@@ -22,8 +22,8 @@ choices.
 ## Symbolic Clifford Coordinates
 
 Sampling a circuit with noise or mid-circuit measurements produces a
-trajectory. At action boundary $j$, condition on the Boolean outcomes
-$s_{\le j}$ sampled so far. The resulting pure state has the factorization
+trajectory. After planned step $j$, let $s_{\le j}$ denote the Boolean outcomes
+sampled so far. The resulting pure state has the factorization
 
 $$
 |\psi_j(s_{\le j})\rangle
@@ -41,8 +41,8 @@ $$
 The factors have distinct roles:
 
 - **$C_j$ (Clifford coordinate map):** Maps the current stabilizer coordinates
-  into the physical qubit basis. Its evolution is determined during
-  compilation and planning, not by per-shot tableau operations.
+  into the physical qubit basis. The front end and planner determine its
+  evolution before sampling; a shot performs no tableau operations.
 
 - **$P_j(s_{\le j})$ (Pauli frame):** Represents branch-dependent Pauli
   corrections from noise, measurement outcomes, and classical feedback. It is
@@ -61,11 +61,12 @@ The factors have distinct roles:
 - **$\gamma_j$ (global scalar):** Carries a common complex weight and phase
   outside the active coefficient array.
 
-Active and dormant coordinates are not subsets of physical qubits. A planned
-Clifford basis change can change the physical Pauli support of every coordinate
-while preserving this factorization. "Active width" therefore means the number
-of stabilizer coordinates represented in the dense coefficient array, not the
-number of physical qubits touched by the circuit.
+Active and dormant coordinates are basis elements, not subsets of physical
+qubits. After Clifford gates change the basis, one coordinate may represent a
+different, possibly multi-qubit, physical Pauli without changing the size of
+the active array. "Active width" therefore means the number of stabilizer
+coordinates represented in the dense coefficient array, not the number of
+physical qubits touched by the circuit.
 
 For a normalized physical circuit, the unconditional noisy state is the
 ensemble over trajectories,
@@ -92,12 +93,19 @@ measurements can return coordinates to the dormant set, so fault-tolerant
 circuits may retain a small $k_{\max}$ even when they contain hundreds of
 physical qubits.
 
-## Compiling the Pauli Frame into Symbolic Signs
+## Planning Pauli-Frame Effects as Symbolic Signs
+
+The main symbolic idea is to resolve the branch-dependent Pauli frame during
+planning instead of updating it after every event in every shot. Following the
+symbolic-frame strategy introduced by
+[Fang, Lou, and Li](https://arxiv.org/abs/2607.28600), the planner expresses
+each relevant dependence as an affine formula of Boolean symbols and attaches
+the resulting sign to the affected operation.
 
 Each stochastic event needed later in a shot receives a Boolean symbol. A
 symbol may represent a presampled Pauli fault, a sampled measurement branch, a
-readout flip, an instrument outcome, or a named parity of earlier symbols.
-Classical effects are represented as affine Boolean expressions,
+readout flip, or an [instrument outcome](noncomputational.md). These effects
+are represented as affine Boolean expressions,
 
 $$
 \ell(s) = c \oplus \bigoplus_{r \in R} s_r.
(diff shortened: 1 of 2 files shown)


docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
 
-For magic-state preparation, distillation, and related QEC circuits, frequent
-measurements can keep this active width small even when the physical circuit
-contains hundreds of qubits. Clifft then allocates $2^{k_max}$ amplitudes
-instead of $2^n$.
+## Symbolic Clifford Coordinates
 
-## Compilation and Execution
+Sampling a circuit with noise or mid-circuit measurements produces a
+trajectory. At action boundary $j$, condition on the Boolean outcomes
```
bachase: Do we need the concept of action boundary? It hasn't been introduced yet. Is there a lighter weight notation to use?

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+
+- **$|\phi_j\rangle_{A_j}$ (active coefficient state):** A dense complex array
+  of size $2^{k_j}$ containing the non-Clifford interference. Ordinary sampling
+  keeps this coefficient state normalized up to floating-point drift.
+
+- **$\gamma_j$ (global scalar):** Carries a common complex weight and phase
+  outside the active coefficient array.
+
+Active and dormant coordinates are not subsets of physical qubits. A planned
+Clifford basis change can change the physical Pauli support of every coordinate
```
bachase: "A planned Clifford basis change can change the physical Pauli support of every coordinate while preserving this factorization. " is a bit dense -- what is this trying to accomplish?

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+scratch storage scale as $O(2^{k_{\max}})$ instead of $O(2^n)$. Frequent
+measurements can return coordinates to the dormant set, so fault-tolerant
+circuits may retain a small $k_{\max}$ even when they contain hundreds of
+physical qubits.
+
+## Compiling the Pauli Frame into Symbolic Signs
+
+Each stochastic event needed later in a shot receives a Boolean symbol. A
+symbol may represent a presampled Pauli fault, a sampled measurement branch, a
+readout flip, an instrument outcome, or a named parity of earlier symbols.
```
bachase: Is " named parity of earlier symbols" really just detector and observable operations? Maybe drop that detail at this level of theory and only mention the others? And have instrument be a hyperlink to the noncomputational theory section?

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+Classical effects are represented as affine Boolean expressions,
```
bachase: Is there a better transiton here, maybe pulling up "Following the symbolic-frame strategy introduced by
[Fang, Lou, and Li](https://arxiv.org/abs/2607.28600), the Pauli frame's dependenc at any time depends only on affine boolean formulas of these symbols, which can be resolved at compile time."... Basically setup at a high level the point, so readers that don't want the details can skip ahead.

(Side note, we are mixing compiling and planning in a few spot .. might be best to standardize)

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+
+Following the symbolic-frame strategy introduced by
+[Fang, Lou, and Li](https://arxiv.org/abs/2607.28600), the planner computes
+$\ell_Q$ once and attaches it to the planned operation. It also maps the
+unsigned body of $Q$ into the current stabilizer coordinates. Operations that
+need coefficient work are consequently represented by an active-coordinate
+Pauli and an affine sign, rather than by a physical Pauli string plus a mutable
+runtime frame.
+
+The full symbolic Pauli frame is planning workspace. Before hot execution, its
```
bachase: The change above to improve the start of these section can also simplify this closing section to be more "As a result, rather than needing to carry and update the Pauli frame at each operation of the circuit, we instead resolve its action symbolically on the other operations (e.g. rotations), and need only reference the symbols realized value at sample time."

Or similar/clear and concise version of this

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+## Adaptive Stabilizer Coordinates
+
+The planner maintains a stabilizer-destabilizer basis in which the first
+$k_j$ coordinates are active and the remainder are dormant. It resolves basis
+changes, Pauli support, measurement pivots, and active-width transitions once
+for all shots. The resulting semantic actions have the following effects:
+
+| Situation | Planned state action | Width transition |
+|---|---|---|
+| Pauli rotation supported on active coordinates | Rotate the active Pauli directly | $k \to k$ |
```
bachase: Should these each have a concrete example gate? Its pretty abstract as written?

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+
+The full symbolic Pauli frame is planning workspace. Before hot execution, its
+effects have been lowered into plan expressions, expression registers, and
+record outcomes. A shot evaluates those prepared expressions as symbols become
+available; it does not update an $n$-qubit Pauli frame after each fault or
+measurement.
+
+## Adaptive Stabilizer Coordinates
+
+The planner maintains a stabilizer-destabilizer basis in which the first
```
bachase, line 136: Should we make this more explicit? The basis is Z_k on dormant qubits, and then activation/collapse is what changes them?

The downside is of course this might expand to be too much detail in this theory summary.

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
-    v
-Sampling plan
-    |  Prepare fixed action storage and select scalar/SIMD kernels.
-    v
-Executable program
-    |  Reuse the plan for every shot.
-    v
-Measurement, detector, observable, and expectation-value results
+Circuit
+  -> Heisenberg IR (HIR)
```
bachase: Should there be a visual or other way to distinguish representations in this series from the functions/operations that are moving betwen them? I liked the old one for the legacy pipeline that had a little more detail interspersed in between

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
-
-See [Basis-State Probabilities](basis_probabilities.md) and
-[Strong Simulation with Exact Probabilities](../guide/strong-simulation.md).
+The front end absorbs physical Clifford operations into an offline tableau and
+maps relevant operations into the Heisenberg basis. HIR optimization reasons
+algebraically about the resulting Pauli operations. The symbolic-coordinate
+planner then performs all stabilizer-coordinate changes and symbolic dependency
+discovery before emitting `SamplingPlan`.
+
+`SamplingPlan` is the executor-independent semantic boundary. It describes
```
bachase, line 217: This is fairly jargony for someone just learning this theory. Maybe this hsould more distinguish the purpose of these instead of the descriptions?

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+
+Target executable lowering prepares expressions, descriptors, fusion products,
+and kernels for a particular executor. The executor owns the mutable per-shot
+coefficient, scratch, symbol, record, output, and RNG state. Ordinary dispatch
+uses only prepared actions: it performs no tableau evolution, commutation
+analysis, Pauli localization, coordinate selection, or dependency discovery.
+
+## Continuations and Noncomputational Trajectories
+
+A noncomputational transition can invalidate the remaining compiled operations
```
bachase: this might be the first itme someone reads "noncomputational transition", we should have some brief intro of what this means and the bounadry concept

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+[`clifft.basis_probabilities()`](basis_probabilities.md) computes selected
+full-register probabilities without expanding the full $2^n$ statevector. Its
+exponential component still scales with the active width.
+
+For eligible programs with measurements, `clifft.record_probabilities()`
+evaluates selected measurement records exactly without sampling.
+
+## References
+
+- Bradley A. Chase and Farrokh Labib, "Clifft: Fast Exact Simulation of
```
bachase, line 279: Why does one have a DOI And the other doesn't?

docs/theory/overview.md
```diff
@@ -1,107 +1,226 @@
...
+
+## Exact Final-State Queries
+
+For eligible pure-state plans, Clifft retains the final Clifford coordinate map
+needed to relate the active coefficient state back to physical qubits. This
+metadata is not read by ordinary sampling dispatch. It supports exact queries
+such as dense statevector expansion and sparse computational-basis
+probabilities.
+
+[`clifft.basis_probabilities()`](basis_probabilities.md) computes selected
```
bachase, line 265: Why not keep the link to the strong simulation tutorial like we had beofre?

docs/theory/overview.md
bachase: Stepping back - I like the more user experienc oriented descriptions of the prior legacy version. There's enough info to understand what each stage does, but the story is more about "what happens on the path to my sampling a circuit"?

docs/theory/overview.md
bachase: Removed the SymFT DOI so both references now use the same arXiv citation style.

docs/theory/overview.md
bachase: Removed the named-parity detail from this overview. The list now stays with stochastic sources and links instrument outcomes to the noncomputational theory page.

docs/theory/overview.md
bachase: Reworked the full section around the reader journey from circuit text to reused executable plan and results. Each stage now explains its purpose in plain language, while the representation boundary details remain available underneath.

docs/theory/overview.md
bachase: Restored the vertical walkthrough. Bracketed lines are stored representations; labeled arrows are the transformations, so the two roles are visually distinct.

docs/theory/overview.md
bachase: Rephrased this in basis-oriented terms: a coordinate is not a physical qubit, and Clifford gates can make it represent a different multi-qubit Pauli without changing the active-array size.

docs/theory/overview.md
bachase: Introduced noncomputational transitions with leakage/loss first, then defined a boundary as a prepared stopping point where the suffix can be rewritten for one trajectory.

docs/theory/overview.md
bachase: Simplified the close around this result: a shot evaluates prepared expressions and realized signs instead of carrying and updating an n-qubit Pauli frame.

docs/theory/overview.md
bachase: Added a concrete example column covering an active Z rotation, dormant X-axis rotation, active and dormant measurements, and classical outputs.

docs/theory/overview.md
bachase: Made the basis explicit: dormant coordinates are |0> with Z stabilizers, promotion moves one into the dense state, and planned measurement/collapse can return one to the dormant set.

docs/theory/overview.md
bachase: Pulled the SymFT strategy and its practical consequence to the start of the section, before the algebra. Symbolic frame decisions are now consistently described as planning; compilation remains the umbrella pipeline.

docs/theory/overview.md
bachase: Changed this to "after planned step j". The continuation boundary concept is now introduced only in the later section where it is needed.

docs/theory/overview.md
bachase: Restored the Strong Simulation with Exact Probabilities tutorial link alongside the basis-probabilities reference.

bachase:
Pausing this PR as an intentional sequencing choice. We will complete the symbolic WASM/playground cutover and remove the legacy SVM first, then rebase and finish this theory update as a separate PR in the final #301/#303 documentation and release-polish series. The branch and addressed review work are being preserved.

PR #316, opened 2026-08-12
refactor(sampling): extract executable plan builder

bachase:
## Summary

Extract executable-plan construction from the `ExecutablePlan` constructor into a focused `ExecutablePlanBuilder`.

The builder owns construction-only state and makes the lowering sequence explicit:

1. validate the `SamplingPlan` and initialize program storage;
2. prepare noise sites and continuation boundaries;
3. lower typed actions, fusion, and kernel selections;
4. transpose affine terms into the symbol dependency CSR;
5. validate completed expression and action references in checked builds.

This is the first behavior-neutral implementation slice from discussion #298 and closes #315. It deliberately preserves the existing `Action` variant, descriptors, fusion policy, runtime ISA selection, target preparation, RNG behavior, and continuation semantics. It does not add bytecode, a pass manager, instrument specialization, or storage pools.

The previous expression-term reserve estimate remains because removing it regressed preparation on expression-heavy QEC plans. Completed-program validation uses assertions, consistent with compiler-produced internal invariants, so Release builds do not pay for the extra scan.

## Preservation evidence

Release build: GCC 13.3, `-O3 -DNDEBUG -ffast-math -march=x86-64-v2`, AMD EPYC 9554P. The final target-QEC result is the median of seven alternating baseline/candidate rounds with 2,000 compiles per round. The larger-circuit screening uses three alternating rounds with 500 cultivation or 200 QV compiles per round.

| Circuit | Exact-main prepare | Builder prepare | Change |
| --- | ---: | ---: | ---: |
| target-QEC | 0.115345 ms | 0.113993 ms | -1.2% |
| cultivation d5 | 2.054680 ms | 1.974640 ms | -3.9% |
| QV-20 seed 42 | 12.509400 ms | 12.742700 ms | +1.9% |

The review follow-up force-inlines the named construction stages into the single `build()` entry in optimized builds. Release disassembly contains no out-of-line calls to expression preparation, capacity checking, or per-action lowering helpers, restoring the old constructor's code shape while retaining the staged source organization.

Program characterization is unchanged:

- target-QEC: 91 planned to 91 executable actions, 2,061 symbols;
- cultivation d5: 503 planned to 503 executable actions, 16,224 symbols;
- QV-20: 3,723 planned to 558 executable actions, 20 symbols;
- `ExecutablePlan` data members and all action/descriptor types are unchanged, so the existing 72-, 88-, and 104-byte layout assertions remain intact;
[...]

PR #319, opened 2026-08-12
refactor(sampling): specialize instrument actions

### LINKED ISSUE #318: Specialize symbolic executable instrument actions
Follow-up to discussion #298 and tracker #247.

After the `ExecutablePlanBuilder` extraction, lower each planned instrument mode into a concrete prepared execution form instead of retaining `InstrumentMode` plus optional fields in the hot executor. Keep the forms behind the existing outer instrument action so ordinary action dispatch does not grow.

Acceptance:
- preserve instrument, trap/resume, replay-rejection, and deterministic RNG semantics;
- keep hot execution allocation- and exception-free;
- retain the new-X specialization and generic activation fallback;
- measure instrument-heavy execution, ordinary-sampling preservation, compile time, and descriptor storage;
- add differential and structural coverage for the concrete lowering.

This is a bounded typed-action refactor, not a bytecode/pass-manager expansion.

bachase:
## Summary

- lower planned instrument modes into five concrete prepared execution forms
- retain one outer `ExecuteInstrument` action so the common action visitor does not grow
- remove hot mode and optional-field checks while preserving classical, dormant-trap, active, generic-activation, and new-X behavior
- add a generic activation regression test alongside the existing new-X structural coverage

Closes #318. Follows discussion #298 and tracker #247.

## History

This PR was developed as the direct child of #316. It is now rebased onto that PR's squash merge on `main`, and its diff contains one focused commit.

## Design

The first prototype put all five forms in the global action variant. Clean measurement showed that broadening the global visitor affected unrelated ordinary dispatch. The final design instead nests the concrete forms behind the existing instrument alternative. That keeps the global visitor vocabulary unchanged while still making instrument execution type-directed.

The dormant form remains the first nested alternative because GCC queries nested variant default constructibility before `ExecutablePlan` is complete.

On GCC 13 with libstdc++, the largest concrete descriptor is 88 bytes, the nested instrument variant is 96 bytes, and the global action shrinks from 112 to 104 bytes.

## Measurement

Same x86-64-v2 Release build, pinned AMD EPYC 9554P:

- generated repetition-code d17 r5, low-leakage noncomputational model, 1000 shots: median 0.428851 s -> 0.417111 s, 2.74% faster
- target-QEC ordinary survivor sampling, 2,000,000 shots: 12,174,163,579 -> 12,174,162,920 instructions and 1,914,439,953 -> 1,914,439,717 branches
- coherent d3 r3 ordinary survivor sampling, 300,000 shots: 11,374,793,240 -> 11,374,791,794 instructions and 870,360,723 -> 870,360,712 branches
- full record/detector/observable/status/herald hash over 2000 noncomputational shots with seed 281: identical (`10590764395167317276`)

The richer typed lowering has a small compile tradeoff on narrow plans: isolated prepare medians rose about 4-5% on target-QEC and cultivation d5, corresponding to about 5 us and 100 us respectively, or roughly 2% end-to-end. QV-20 prepare and total time remained neutral within the alternating-run noise.

## Validation

- `clifft_tests ~[bench]`: 1,280 cases, 575,601 assertions
- same full non-benchmark suite under forced scalar, AVX2, and AVX-512 dispatch
- GCC 13 and Clang 18 no-runtime-dispatch compilation of the changed production translation units

### DIFF
--- a/src/clifft/sampling/executable_plan.cc
+++ b/src/clifft/sampling/executable_plan.cc
@@ -29,7 +29,8 @@ size_t ExecutablePlan::num_new_x_instrument_activations() const {
     return static_cast<size_t>(
         std::count_if(actions_.begin(), actions_.end(), [](const Action& action) {
             const auto* instrument = std::get_if<ExecuteInstrument>(&action);
-            return instrument != nullptr && instrument->activates_new_x();
+            return instrument != nullptr &&
+                   std::holds_alternative<ExecuteNewXInstrumentActivation>(instrument->action);
         }));
 }
 
--- a/src/clifft/sampling/executable_plan.cc
+++ b/src/clifft/sampling/executable_plan.cc
@@ -30,7 +30,7 @@ size_t ExecutablePlan::num_new_x_instrument_activations() const {
         std::count_if(actions_.begin(), actions_.end(), [](const Action& action) {
             const auto* instrument = std::get_if<ExecuteInstrument>(&action);
             return instrument != nullptr &&
-                   std::holds_alternative<ExecuteNewXInstrumentActivation>(instrument->action);
+                   std::holds_alternative<ExecuteNewXInstrumentActivation>(instrument->form);
         }));
 }
 
--- a/src/clifft/sampling/executable_plan.h
+++ b/src/clifft/sampling/executable_plan.h
@@ -188,6 +188,10 @@ class ExecutablePlan {
         uint32_t destination_flip = 0;
     };
 
+    // These forms intentionally store the same prepared operands but remain
+    // distinct so dispatch encodes whether a clean coordinate must be added.
+    // The current planner selects new-X activation, while validated plans and
+    // future planners may still require the generic measured-source fallback.
     struct ExecuteMeasuredInstrumentActivation {
         PreparedMeasurement measurement;
         PreparedExpression sign;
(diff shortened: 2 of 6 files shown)


bachase:
Addressed the review in b220f85:

- restored a portable compact-descriptor guard (`sizeof(ExecuteInstrument) <= 96`);
- documented why the field-identical active and measured-activation forms remain distinct, including the generic fallback;
- replaced the lowering chain with an exhaustive mode switch plus a Release-safe failure and regression test for an unrecognized mode;
- renamed the nested field from `action` to `form`.

I tested rather than accepted the inlining recommendation. On the pinned low-leakage d17 r5 workload, forcing `finish_instrument_fire`, `trap_instrument`, and the quantum template inline changed the seven-block 3000-shot median from 1.24166 s to 1.26014 s (1.49% slower). Forcing only the template was neutral in seven 5000-shot paired blocks: the median paired ratio was +0.02%, with inconsistent block effects from -3.2% to +3.7%. The current compiler-selected outlining is therefore retained.

I also tested the distribution-lookup helper. As a normal helper GCC emitted a new call on each instrument path; force-inlining infrastructure solely to deduplicate the existing three-line checked lookup did not improve the design, so that small repetition remains.

Validation: 1,280-case full non-benchmark Release suite before the additional regression, 141 focused instrument/trajectory/continuation cases under forced scalar, AVX2, and AVX-512, the new lowering regression, GCC 13 and Clang 18 no-dispatch compiles, and pre-commit.

PR #324, opened 2026-08-13
refactor(sampling): clarify pipeline boundaries

bachase:
## Summary

- document the symbolic sampling path from optimized HIR through planning, CPU lowering, per-shot execution, and result collection
- organize executable descriptors and storage by responsibility, and name the single action-stream lowering stage explicitly
- clarify continuation-plan ownership and the portable/ISA-specific kernel boundary
- keep measurement-branch classification private to the executor instead of exposing it solely for a direct unit test

This is a behavior-neutral navigation and terminology cleanup following the symbolic-backend cutover. It deliberately does not move directories, merge the plan builder back into another translation unit, alter `SamplingPlan` validation, change executable storage, or redesign SIMD dispatch.

## Verification

- `cmake --build build -j4`
- `ctest --test-dir build -E '^Bench:' -j4 --output-on-failure` — 1,283/1,283 passed

The removed direct classifier test covered a test-only interface. Executor-level tests retain end-to-end coverage of both deterministic branches, numerical-dust clamping, replay reachability, and RNG consumption, including asymmetric dust on either branch.

## Performance

Five alternating rounds compared this branch with exact baseline `30114fd` in separate GCC 13.3 RelWithDebInfo/native builds:

| Circuit | Plan | Prepare | Total |
| --- | ---: | ---: | ---: |
| target-QEC | -1.02% | -1.02% | -0.95% |
| cultivation d5 | -0.90% | -2.02% | -0.02% |
| QV20 seed 42 | +0.01% | +0.15% | -0.26% |

The changes are neutral within run-to-run noise. GCC also inlined the now-private measurement classifier, reducing the generated sizes of `sample_active_branch` from 261 to 215 bytes and `force_active_branch` from 215 to 189 bytes.

Related to #247 and discussion #298.

### DIFF
--- a/src/clifft/sampling/executable_plan.h
+++ b/src/clifft/sampling/executable_plan.h
@@ -60,8 +60,8 @@ class ExecutablePlan {
     friend class Executor;
     friend class ExecutablePlanBuilder;
 
-    // Compact handles embedded in actions. Registers refer to the prepared
-    // affine-expression storage owned near the end of this class.
+    // To keep actions compact, register_id refers to expression details in the
+    // storage near the end of this class.
     struct PreparedExpression {
         uint32_t register_id = 0;
     };
@@ -134,7 +134,7 @@ class ExecutablePlan {
     static_assert(sizeof(ExecuteActiveMeasurement) == 88,
                   "active measurement dispatch must not expand its action descriptor");
 
-    // Classical output and symbol actions.
+    // Measurement records and derived symbols.
     struct ExecuteDormantMeasurement {
         PreparedExpression correction;
         uint32_t branch = 0;
@@ -151,6 +151,7 @@ class ExecutablePlan {
         uint32_t symbol = 0;
     };
 
+    // Stochastic readout and derived circuit outputs.
     struct ExecuteReadoutNoise {
         PreparedExpression source;
         uint32_t flip = 0;
@@ -171,14 +172,15 @@ class ExecutablePlan {
         uint32_t observable = 0;
     };
 
+    // Exact expectation-value probe of the current state.
     struct ExecuteExpectation {
         std::optional<PreparedPauli> active_projection;
         PreparedExpression sign;
         uint32_t exp_val = 0;
     };
 
-    // Instrument modes become distinct concrete forms so the hot executor does
-    // not branch over semantic optionals that planning has already resolved.
+    // Each instrument mode has a separate action type so sampling does not
+    // check the mode on every shot.
     struct ExecuteClassicalInstrument {
         PreparedExpression sign;
         uint32_t site = 0;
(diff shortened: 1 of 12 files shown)


src/clifft/sampling/executable_plan.h
```diff
@@ -60,10 +60,14 @@ class ExecutablePlan {
     friend class Executor;
     friend class ExecutablePlanBuilder;
 
+    // Compact handles embedded in actions. Registers refer to the prepared
```
bachase: This is too jargony -- Just say something like "To keep this compact, the register_id references the expression details in the storage at the end of this class"

src/clifft/sampling/executable_plan.h
```diff
@@ -172,6 +177,8 @@ class ExecutablePlan {
         uint32_t exp_val = 0;
     };
 
+    // Instrument modes become distinct concrete forms so the hot executor does
```
bachase: Too jargony -- "Instruments have a few different execution modes. Since its a specialized use case, we gather them together to avoid extra branching at execution/sample time"

src/clifft/sampling/kernels.h
```diff
@@ -8,6 +8,11 @@
 
 namespace clifft::sampling {
 
+// This file defines the architecture-neutral kernel descriptors and scalar
+// operations. Shape-specific *_dispatch files select an implementation once
+// during executable lowering; explicit x86 implementations live in the
+// separately compiled kernels_avx2.cc and kernels_avx512.cc translation units.
```
bachase: This sitll is really confusing to me. To understand how a single operation is handled -- say active measurements, I have to trace through this file, potentnially the three kernels.cc files and the corresponding dispatch /simd files. Isn't that excessive?

src/clifft/sampling/plan.h
```diff
@@ -33,9 +33,19 @@
 
 namespace clifft::sampling {
 
-// A symbol is a plan-local Boolean value assigned once per shot. It represents
-// a presampled stochastic event, a sampled measurement branch, or a named
-// parity derived from earlier symbols.
+// Symbols are plan-local Boolean slots with one value per shot. They carry
```
bachase: How about "A symbol is a plan-local boolean variable that corresponds to a stochastic outcome in the circuit."

src/clifft/sampling/executable_plan.h
bachase: Addressed in 2eb48f1. The comment now says each instrument mode has a separate action type so sampling does not check the mode on every shot.

src/clifft/sampling/kernels.h
bachase: Agreed: the comment explained the fragmentation without fixing it. I shortened it here and will keep the mechanical consolidation separate: one kernel_dispatch pair and one private SIMD declarations header, while retaining the scalar and separately compiled AVX2/AVX-512 implementations.

src/clifft/sampling/executable_plan.h
bachase: Addressed in 2eb48f1 with the simpler wording: the compact register id refers to expression details stored later in the class.

src/clifft/sampling/plan.h
bachase: Addressed in 2eb48f1, with one qualification for derived values: a symbol represents a sampled circuit outcome or a parity derived from earlier symbols.

PR #325, opened 2026-08-13
refactor(sampling): specialize executor kernel backends

bachase:
## Summary

- resolve one `ExecutorBackend` (`Scalar`, `Avx2`, or `Avx512`) while lowering an executable plan
- instantiate the hot action loop for each backend and select it once at the shot/resume boundary
- keep executable action tags architecture-neutral: they describe scalar, diagonal, lane-paired, or high-pivot traversal rather than an ISA
- call the matching separately compiled SIMD entry points directly from each specialized loop
- retain the fused-rotation function pointer only where it couples type-erased sidecar storage to its prepared kernel
- consolidate the former operation-specific dispatch files into `kernel_dispatch.{h,cc}`

The plan and executor carry the same backend chosen from the process ISA, and continuation resume rejects a plan prepared for a different backend. The normal builder path therefore cannot prepare tags for one ISA and execute them through another. Scalar, AVX2, and AVX-512 share one source implementation of classical execution and continuation logic; only the action loop is instantiated three ways in the portable translation unit. SIMD intrinsics remain isolated in explicitly flagged translation units.

This removes the ambient process-ISA lookup from ordinary action dispatch and reduces twelve dispatch/SIMD plumbing files to two. It also avoids compiling the common executor under AVX flags, so architecture-specific instructions cannot leak into the portable fallback.

## Verification

- full 1,290-test non-benchmark suite under each of `CLIFFT_FORCE_ISA=scalar`, `avx2`, and `avx512`
- compiled `executor.cc`, `executable_plan.cc`, `executable_plan_builder.cc`, and `kernel_dispatch.cc` without `CLIFFT_ENABLE_RUNTIME_DISPATCH`; all objects have zero AVX symbol references
- compiled the generic dispatch TUs for `x86-64-v2`; disassembly contains no AVX2 or AVX-512 vector registers
- confirmed `ExecutablePlan`, `Executor`, and the guarded action descriptor sizes are unchanged
- matched workload checksums between the previous PR head and this revision

## Performance

GCC 13.3 RelWithDebInfo/native builds compared this revision with the previous PR head `a0249e6` in five balanced blocks on the AVX-512 performance VM. Times are medians per shot after a warmup; QV-20's much larger host variance is reflected by reporting only the median comparison.

| Circuit | Previous PR head | Specialized executor | Change |
| --- | ---: | ---: | ---: |
| Empty-plan dispatch control | 13.72 ns | 13.56 ns | -1.2% |
| target-QEC | 0.637 us | 0.649 us | +1.9% |
| coherent d5 r1 | 68.43 us | 67.69 us | -1.1% |
| QV-10 | 72.05 us | 71.25 us | -1.1% |
| QV-20 seed 42 | 310.32 ms | 310.51 ms | +0.1% |

Preparation medians were also neutral or slightly better:

| Circuit | Previous PR head | Specialized executor | Change |
| --- | ---: | ---: | ---: |
| target-QEC | 0.1167 ms | 0.1139 ms | -2.4% |
| cultivation d5 | 2.0535 ms | 2.0030 ms | -2.5% |
| QV-20 seed 42 | 9.9225 ms | 9.8744 ms | -0.5% |

The three specialized action loops add 20,480 bytes to the stripped profiling executable (+2.3%). They do not expand plans, executors, or per-action storage. The runtime and preparation changes are within host noise, with no material regression.

Related to #247 and discussion #298.


src/clifft/sampling/kernel_dispatch.h
```diff
@@ -0,0 +1,99 @@
...
+#include <cstdint>
+#include <memory>
+
+namespace clifft::internal {
+enum class RuntimeIsa;
+}
+
+namespace clifft::sampling {
+
+// Compact, architecture-neutral choices stored in executable actions. Lowering
```
bachase: THis is a bit jargon-ful still. Can it be clearer " These functions encapsulate architecture-neutral choices of how to execute actions, such as when the update to the complex state vector is a strided offset that could be paired, or has a high pivot <insert>. When the executor dispatches during sampling, it can then direct these to the relevant architecture specific implementation (scalar/AVX2/AVX512) that is available on the machine."

src/clifft/sampling/simd_kernels.h
bachase, line 1: Why can't these be declared in kernel_dispatch as well? We need to compile them into different architecture specific translation units? If so, document accordingly.

src/clifft/sampling/kernel_dispatch.h
bachase: Updated in 6110071 and superseded more broadly by 124ee79: the comment now explains that the plan records one backend while action tags describe operation shapes such as diagonal rotations or within-vector amplitude pairs. The specialized executor then calls the matching scalar/AVX2/AVX-512 entry point; tags no longer encode or rediscover the ISA.

src/clifft/sampling/simd_kernels.h
bachase: Done in 124ee79. simd_kernels.h is removed and the ISA-specific entry points are declared in kernel_dispatch.h, with a comment explaining that their definitions remain in separately compiled, explicitly flagged translation units.

PR #326, opened 2026-08-14
refactor(sampling): clarify executor modes and dependencies

bachase:
## Summary

- replace three Boolean executor template parameters with four named shot modes
- remove mode templates from action handlers whose behavior does not vary by shot policy
- encapsulate the expression dependency CSR behind a typed construction and lookup interface
- retain compile-time specialization, compact plan storage, and allocation-free hot execution

## Motivation

The previous Boolean template arguments allowed nonsensical combinations to be represented and made call sites such as `execute_actions<false, true, false>` difficult to review. The dependency offsets and targets likewise relied on parallel-vector correspondence at every construction and consumption site.

This keeps the same architecture and execution policies while making both invariants structural.

## Performance

Measured from the same GCC 13 RelWithDebInfo build with `-O2 -ffast-math -march=native -mtune=native`, using alternating processes pinned to an idle EPYC core.

- target-QEC prepare median: approximately 0.114 ms on both revisions; observed difference remained within about 1%
- cultivation-d5 and QV-20 preparation: neutral within run noise
- ordinary, survivor, fixed-k, presampled-value, and replay checksums matched
- execution showed no systematic regression; the expression-heavy presampled control improved after the compiler reduced the CSR range walk to a direct pointer loop
- the action descriptors and `ExecutablePlan` storage layout are unchanged

The dependency factory is force-inlined because this campaign previously measured fixed call overhead on small-plan preparation.

## Verification

- Debug full suite under `CLIFFT_FORCE_ISA=scalar`: 1,284 cases, 416,765 assertions
- Debug full suite under `CLIFFT_FORCE_ISA=avx2`: 1,284 cases, 479,221 assertions
- Debug full suite under `CLIFFT_FORCE_ISA=avx512`: 1,284 cases, 575,621 assertions
- Release non-benchmark suite: 1,284 cases passed

Main currently has an unrelated cross-platform build failure with a separate fix in progress; this PR deliberately does not overlap that work.

### DIFF
--- a/src/clifft/sampling/executable_plan.cc
+++ b/src/clifft/sampling/executable_plan.cc
@@ -3,6 +3,7 @@
 #include "clifft/sampling/executable_plan_builder.h"
 
 #include <algorithm>
+#include <cassert>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -43,6 +44,25 @@ ExecutablePlan::ExecutablePlan(const SamplingPlan& plan)
       instrument_distributions_(plan.instrument_distributions) {
     // Keep construction-only lowering state out of the immutable executable.
     ExecutablePlanBuilder::build(*this, plan);
+}
+
+void ExecutablePlan::ExpressionDependencies::validate(uint32_t num_symbols,
+                                                      size_t num_registers) const noexcept {
+#ifndef NDEBUG
+    assert(offsets_.size() == static_cast<size_t>(num_symbols) + 1 &&
+           "expression dependency offsets have the wrong size");
+    assert(!offsets_.empty() && offsets_.front() == 0 && offsets_.back() == targets_.size() &&
+           "expression dependency ranges are inconsistent");
+    for (size_t i = 1; i < offsets_.size(); ++i) {
+        assert(offsets_[i] >= offsets_[i - 1] && "expression dependency offsets are not ordered");
+    }
+    for (uint32_t target : targets_) {
+        assert(target < num_registers && "expression dependency target is out of range");
+    }
+#else
+    static_cast<void>(num_symbols);
+    static_cast<void>(num_registers);
+#endif
 }
 
 size_t ExecutablePlan::num_new_x_instrument_activations() const {
--- a/src/clifft/sampling/executable_plan.h
+++ b/src/clifft/sampling/executable_plan.h
@@ -89,10 +89,6 @@ class ExecutablePlan {
     // Construction transposes expression terms into one contiguous range per symbol.
     class ExpressionDependencies {
       public:
-        [[nodiscard]] static ExpressionDependencies build(
-            uint32_t num_symbols, std::span<const uint32_t> expression_terms,
-            std::span<const uint32_t> expression_term_begins);
-
         [[nodiscard]] std::span<const uint32_t> dependent_registers(
             uint32_t symbol) const noexcept {
             assert(static_cast<size_t>(symbol) + 1 < offsets_.size() &&
@@ -102,9 +98,14 @@ class ExecutablePlan {
             return std::span<const uint32_t>(targets_).subspan(begin, end - begin);
         }
 
+      private:
+        friend class ExecutablePlanBuilder;
+
+        [[nodiscard]] static ExpressionDependencies build(
+            uint32_t num_symbols, std::span<const uint32_t> expression_terms,
+            std::span<const uint32_t> expression_term_begins);
         void validate(uint32_t num_symbols, size_t num_registers) const noexcept;
 
-      private:
         std::vector<uint32_t> offsets_;
         std::vector<uint32_t> targets_;
     };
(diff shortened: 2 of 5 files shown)


bachase (commented):
**Verdict: approve — no blockers.** The mode collapse is behavior-preserving, and I reproduced the verification locally.

## What I checked

The named modes map 1:1 onto the old Boolean combinations at every call site:

| call site | old `<ForceRecords, SampleNoise, ForceFaults>` | new |
| --- | --- | --- |
| `run_shot()` / `resume()` | `<f,t,f>` | `SampleNoise` |
| `run_shot(presampled)` | `<f,f,f>` | `UsePresampledNoise` |
| `run_shot(KFaultSampler&)` | `<f,f,t>` | `FixedFaultCount` |
| `replay_shot` | `<t,f,f>` | `ReplayRecords` |

`ForceRecords` is equivalent to `Mode == ReplayRecords` at all five guard sites, `ForceFaults` to `FixedFaultCount` in the readout handler, and `SampleNoise` to `SampleNoise` at the boundary. The handlers that lost a template parameter — `ExecuteRotation`'s `ForceRecords` and the seven in the dispatcher's `else` arm — never read it. Explicit-argument overload resolution in the dispatcher is unambiguous because a scoped-enum argument of the wrong type removes the other overloads from the set.

Verification on my side (GCC, Debug, `-j8`): 1,284/1,284 tests pass under `CLIFFT_FORCE_ISA=scalar`, `avx2`, and `avx512`; build is warning-clean.

Two side effects worth keeping: the mode-independent handlers are now one function each instead of four instantiations, and the `else` arm makes a newly added `Action` alternative that needs `Mode` or `Backend` a compile error rather than a silent dispatch to a mode-blind handler.

## Findings

**F1 (minor, linkage): `ExpressionDependencies::build` is declared publicly but only linkable inside the builder TU.** `executable_plan.h:92` declares it non-inline; `executable_plan_builder.cc:40` defines it `inline __attribute__((always_inline))`. `nm` on the object confirms no symbol is emitted for it, while `validate` is a strong `T`. A call from any other translation unit — `executable_plan.cc`, a friended test, a future device builder — therefore fails with an undefined reference instead of a readable error. This matches the convention already used in this file, but those cases are private members of `ExecutablePlanBuilder` declared in the builder's own header, whereas this declaration hangs off `ExecutablePlan` in a widely included one. A one-line comment on the declaration noting that the definition stays in the builder TU for inlining would cover it.

**F2 (minor, robustness): one positional invariant survives the new interface, unasserted.** `executable_plan_builder.cc:55-65` takes two independent spans and derives each expression's extent from the next `expression_term_begins` entry. If that array were ever non-monotonic, `i < end` is false on the first iteration and the expression silently receives zero dependencies — wrong registers with no assertion. Since the goal here is making invariants structural, an `assert(std::ranges::is_sorted(expression_term_begins))` plus a `front() == 0` check inside `build` closes the gap the widened signature opens.

**F3 (nit, simplification): four dispatcher arms do the same thing.** At `executor.cc:697-711`, `ExecuteBoundary`, `ExecuteReadoutNoise`, `ExecuteDormantMeasurement`, and `ExecuteClassicalRecord` all call `execute_action<Mode>(...)`. Merging them leaves a four-arm chain — Mode only, Backend only, Backend and Mode, neither — which reads as the actual taxonomy.

**F4 (nit, organization): `validate` has no inlining requirement.** Unlike `build`, `ExecutablePlan::ExpressionDependencies::validate` (`executable_plan_builder.cc:410`) does not need to live in the builder TU and would sit more naturally beside the other `ExecutablePlan` members in `executable_plan.cc`. Relatedly, `expression_register_constants_` is indexed by the same register id but stayed outside the new class, which is why `validate` takes `num_registers` as a parameter and why `validate_executable_plan` still cross-checks its size against `expression_term_begins_`. Folding the constants into `ExpressionDependencies` would finish the encapsulation; not needed for this PR.

bachase:
Thanks for the careful review. Addressed in a2cd809:

- F1: made `ExpressionDependencies::build` private and explicitly friended only `ExecutablePlanBuilder`, so unintended callers now fail at compile time instead of at link time.
- F2: added Debug checks that the temporary expression ranges start at zero, are ordered, and remain within the term tape.
- F3: collapsed the four `ShotMode`-only action alternatives into one visitor category.
- F4: moved the persistent dependency validator to `executable_plan.cc`.

I kept expression-register constants separate. They initialize registers by register index, whereas `ExpressionDependencies` is specifically the symbol-to-register reverse index used during propagation; combining them would broaden this small abstraction without removing an executor invariant.

I also rebased the PR onto merged #327. Verification: pre-commit clean; the full 1,284-case Debug suite passes under forced scalar, AVX2, and AVX-512; target-QEC Release prepare medians were 0.1153 ms before versus 0.1146 ms after; and the generated hot executor function sizes are unchanged.

PR #330, opened 2026-08-14
feat(sampling): expose plan inspection provenance

bachase:
## Summary

- expose deterministic per-action inspection for SamplingPlan and the target-specific ExecutablePlan
- optionally retain one-based HIR source lines in compact CSR storage
- preserve semantic plan-action ranges when executable lowering fuses several rotations
- keep provenance disabled for ordinary compilation, with no per-action production storage

This is the enabling slice for the symbolic WASM/playground cutover discussed in #298. The playground integration will follow as a stacked PR; SamplingPlan will be the default view and the target-specific WASM program will be an advanced view.

## Verification

- full Debug non-benchmark suite: 1,288 tests passed
- focused planner, instrument, observable, and fused-rotation provenance coverage
- pre-commit hooks passed
- Release native GCC 13 compile comparison against exact main, three alternating rounds:
  - target-QEC planner median: 0.315 ms -> 0.311 ms; total: 0.739 ms -> 0.727 ms
  - cultivation d5 planner median: 3.648 ms -> 3.692 ms; total: 7.803 ms -> 7.867 ms
  - QV-20 planner median: 2.434 ms -> 2.452 ms; total: 16.451 ms -> 16.274 ms

The small changes vary in both directions and remain within roughly 2%, consistent with a neutral default compile path.


bachase:
Follow-up fixes from review are now in `9b13c50`:

- `RemoveNoisePass` and `DropNonUnitaryPass` preserve a parallel HIR source map while filtering operations.
- Requesting sampling provenance now rejects an HIR whose source map is unavailable instead of silently producing empty provenance.
- Ordinary planning skips source-line work when provenance is disabled.
- Direct rotations include prepared cosine/sine values, masks use consistent hexadecimal formatting, and ranges use half-open notation.
- Tests cover pass preservation, the missing-provenance failure, non-rotation action mappings, and distinct direct-rotation inspection.

I kept `PlanSourceMap` separate from the legacy bytecode `SourceMap`: the former maps semantic plan actions to source lines, while the latter also carries VM-instruction-specific metadata (`active_k`) that is being retired with the SVM.

Verification: 1,291 non-benchmark C++ tests pass; pre-commit is clean. Three alternating 1,000-compile target-QEC runs were neutral within noise (median total 0.724 ms on main vs 0.731 ms on the branch; planner 0.309 vs 0.311 ms). PR #331 was also rebased and its WASM smoke tests plus playground lint/build pass.

src/clifft/sampling/plan.h
```diff
@@ -308,6 +309,24 @@ struct InstrumentDistribution {
     std::array<std::array<double, 2>, 2> p_computational_dest{};
 };
 
+// Optional debug provenance parallel to SamplingPlan::actions. Source lines
+// use the same one-based numbering as HirModule::source_map. CSR storage keeps
```
bachase: This is more telling a story than telling what this is -- make the comment concisely about what this is.

src/clifft/sampling/executable_plan.h
```diff
@@ -43,6 +44,13 @@ class PreparedFusedRotationExecution {
 // here so a shot only reads fixed storage.
 class ExecutablePlan {
   public:
+    struct PlanActionRange {
```
bachase, line 48: Worth a comment on what this is/why we'd be tracking it?

src/clifft/sampling/executable_plan.cc
```diff
@@ -98,4 +160,133 @@ std::vector<double> ExecutablePlan::noise_site_probabilities() const {
...
+        out << "  " << i;
+        if (const auto range = action_plan_range(i)) {
+            out << " plans=[" << range->begin << ',' << range->end << ')';
+        }
+        out << ' ' << inspect_action(i) << '\n';
+    }
+    return out.str();
+}
+
+std::string ExecutablePlan::inspect_action(size_t action) const {
```
bachase: Does it make sense to have this big visit with if constexpre, versus inspect_action(ActionType> ? Liek is there somethign shared here worth keeping?

Are any of the Action types viewable/visible in python anyway, and should this be building from the str/repr representation?

Should we move this to another translation unit for debug/printing like we did for bytecode and HIR?

src/clifft/sampling/plan.h
bachase: Updated in a39f43e to define the type directly: a CSR mapping from each SamplingPlan action to its one-based circuit source lines.

src/clifft/sampling/executable_plan.h
bachase: Added a concise definition in a39f43e: this is the half-open SamplingPlan action range lowered into one executable action. It is retained only when provenance is requested.

src/clifft/sampling/executable_plan.cc
bachase: Agreed on separating the formatting. Commit a39f43e moves inspect, inspect_action, action_plan_range, and their helpers into executable_plan_inspection.cc. The variant is now handled by exhaustive overloaded visitors instead of one if-constexpr chain. These action forms are private internal C++ descriptors and are not exposed as Python objects, so there is no existing repr to reuse. Because clifft_core is static, ordinary consumers that never call inspection no longer link this formatter; profile_compile contains zero ExecutablePlan inspection symbols.

PR #331, opened 2026-08-14
feat(wasm): use symbolic sampling backend in playground

bachase:
## Summary

- replace the playground WASM legacy VM execution path with SamplingPlan -> ExecutablePlan -> the symbolic-coordinate sampler
- show the executor-independent Sampling Plan by default, with an advanced WASM toggle for the prepared scalar executable actions
- retain source provenance through planning and lowering so source, HIR, plan, prepared actions, and the active-width chart remain linked
- expose only HIR optimization passes in the playground
- split HIR formatting from legacy VM introspection and bound large affine expressions in the interactive plan view

This is stacked on #330, which adds the opt-in plan provenance and deterministic inspection API.

The WASM target no longer links or executes the legacy SVM. It temporarily retains the legacy lowerer and bytecode pass sources because the combined pass registry owns both HIR and bytecode factories; that dependency can disappear with the planned SVM deletion.

## Performance

Alternating five-round Node 22.22.1 measurements on an AMD EPYC 9554P. Times include WASM compilation and simulation.

| Circuit | Shots | Legacy median | Symbolic median | Change |
| QV-10 | 100 | 100.321 ms | 27.024 ms | 3.71x faster |
| cultivation d5 | 1,000 | 265.531 ms | 95.156 ms | 2.79x faster |
| synthetic active k=10 | 5,000 | 48.136 ms | 50.408 ms | 1.05x slower |
| target-QEC, active width 0 | 100,000 | 212.672 ms | 399.464 ms | 1.88x slower |

The active-width-zero result is the known scalar symbolic overhead; even at the playground shot cap it remains below 0.4 seconds. Active workloads improve substantially.

Alternating seven-round compile/inspection measurements remained below 45 ms in all tested cases: target-QEC 10.1 ms, cultivation d5 44.0 ms, and QV-10 28.9 ms. Compact plan formatting reduced the cultivation d5 JSON payload from an initial 2.9 MB prototype to about 245 KB.

The WASM artifact changes from 749 KiB to 873 KiB (+16.6%).

## Verification

- 1,288 native non-benchmark tests
- WebAssembly build and Node smoke suite, including symbolic noise, measurement, EXP_VAL, limits, and fused-action provenance
- playground TypeScript build and ESLint
- forced full repository pre-commit suite


bachase:
Addressed the follow-up review in c31c0c7:

- wired executable plan ranges into exact active-width chart navigation, including fused actions and repeated source lines
- prevented programmatic editor refreshes and Plan/WASM toggles from carrying stale cursor selections into the other representation
- renamed the remaining bytecode-PC chart terminology and fixed plan register, attribute, float, and integer tokenization
- disabled source-provenance retention for simulation-only compilation
- changed fused source-line unioning to collect, sort, and deduplicate
- made the HIR pass test extensible and explicitly reject the retired bc configuration key
- named the compact affine-expression limit

I kept the success-result optional invariant unchanged because pipeline errors are checked before either caller dereferences the result. New semantic-action hover documentation remains part of the deferred documentation and playground-polish work. backend.cc remains required by the combined pass registry until legacy deletion.

Verification: 1,288 native tests, WASM smoke suite, playground lint/build, pre-commit, and an alternating cultivation-d5 browser benchmark with no performance regression.

src/clifft/sampling/plan.cc
```diff
@@ -18,6 +18,8 @@ namespace clifft::sampling {
 
 namespace {
 
+constexpr size_t kCompactInspectionExpressionTerms = 8;
```
bachase: Why is all this formatting/output in plan.cc versus in a separate translation unit?

PR #341, opened 2026-08-18
docs: present symbolic sampling as the default path

### LINKED ISSUE #305: Update user guides and Python API documentation for symbolic sampling
Update README, site entry points, quickstart, simulation guides, noncomputational and importance-sampling guides, strong-simulation material, and Python API reference for the symbolic-only path.

After cutover, describe the ordinary path as **Clifft's sampler** rather than a selectable "symbolic backend." Use **symbolic sampler** only when explaining the migration or algorithm. Reserve **backend** for concrete execution targets.

Remove examples based on deleted SVM, bytecode, State, execute, or lower interfaces according to the final public API. Keep tutorial scripts and documented code blocks runnable. Land with the symbolic-only switch.

bachase:
## Summary

- describe `compile()` and `lower()` as producing reusable sampling `Program` objects instead of presenting symbolic sampling as a selectable backend
- remove the unreleased `clifft.experimental` transition module now that its APIs live directly under `clifft`
- align user guides and Python docstrings on Clifft sampler, active stabilizer coordinates, and concrete-target use of backend terminology
- retain the documented `lower()` workflow because it remains a supported, exported, and tested public API

The useful integration, statevector, ISA-boundary, and continuation tests that
formerly exercised the alias module now exercise the production API directly.

The importance-sampling and strong-simulation guides already use the current public APIs; this audit found no migration edits needed there.

## Stack

- based on #340, which renames the obsolete opcode/instruction reference as compiler IR
- review the final commit in this PR for the #305-specific changes

## Verification

- rebuilt the editable Python package and verified `clifft.experimental` is absent
- `uv run --frozen pytest -q tests/python` (790 passed)
- `uv run --frozen pytest --codeblocks docs/ README.md -q` (31 passed, 7 skipped)
- `uv run --frozen --group docs mkdocs build --strict`

Closes #305.


docs/reference/python-api.md
```diff
@@ -1,5 +1,11 @@
 # Python API Reference
 
+`clifft.compile()` returns a reusable `Program` consumed by the sampling and
+exact-query functions below. `clifft.experimental` remains as a compatibility
```
bachase: We should deprecate clifft.experimental and not document it. It was never part of an official release and only aided the transition.

src/python/clifft/experimental.py
bachase, line 1: AS noted -- lets drop this.

docs/reference/python-api.md
bachase: Agreed. Removed the documentation entirely and now state only that the public API has no sampler-selection module or backend switch. The unreleased transition module is deleted in ba0c778.

src/python/clifft/experimental.py
bachase: Dropped in ba0c778. I retained the useful integration, statevector, ISA-boundary, and continuation coverage by moving those tests onto the production clifft and clifft.noncomp APIs.

Issue #329, opened 2026-08-14
ci: strengthen arm64, AVX-512, optimized-build, and sanitizer coverage


bachase:
## Motivation

The symbolic sampling backend is now the Python default (#321), and the remaining migration work in #247 is approaching the point where the new executor will be the only native sampling path. At the same time, #327 showed why platform-specific build coverage must gate changes before they reach `main`, and #328 is moving the existing macOS and Windows jobs onto pull requests.

Once #328 lands, a few focused gaps remain across a shipped architecture, deterministic ISA execution, optimized Python behavior, and memory safety. The goal here is strong feedback without multiplying overlapping build configurations: reuse existing jobs where possible, keep fast deterministic checks on pull requests, and reserve expensive or initially noisy checks for scheduled runs.

The work items below are independently landable.

### 1. Exercise AVX-512 deterministically on pull requests

The current QEMU matrix covers AVX2/scalar fallback, ISA leakage, and clean traps on Haswell and Nehalem, but it never executes the AVX-512 backend. QEMU's TCG emulation stops at AVX2 and masks every AVX-512 CPUID bit, so an emulated Skylake-Server resolves to a clean trap rather than exercising the kernels, and native hosted-runner selection is not a deterministic AVX-512 test.

- Use Intel SDE (`sde64 -skx`) as the deterministic AVX-512 executor, downloaded from a pinned, checksum-verified release the same way CI pins its other tool downloads.
- Extend the existing Release-job wheel smoke with SDE legs that run auto-detect and `CLIFFT_FORCE_ISA=avx512`, both expecting success.
- Expose the resolved runtime ISA to the smoke, and make the smoke print and verify that it is `avx512` before exercising sampling and exact/replay behavior.
- On native runners, run the forced-`avx512` C++ suite when the host CPU supports it, mirroring the existing gated AVX2 leg. This is supplemental coverage that varies with the runner fleet, not a merge gate; the SDE smoke is the deterministic check.
- Keep the deterministic SDE check focused on the smoke on pull requests. Add a broader SDE-emulated test subset only if measurement shows that important AVX-512 kernels remain uncovered; do not introduce a test-tag taxonomy solely for this job.

### 2. Add Linux arm64 to pull-request coverage

`release.yml` publishes a Linux aarch64 wheel, but routine CI does not build or test Linux on arm64.

- Add `ubuntu-24.04-arm` to the existing cross-platform matrix with the generic CPU baseline.
[...]

PR #353, opened 2026-08-18
ci: execute the AVX-512 backend deterministically with Intel SDE

bachase:
## Summary

Implements item 1 of #329: CI never actually executes the ~500-line AVX-512 kernel TU. QEMU cannot fix this — its TCG engine stops at AVX2 and strips every AVX-512 CPUID bit from emulated models (verified on the qemu 8.2.2 the `ubuntu-24.04` image ships: `-cpu Skylake-Server` warns `TCG doesn't support requested feature: CPUID.07H:EBX.avx512f` and resolves to a clean trap). Native runner selection is not deterministic either, since hosted CPU models vary. Intel SDE is the standard deterministic AVX-512 executor, so:

- **`clifft.runtime_isa()`** (new, small API addition): reports the backend runtime dispatch resolved (`scalar` / `avx2` / `avx512`, or `trap:...` for an unavailable `CLIFFT_FORCE_ISA` request), so smoke tests verify the resolved ISA instead of inferring it.
- **`wheel_smoke.sh`** now takes an emulator argument (`qemu` | `sde`) and an optional expected resolved ISA that pass legs must report. The existing QEMU legs keep their semantics and additionally assert their resolved ISA (Haswell→`avx2`, Nehalem→`scalar`).
- **`release-cpp-smoke`** (PRs) and the **Linux x86_64 release wheel job** install a pinned, checksum-verified SDE (10.13.1) and add two Skylake-X legs — auto-detect and forced `avx512` — both expecting success with `isa=avx512` verified.
- **`build-and-test`** runs the full non-benchmark C++ suite with `CLIFFT_FORCE_ISA=avx512` when the host CPU supports it, mirroring the gated AVX2 leg. This is supplemental host-dependent coverage; the SDE smoke is the deterministic gate.

Per the issue, job durations should be recorded after landing; a broader SDE-emulated test subset stays a follow-up only if a measured gap remains.

## Test plan

- [x] C++ tests pass (`ctest --test-dir build -E Bench --output-on-failure`) — 831/831 (Debug, x86-64-v2)
- [x] Python tests pass (`uv run pytest tests/python/ -v`) — 803 passed on a Debug extension, plus a `CLIFFT_FORCE_ISA=scalar` rerun of `test_api.py`

Additional validation on a Release `x86-64-v2` extension (matching the CI smoke configuration) with SDE 10.13.1 on an AMD Zen 4 host:

- Both new CI legs pass with `isa=avx512` verified: `sde skx auto pass avx512`, `sde skx avx512 pass avx512`.
- Cross-ISA checks pass: `sde hsw auto pass avx2`, `sde hsw avx2 pass avx2`, `sde nhm auto pass scalar`.
- Negative controls behave: `sde hsw auto pass avx512` fails on the ISA assertion; `sde hsw avx512 fail` reports the established clean trap (no SIGILL), confirming clifft's dispatch follows the emulated CPUID rather than the host.
- The full non-benchmark C++ suite passes under `sde64 -skx` with `CLIFFT_FORCE_ISA=avx512` (831 cases, ~14 s on this host).
- The QEMU legs could not be re-run locally (qemu-user 8.2 crashes on any CPython process under this host's 6.12 kernel, independent of this change); their invocation is unchanged apart from the new arguments, and CI exercises them on every PR.

## Contributor agreement


bachase:
Observed durations from this PR's run, per the measurement guidance in #329:

- `release-cpp-smoke`: **1m53s** total. The increment is ~15s: SDE download+verify ~2s, each Skylake-X leg ~6.6s (isa=avx512 verified in both, on an AMD runner — i.e. genuinely software-emulated AVX-512).
- `build-and-test`: **2m37s**; this runner lacked avx512f/avx512dq, so the native avx512 leg skipped with the `::notice` as designed.
- The Windows critical path (2m54s) is unaffected.

No move to scheduled coverage warranted at this cost.

PR #354, opened 2026-08-18
feat(sampling): add cross-shot worker parallelism

### LINKED ISSUE #343: Add cross-shot worker parallelism to the sampling entry points
## Goal

Add an opt-in multithreaded sampling mode that distributes shots across worker threads, each running the existing single-shot executor against the shared immutable `ExecutablePlan`.

The symbolic sampling stack is entirely single-threaded today (no threading anywhere in `src/clifft/sampling`, `src/clifft/api`, or the Python bindings). #312 covers intra-shot kernel parallelism and #313 covers single-threaded packed batching; both explicitly set cross-shot worker parallelism aside as a separate concern, so nothing currently tracks it. For end users estimating logical error rates it is the largest and cheapest available multiplier: shots are independent and the plan is immutable, so scaling should be near linear.

This is a future performance enhancement. It does not block anything.

## Proposed direction

- Keep the `ExecutablePlan` shared and immutable; give each worker its own `Executor`, `State`, and output storage, allocated before dispatch per the existing hot-path rules.
- Chunk shots across workers and write records into disjoint preallocated row ranges so shot order is preserved without a merge step.
- Decide the seeded-determinism contract explicitly. Today one `Xoshiro256PlusPlus` is seeded per sampling call and shots consume the stream sequentially (`src/clifft/sampling/sampler.cc`), so threaded output under a fixed seed cannot reproduce the current single-threaded stream. Options: derive an independent per-shot stream from the seed and shot index (the same direction as the per-(shot,domain) scheme the noncomputational RNG already uses), possibly adopting it for single-threaded sampling too as a documented breaking change; or keep the sequential stream for `threads=1` and define threaded determinism separately (deterministic for a fixed seed regardless of thread count requires per-shot or per-chunk streams either way).
- Define survivor-sampling semantics under threading: total/passed accounting and survivor row order must stay deterministic.
- Expose a `threads` parameter on the native and Python entry points. Native-only; the wasm build stays single-threaded.

## Evaluation

- Scaling at 2-32 workers across the migration corpus regimes (pure-Clifford d7/d9, cultivation, distillation, QV preparation).
- Postselection-heavy workloads, where per-shot cost varies strongly across the batch.
- Per-worker memory footprint at large active widths.

## Reference

SymFT exposes `threads=N` chunked worker execution on its prepared samplers, with per-worker reusable execution buffers retained across chunks and sampling calls: https://arxiv.org/abs/2607.28600 (section 6, "Reusing the prepared state"), `cpp/src/sampler/prepared_sampler.{hpp,cpp}` in https://github.com/haoliri0/SOFT.

bachase:
## Summary

- add one shared atomic chunk scheduler with preconstructed worker contexts and join-safe exception propagation
- parallelize `sample`, `sample_survivors`, `sample_k`, and `sample_k_survivors` across independent per-worker executors
- preserve exact seeded rows, survivor order, and metadata across worker counts
- default to `threads=1`; accept positive counts or `threads="auto"` for the implementation-reported hardware concurrency, capped at the shot count
- document per-worker active-state memory scaling

Depends on #352. Closes #343 once the stack reaches `main`.

## Local scaling smoke tests

Release build on a 14-logical-CPU Apple Silicon host, active width 10:

- fixed rows, 2,000 shots: 353k shots/s at 1 worker; 684k at 2; 1.16M at 4; 2.14M at auto
- postselection-heavy, 10,000 attempts: 705k attempts/s at 1 worker; 1.39M at 2; 2.66M at 4; 5.92M at auto
- every worker count produced identical seeded rows and exactly 4,921 survivors in the postselection case

These are implementation smoke tests, not published benchmark claims.

## Validation

- full native non-benchmark suite
- join-and-rethrow synchronization test: 100 consecutive passes under both default and forced-scalar ISA
- `just py-test`: 810 passed, 1 skipped
- focused Python threading tests: 10 passed
- `just docs-build`
- `just py-doctest`: 32 passed, 7 skipped
- `just lint`

Wasm remains serial through the Emscripten compile-time path. The local `just test-wasm` recipe could not start because Docker Desktop was not running; CI exercises that build.

### DIFF
--- a/docs/guide/simulation.md
+++ b/docs/guide/simulation.md
@@ -250,8 +250,10 @@ processes with CPU-affinity limits, set an explicit count if the reported
 hardware concurrency exceeds the available CPU quota.
 
 Workers dynamically claim contiguous shot ranges from a shared scheduler.
-Seeded rows and surviving-row order are identical for `threads=1`, an explicit
-worker count, and `threads="auto"`.
+With a fixed seed, changing `threads` produces exactly the same result.
+`sample()` and `sample_k()` keep each shot at the same row, and the survivor
+APIs return accepted shots in the same order as the corresponding one-thread
+run.
 
 Each worker owns a separate executor. Its dense coefficient and measurement
 scratch storage uses roughly $24 \times 2^k$ bytes at peak active width $k$,
--- a/docs/guide/simulation.md
+++ b/docs/guide/simulation.md
@@ -239,6 +239,25 @@ If `seed` is omitted or set to `None`, Clifft uses hardware entropy from the ope
 Each shot derives an independent random stream from the call seed and its shot
 index. Seeded results therefore do not depend on how shots are scheduled.
 
+## Parallel Shots
+
+`sample()`, `sample_survivors()`, `sample_k()`, and
+`sample_k_survivors()` accept a `threads` argument. It defaults to `1`, so
+existing calls remain serial. Pass a positive worker count to control resource
+use, or `threads="auto"` to use the implementation-reported hardware
+concurrency. Clifft never creates more workers than shots. In containers or
+processes with CPU-affinity limits, set an explicit count if the reported
+hardware concurrency exceeds the available CPU quota.
+
+Workers dynamically claim contiguous shot ranges from a shared scheduler.
+Seeded rows and surviving-row order are identical for `threads=1`, an explicit
+worker count, and `threads="auto"`.
+
+Each worker owns a separate executor. Its dense coefficient and measurement
+scratch storage uses roughly $24 \times 2^k$ bytes at peak active width $k$,
+plus symbolic state, records, and other executor metadata. Set an explicit
+worker count when memory is more constrained than CPU availability.
+
 ## Importance Sampling (Forced k-Faults)
 
 For circuits where logical errors are rare, standard Monte Carlo can require an impractical number of shots. Clifft provides stratified importance sampling via `sample_k` and `sample_k_survivors`, which force exactly `k` physical faults per shot. Results from different `k` strata must be combined using the corresponding Poisson-binomial probability $P(K = k)$.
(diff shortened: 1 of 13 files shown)


docs/guide/simulation.md
```diff
@@ -239,6 +239,25 @@ If `seed` is omitted or set to `None`, Clifft uses hardware entropy from the ope
...
+`sample()`, `sample_survivors()`, `sample_k()`, and
+`sample_k_survivors()` accept a `threads` argument. It defaults to `1`, so
+existing calls remain serial. Pass a positive worker count to control resource
+use, or `threads="auto"` to use the implementation-reported hardware
+concurrency. Clifft never creates more workers than shots. In containers or
+processes with CPU-affinity limits, set an explicit count if the reported
+hardware concurrency exceeds the available CPU quota.
+
+Workers dynamically claim contiguous shot ranges from a shared scheduler.
+Seeded rows and surviving-row order are identical for `threads=1`, an explicit
```
bachase: What does this "seeded rows and surviving-row order are identical" mean?

src/clifft/util/shot_parallel.h
```diff
@@ -0,0 +1,144 @@
...
+                std::lock_guard lock(error_mutex);
+                if (first_error == nullptr) {
+                    first_error = std::current_exception();
+                }
+            }
+            cancelled.store(true, std::memory_order_relaxed);
+        }
+    };
+
+    // std::thread keeps the documented compiler-library floor. This guard is
```
bachase: This is too jargony (compiler library floor). Just state more clearly that std::jthread is not available for all the minimum compiler library versions we support, so we add this liftetime management helper.

bachase: Reworded this to state directly that `std::jthread` is unavailable on some supported compiler and standard-library versions, so the helper guarantees that every `std::thread` is joined. Updated in dfcdc672.

docs/guide/simulation.md
bachase: Clarified this to say that, with a fixed seed, changing `threads` produces exactly the same result: ordinary sampling preserves each shot row, and survivor sampling preserves accepted-shot order. Updated in dfcdc672.

PR #355, opened 2026-08-19
feat(noncomp): add cross-shot worker parallelism

### LINKED ISSUE #343: Add cross-shot worker parallelism to the sampling entry points

bachase:
## Summary

- reuse the shared dynamic shot-range scheduler for `noncomp.sample()`
- keep each worker's mutable trajectory state private, including its lazy all-ground continuation cache and executor
- preallocate result sidecars and write by global shot index, preserving deterministic row order
- expose `threads=1` by default and `threads="auto"` for the implementation-reported hardware concurrency in Python; C++ uses `threads=0` for auto
- document trajectory-specific load balancing and per-worker memory scaling

Workers claim ranges from the same atomic dispenser introduced in #354. This is useful for noncomputational sampling because different shots can compile and execute different numbers of continuations. The annotated circuit and model remain shared and immutable; continuation planning and executor mutation are worker-local.

The all-ground continuation remains lazy per worker. In particular, a shot population that starts entirely noncomputational does not compile or validate an unreachable all-computational module.

## Determinism

For a fixed seed, `threads=1`, explicit worker counts, and `threads="auto"` produce exactly identical measurements, detectors, observables, final-status sidecars, herald sidecars, and row ordering.

## Performance

Local release-build smoke benchmark on 14 logical CPUs, using exact-output checks before timing:

| Workload | 1 worker | 2 workers | 4 workers | auto |
| All-ground 8-qubit trajectories | 234k shots/s | 452k shots/s | 865k shots/s | 2.10M shots/s |
| Variable continuation-heavy trajectories | 18.5k shots/s | 36.7k shots/s | 62.2k shots/s | 105k shots/s |

## Validation

- full native non-benchmark suite
- 816 Python tests passed, 1 expected skip
- join-and-rethrow synchronization test: 100 consecutive passes under both default and forced-scalar ISA
- strict documentation build
- 32 documentation code blocks passed, 7 skipped
- full pre-commit suite
- [ThreadSanitizer on final tip](https://github.com/unitaryfoundation/clifft/actions/runs/32267609034)

## Stack

Depends on #354, and therefore #352. No issue changes are needed for this final stack layer; #354 already closes #343.

### DIFF
--- a/src/clifft/noncomp/trajectory_driver.cc
+++ b/src/clifft/noncomp/trajectory_driver.cc
@@ -24,9 +24,11 @@
 #include "clifft/optimizer/pass_registry.h"
 #include "clifft/sampling/executor.h"
 #include "clifft/sampling/planner.h"
+#include "clifft/util/shot_parallel.h"
 #include "clifft/util/shot_seed.h"
 #include "clifft/util/xoshiro.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cstdint>
 #include <iterator>
@@ -282,6 +284,13 @@ struct SamplingCompiledContinuation {
     std::vector<AnnotationTarget> site_targets;
     std::vector<QubitStatus> final_status;
     std::optional<size_t> forced_traceout_slot;
+};
+
+struct TrajectoryWorker {
+    // The all-ground path remains lazy because a noncomputational initial
+    // state can make it unreachable, including paths rejected by a width cap.
+    std::optional<SamplingCompiledContinuation> all_ground_start;
+    std::optional<sampling::Executor> all_ground_executor;
 };
 
 // Build the HIR pass pipeline from default passes that preserve measurement
@@ -557,7 +566,8 @@ Circuit prepare_trajectory_circuit(const Circuit& circuit, const NonComputationa
 NonComputationalSample run_trajectory_driver(const Circuit& circuit,
                                              const NonComputationalModel& model, uint32_t shots,
                                              const SeedRoot& root,
-                                             std::optional<uint32_t> max_active_width_cap) {
+                                             std::optional<uint32_t> max_active_width_cap,
+                                             uint32_t threads) {
     NonComputationalSample result;
     result.shots = shots;
     result.num_qubits = circuit.num_qubits;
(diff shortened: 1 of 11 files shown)


src/clifft/noncomp/trajectory_driver.cc
```diff
@@ -284,6 +286,13 @@ struct SamplingCompiledContinuation {
     std::optional<size_t> forced_traceout_slot;
 };
 
+struct TrajectoryWorker {
+    // The all-ground path remains lazy because a noncomputational initial
```
bachase: I'm not sure I like all_ground. Maybe no_jump? Or no transitions?

bachase: Renamed this cache to `shared_start` / `shared_start_executor`. The path is shared by shots whose initial levels are all G, but it may still encounter transitions and jumps later, so `no_jump` or `no_transitions` would be misleading. Updated in 77e3a66e.

PR #357, opened 2026-08-19
ci: run the full Python suite on the optimized extension

bachase:
## Summary

Implements item 3 of #329. `release-cpp-smoke` already built an optimized Python extension, but only ran the export audit and emulated ISA smokes against it — optimized Python coverage stopped there. This keeps `assert()` active via `CLIFFT_FORCE_ASSERTS=ON` and runs the full Python suite against that extension, covering GCC optimized code generation and `-ffast-math` through the public API. The exact Release C++ suite in the same job is untouched (normal `NDEBUG` behavior preserved). The smoke venv now installs test dependencies (drops `--no-dev`), and the QEMU/SDE smokes reuse the asserts-on wheel since ISA dispatch does not depend on `NDEBUG` — the published artifact keeps its own smoke in the release workflow.

## Test plan

Validated locally on a Release + `CLIFFT_FORCE_ASSERTS=ON` + `x86-64-v2` extension (GCC, `-ffast-math`): full Python suite passes 803/803. The new step also validates itself on this PR — watch `release-cpp-smoke` and its duration.

- [x] Pre-commit checks pass (via the commit hook and this PR's lint job)

## Contributor agreement

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -314,16 +314,25 @@ jobs:
           enable-cache: true
           python-version: "3.12"
 
-      - name: Install Python package (release baseline)
+      # assert() stays active in the optimized extension so the full
+      # Python suite checks invariants through GCC optimized codegen and
+      # -ffast-math. ISA dispatch does not depend on NDEBUG, so the
+      # emulated smokes below reuse this wheel; the published artifact
+      # keeps its own smoke in the release workflow.
+      - name: Install Python package (release baseline, asserts on)
         env:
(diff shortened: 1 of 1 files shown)


bachase:
Observed durations from this PR's run, per #329's measurement guidance: `release-cpp-smoke` 3m29s (was ~1m34s). The full Python suite itself costs only **18s**; the growth is the extension build (117s) ccache-missing on the new `CLIFFT_FORCE_ASSERTS` flag variant plus the dev-group install. After the first post-merge main push saves the new-variant cache, the steady-state increment should be roughly the suite + dependency delta (~30s).

PR #372, opened 2026-08-19
perf(frontend): canonicalize Clifford rotations during tracing

### LINKED ISSUE #367: Absorb exact Clifford-valued rotations during tracing
## Problem or motivation

Stim-format circuits can spell Clifford operations as exact arbitrary-angle rotations such as `R_X(0.5)`, `R_Y(1)`, or Clifford-valued components of `U3`. The frontend currently lowers those operations through generic `PHASE_ROTATION` HIR. This enlarges the HIR, makes `PeepholeFusionPass` rediscover Clifford structure through long scans, and can delay localization enough to increase planned active width.

Compilation-only profiles on the monitored brickwork family studied in [Classical Simulations of Low Magic Quantum Dynamics](https://arxiv.org/abs/2508.20252) indicate that this representation choice, rather than a fundamentally expensive simulator kernel, is the primary cause of the observed slowdown.

This work depends on #366, which defines public statevector results up to global phase. That contract permits equivalent Clifford representatives to be absorbed without reconstructing source-matrix global phase.

## Proposed solution

Recognize exact single-qubit Clifford rotations during frontend tracing and absorb them directly into the Clifford tableau before HIR emission.

- Classify an angle only when twice the binary64 value is exactly integral and finite.
- Reduce the resulting quarter-turn count modulo four.
- Absorb exact `R_X`, `R_Y`, and `R_Z` rotations using native Clifford tableau updates.
- Apply the same treatment independently to Clifford-valued `U3` components.
- Do not use a floating-point tolerance: nearby values such as `0.5000000000000001` remain generic rotations.
- Exclude absorbed operations from Pauli-mask capacity estimates.

The circuit generator should emit native Cliffords when that is its intent, but the compiler should normalize equivalent exact spellings efficiently.

## Correctness and architecture constraints

- Perform classification and tableau updates during tracing; do not introduce runtime topology planning.
- Preserve relative amplitudes, probabilities, measurement statistics, source mapping, and deterministic compilation.
- Keep non-Clifford and non-finite angle validation behavior unchanged.
- Do not add a new public pass or a circuit-width-dependent fast path.

## Evaluation and acceptance criteria

- Cover positive and negative half-integer angles across several periods for all three axes.
- Cover exact identity, Pauli, square-root, and inverse-square-root cases.
- Verify nearby non-exact angles still emit `PHASE_ROTATION`.
- Validate small circuits against Qiskit Aer up to global phase.
- Benchmark representative monitored-brickwork points on top of #366 and report compile time, peephole time, HIR size, peak active width, and peak memory.
- Preserve the existing Python and C++ test suites.

## Alternatives considered

Optimizing the generic peephole scanner remains useful for genuinely non-Clifford rotations, but it does not address the unnecessary generic representation. A dedicated Clifford-rotation HIR opcode would still retain a large intermediate stream and would not recover trace-time frame selection or mask-allocation savings.

bachase:
## Summary

- canonicalize finite `R_X`, `R_Y`, and `R_Z` angles within an absolute `1e-12` half-turn tolerance of a Clifford angle during frontend tracing, absorbing the native Clifford directly into the tableau
- use one shared classifier and tolerance in both frontend tracing and HIR peephole canonicalization, eliminating the duplicated local policy
- canonicalize `U3` components independently and exclude absorbed rotations from Pauli-mask capacity estimates
- explicitly document that sub-tolerance deliberate overrotations are treated as their canonical Clifford representative
- cover exact positive and negative periods, inside/outside tolerance boundaries, mixed and all-Clifford `U3`, large finite angles, standalone multi-qubit HIR demotion, and Qiskit Aer statevector equivalence

Closes #367.

## Dependency

This work relies on the projective statevector contract introduced by #369, which permits the frontend to absorb an equivalent Clifford representative without reconstructing the source matrix's scalar phase.

## Compilation snapshot

Compilation-only, one seed at each of the six monitored brickwork points, one worker. The baseline is the parent branch #369; no circuit simulation is included.

| eta | L | T_f/L | compile s, parent -> PR | peak k, parent -> PR | HIR ops, parent -> PR | peak RSS MiB, parent -> PR |
|---:|---:|---:|---:|---:|---:|---:|
| 0.5 | 20 | 8 | 1.176 -> 0.021 | 4 -> 2 | 3,620 -> 3,334 | 79.8 -> 75.9 |
| 0.5 | 20 | 16 | 4.684 -> 0.070 | 4 -> 2 | 7,196 -> 6,604 | 199.3 -> 185.1 |
| 0.5 | 32 | 8 | 7.351 -> 0.110 | 5 -> 3 | 9,124 -> 8,363 | 300.2 -> 275.5 |
| 0.5 | 32 | 16 | 29.459 -> 0.453 | 5 -> 2 | 18,121 -> 16,591 | 1,043.0 -> 955.3 |
| 0.25 | 32 | 8 | 7.523 -> 0.113 | 5 -> 3 | 9,084 -> 8,319 | 297.8 -> 274.2 |
| 1.0 | 32 | 8 | 7.491 -> 0.110 | 5 -> 3 | 9,207 -> 8,459 | 302.4 -> 278.0 |

Across this slice, compile time improves by 55-68x. In a peephole-only ablation, pass time falls from 1.143-29.518 seconds on the parent to 0.034-0.272 milliseconds here. The HIR reduction is modest, but canonicalizing Clifford rotations before HIR prevents them from entering the scanner and lets tracing update the Clifford frame at the source.

## Test plan

- [x] Non-benchmark C++ tests pass: 838 passed
- [x] Python tests pass: 863 passed, 1 skipped
- [x] Qiskit Aer coverage includes 39 exact axis-angle cases and exact/mixed `U3` cases
- [x] Strict documentation build passes
- [x] Documentation examples pass: 32 passed, 7 skipped
- [x] Pre-commit formatting, lint, and type checks pass

## Contributor agreement

- [x] I confirm this contribution is my original work or I have the right to submit it, and I license it under the project Apache-2.0 license.
- [x] I have reviewed the AI-assisted changes and included an `Assisted-by:` git trailer.

### DIFF
--- a/docs/reference/gates.md
+++ b/docs/reference/gates.md
@@ -103,6 +103,14 @@ parameters are in **half-turns** (multiply by pi to get radians).
 | `U3`  | `U3(theta,phi,lambda) target` | General SU(2) gate = `R_Z(phi) R_Y(theta) R_Z(lambda)` |
 | `U`   | `U(theta,phi,lambda) target` | Alias for `U3` |
 
+Finite angles within an absolute tolerance of `1e-12` half-turns of a multiple
+of `0.5`, such as `0.5`, `-1`, and `1.5`, are canonicalized to that Clifford
+rotation during tracing. This policy intentionally absorbs small numerical
+errors introduced by circuit generation or serialization; a deliberate
+overrotation smaller than the tolerance is therefore treated as the canonical
+Clifford. Deviations outside the tolerance remain continuous rotations. Each
+`U3` component receives the same treatment.
+
 !!! note "Name conflicts with Stim"
     Clifft uses `R_X`, `R_Y`, `R_Z` (with underscores) to avoid collision with
     Stim's `RX` / `RY` reset-in-basis instructions.
--- a/src/clifft/frontend/frontend.cc
+++ b/src/clifft/frontend/frontend.cc
@@ -8,6 +8,7 @@
 #include <cmath>
 #include <initializer_list>
 #include <numbers>
+#include <optional>
 #include <random>
 #include <stdexcept>
 #include <string>
@@ -43,6 +44,18 @@ void apply_single_qubit_clifford(stim::TableauSimulator<kStimWidth>& sim, GateTy
             return;
         case GateType::Z:
             sim.inv_state.prepend_Z(q);
+            return;
+        case GateType::SQRT_X:
+            sim.inv_state.prepend_SQRT_X_DAG(q);
+            return;
+        case GateType::SQRT_X_DAG:
+            sim.inv_state.prepend_SQRT_X(q);
+            return;
+        case GateType::SQRT_Y:
+            sim.inv_state.prepend_SQRT_Y_DAG(q);
+            return;
+        case GateType::SQRT_Y_DAG:
+            sim.inv_state.prepend_SQRT_Y(q);
             return;
         default:
             break;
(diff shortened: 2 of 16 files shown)


src/clifft/frontend/frontend.cc
```diff
@@ -44,6 +45,18 @@ void apply_single_qubit_clifford(stim::TableauSimulator<kStimWidth>& sim, GateTy
         case GateType::Z:
             sim.inv_state.prepend_Z(q);
             return;
+        case GateType::SQRT_X:
```
bachase: How come these weren't here previously?

docs/reference/gates.md
```diff
@@ -103,6 +103,14 @@ parameters are in **half-turns** (multiply by pi to get radians).
 | `U3`  | `U3(theta,phi,lambda) target` | General SU(2) gate = `R_Z(phi) R_Y(theta) R_Z(lambda)` |
 | `U`   | `U(theta,phi,lambda) target` | Alias for `U3` |
 
+Finite angles within an absolute tolerance of `1e-12` half-turns of a multiple
+of `0.5`, such as `0.5`, `-1`, and `1.5`, are canonicalized to that Clifford
+rotation during tracing. This policy intentionally absorbs small numerical
```
bachase: This is too jargony -- just keep it simple that we canonicalize up to that absolute tolerance; don't need the later sentences.

PR #384, opened 2026-08-22
feat(hip): add gfx942 sampling backend vertical slice

bachase:
Add an optional gfx942 HIP backend that lowers `SamplingPlan` into a
backend-private packed executable and runs one thread per shot for active width
`k <= 4`. Support FP64 and experimental FP32 coefficient evolution while
keeping reductions, normalization, aggregate statistics, replay likelihoods,
and expectation values in FP64.

Reuse shared Pauli preparation, reject unsupported work during lowering, and
cover forced replay, stochastic CPU-oracle comparisons, noise, postselection,
survivor compaction, observables, and GPU-free ROCm compilation. Hardware
conformance on MI300X remains required before broader use.


tools/ci/audit_hip_fatbin.py
bachase, line 1: Let's not actually add this check to our automation or repo. It was useful as a one-off and did detect an issue with the build, but that was a temporary thing. As we really move this to production, we will be building and testing on actual GPUs in some way that would be the better way to detect.

docs/development/hip-backend.md
```diff
@@ -0,0 +1,92 @@
...
+measurements, affine records and symbols, categorical Pauli noise, asymmetric
+readout noise, detectors and postselection, observables, and expectation-value
+probes. Fixed-row sampling rejects postselected plans and directs callers to
+survivor sampling instead.
+
+The initial backend rejects transition instruments. Leakage and loss,
+importance sampling, `sample_k`, asynchronous execution, and multi-GPU
+execution are also outside this tier.
+
+## Conformance and Next Tiers
```
bachase: Make this easier to read/less jargony (list of features) -- instead be clear (1) what is tested on CPU vs what is tested on GPU, (2) how we determine correctness/validate the results

Then have a separate section that briefly lists what is next)

src/clifft/sampling/hip/device_program.h
```diff
@@ -0,0 +1,98 @@
...
+    RotateActivePauli,
+    PromoteDormantRotation,
+    MeasureActivePauli,
+    MeasureDormantRandom,
+    RecordClassical,
+    DefineSymbol,
+    ApplyReadoutNoise,
+    WriteDetector,
+    WriteObservable,
+    WriteExpectationValue,
```
bachase, line 25: Where do we catch at compile time if a new action is added?

src/clifft/sampling/hip/executable.h
```diff
@@ -0,0 +1,60 @@
...
+#include "clifft/sampling/plan.h"
+
+#include <cstdint>
+#include <span>
+#include <vector>
+
+namespace clifft::sampling::hip {
+
+// The first device tier assigns a complete shot to one thread. Wider plans
+// will be accepted when the cooperative LDS tier owns their state layout.
```
bachase: LDS is a new acronomym -- just use clearer language here.

mkdocs.yml
```diff
@@ -122,6 +122,7 @@ nav:
   - Development:
     - Contributing: development/contributing.md
     - Building from Source: development/building.md
+    - Experimental HIP Backend: development/hip-backend.md
```
bachase: Should we instead start a new backends category? Or while we are developing an experimental category?

src/clifft/sampling/hip/device_program.h
bachase: Executable::lower_action already uses an exhaustive std::visit whose fallback is a dependent static_assert. Because the host-only clifft_hip_plan target builds in ordinary CPU CI, adding a SamplingAction alternative without explicit HIP lowering fails that build. I added a comment here to make that contract discoverable.

mkdocs.yml
bachase: Nested the page under Development > Experimental. That keeps its status explicit now and leaves Backends as the natural destination if the API later becomes supported.

src/clifft/sampling/hip/executable.h
bachase: Replaced LDS with the plain description: cooperative thread blocks keeping state in on-chip shared memory.

tools/ci/audit_hip_fatbin.py
bachase: Removed the repository audit and CI step in 9b2fba55. The ROCm job still compiles the gfx942 device code and links clifft_hip_tests. I also revised the PR description so the four-kernel extraction is recorded only as one-off manual evidence, while MI300X execution remains the experimental validation gate.

docs/development/hip-backend.md
bachase: Reworked this into separate sections for tests without a GPU, tests on AMD hardware, how correctness is determined, and brief next work. I also corrected the documented build command to compile both test executables.

src/clifft/sampling/hip/executable.h
```diff
@@ -0,0 +1,61 @@
...
+#include <vector>
+
+namespace clifft::sampling::hip {
+
+// The first device tier assigns a complete shot to one thread. Wider plans
+// will be accepted when cooperative thread blocks keep their state in on-chip
+// shared memory.
+inline constexpr uint32_t kThreadPerShotMaxActiveWidth = 4;
+
+class Executable {
```
bachase: Is it confusing we call this Executable but in the CPU backend we call it ExecutablePlan?

FarLab:
Reviewed with Claude Code (findings adversarially verified; two plausible-looking bugs were refuted with code-level proof and dropped). **The port itself checks out**: the device rotation/promotion/measurement/expectation math was verified term-for-term against the CPU kernels including sign conventions and the conjugate-phase factor, noise-firing probabilities match the CPU builder, and the exhaustive-visit `static_assert` genuinely fails CPU CI for new unhandled `SamplingAction`s. Nice slice.

1. **Seeded CPU and HIP runs are bit-correlated, not independent** (`sampler.hip:53`). The kernel re-declares the CPU executor's seed domain (`kSamplingExecutorDomain = 0x01`) and reproduces `derive_state` bit-for-bit, so for the same seed, HIP shot *i* and CPU shot *i* run identical xoshiro streams. For a no-noise circuit both backends decide the first branch with the same formula, so `sample(cpu_plan, N, seed=42)` and `hip::sample(..., {.seed=42})` return byte-identical rows — a user splitting a 2M-shot job across CPU+GPU gets 1M duplicated samples and underestimated error bars, and the docs' "different random-number ordering" claim is false exactly when it matters. Fix: give HIP its own domain constant in `util/shot_seed.h` (like the trajectory domains 0x11/0x12).
2. **The device-free conformance tests never run in CI** (`ci.yml:113`). `hip-offline-build` compiles `clifft_hip_tests` but has no ctest step, no other job enables `CLIFFT_ENABLE_HIP`, and the job isn't in `ci-gate`'s needs — so even the compile isn't enforced by branch protection. `ctest --test-dir build-hip` (skips report cleanly without a GPU) plus a `ci-gate` entry closes both gaps.

**Smaller items:**

3. `sample_survivors(keep_records=false)` still allocates and downloads shots-scaled records/detectors/exp_vals it provably never reads — O(shots) memory where the CPU path is O(1). (May be subsumed by #395's batching; worth checking there.)
4. Kernel shot index is computed in uint32 while the host can launch `ceil(shots/block)*block > 2^32` threads for non-power-of-two block sizes near max shots — tail threads wrap and race shots 0..k. Compute in uint64.
5. `sampler.hip` privately re-implements five bit-identity-critical helpers (xoshiro, mix64/derive_state + the domain literal, `kInvSqrt2`/`kLogHalf`, `insert_zero_bit`, `compact_rows`). Nothing couples them to the originals at compile time; sharing them via host/device-annotated headers prevents silent drift — drift has already cosmetically begun.
6. The nondiagonal measurement loop calls `compact_nondiagonal` 3x per packed index (both branches, then the winner again) plus an element-wise copy-back; fusing the probability pass is bit-identical under IEEE, and full-capacity scratch enables a pointer swap.
7. `pair_stride_or_z_without_pivot`'s measurement use is provably a disguise for `action.z` (the pivot bit is already cleared by `insert_zero_bit`); the field can become single-purpose `pair_stride`.
8. The "visible records occupy `[0, num_visible)`" prefix-slicing rule that all three entry points rely on exists only as a comment on the CPU Executor — stating and enforcing it in `SamplingPlan::validate()` would put both backends on a checked contract.
9. `ProgramView.num_visible_records` and `Action.reserved_index` are dead on the device; deleting them simplifies the kernel-facing ABI struct.

One cross-backend note: finding 1 applies verbatim to the CUDA slice on `research/sampling-gpu-bench` (it transplanted the same `0x01` literal) — whatever domain scheme lands in `shot_seed.h`, we'll mirror it there. A good argument for the shared-scaffold approach: one fix location serves both vendors.

bachase:
Thanks—addressed the actionable items across #384 and stacked #395: CPU and HIP now use distinct RNG domains while HIP remains same-seed and batch-size repeatable; the offline HIP tests are required by ci-gate; launch indexing is 64-bit; the packed pairing metadata and dead fields are simplified; and the record-prefix contract is documented. #395 also skips unused record, detector, and EXP_VAL device-to-host transfers for aggregate-only survivor sampling.

Broader host/device helper unification and measurement-loop tuning remain hardware-profiled follow-ups; device-side survivor aggregation is likewise still follow-up work.

src/clifft/sampling/hip/executable.h
bachase: Renamed the C++ type to hip::ExecutablePlan and the source/test files to executable_plan.* and test_hip_executable_plan.cc. The experimental Python facade keeps the shorter user-facing Program name.

Issue #385, opened 2026-08-23
Replace the Stim build dependency with native Pauli and tableau support


bachase:
## Problem or motivation

Clifft fetches and builds Stim as a production dependency, but uses only a narrow part of its functionality: Pauli/tableau algebra, Clifford-frame updates, composition/inversion, and final-state query support. Stim also contains general-purpose parsing, sampling, circuit generation, detector-error analysis, diagrams, commands, and other functionality that Clifft does not need.

The goal is to reduce clean build time, build-system complexity, and the WebAssembly/package dependency footprint. This proposal does not assume a sampling-throughput improvement; Stim is already outside the ordinary per-shot dispatch path.

## Proposed solution

Implement compact Clifft-owned Pauli and Clifford-tableau types sufficient for the frontend, sampling planner, optimizer, and exact state queries. Provide direct updates for the supported Clifford gates, phase-aware Pauli operations, tableau composition/inversion, and planner coordinate transformations. Then migrate the current Stim-typed fields and remove `FetchStim.cmake` and the `libstim` production link.

Keep the Python Stim package as a test-only oracle for stochastic equivalence and differential validation. SOFT/SymFT demonstrates that the required algebra can be implemented in-tree: https://github.com/haoliri0/SOFT

Acceptance criteria:

- No production CMake fetch or link dependency on Stim.
- Existing supported gates, sampling behavior, statevector queries, and basis-probability queries remain correct.
- Independent Stim-based stochastic tests remain available as development tests.
- Record clean-build time and artifact-size changes; require no material sampling regression, without treating runtime speedup as a goal.

## Alternatives considered

As an interim experiment, define a narrow Stim target containing only the required implementation files. This may recover much of the clean-build benefit, but retains the fetch dependency and reliance on Stim internal C++ APIs.

bachase:
Implementation and review hardening are complete as [stack #388](https://github.com/unitaryfoundation/clifft/pulls?q=is%3Apr+stack%3A388). Each semantic layer can be reviewed independently:

- [x] **Layer 1 - specification and independent oracles:** [#386](https://github.com/unitaryfoundation/clifft/pull/386) documents the conventions and strengthens equivalence coverage without changing production behavior.
- [x] **Layer 2 - unused native algebra:** [#387](https://github.com/unitaryfoundation/clifft/pull/387) adds Clifft-owned Pauli and tableau primitives with differential Stim tests while production still uses Stim.
[...]

PR #391, opened 2026-08-23
build: remove Stim from production targets

bachase:
Keep Stim isolated as an independent C++ test oracle while production C++, Python, and WebAssembly builds use only native Clifft tableau code. This reduces production build time and build-tree size without claiming a runtime performance benefit.

Part of #385.


FarLab:
Reviewed this closely (with Claude Code doing the heavy lifting; findings below were adversarially verified, several candidates were refuted and dropped). **The migration itself checks out**: we independently confirmed no stim usage remains in `src/` or `tools/`, a `CLIFFT_BUILD_TESTS=OFF` build of `clifft_core` compiles and links clean, and the moved `tests/stim_mask.h` is logic-identical to the deleted production header. No correctness issues found in the production code.

The findings are all about the *enforcement* layer — the no-Stim invariant is currently easier to break silently than the PR intends:

1. **The shipped wheel is never checked for Stim** (`ci.yml` ~332). `audit_python_exports.py` only inspects dynamic/exported symbols, and `clifft_core.lds`/`.exp` deliberately localize everything else — so statically re-linked Stim would be invisible to it. The package-smoke job that builds the actual distributed wheel runs no audit at all. Suggested fix: in the existing smoke step, scan the wheel's `_clifft_core.so` for *local* stim symbols (`nm` without `-D`/`-g`).
2. **The production guard has no positive control** (`ci.yml` ~136). `test ! -e build-production/_deps/stim-src` probes a CMake-internal path keyed to the fetch name; if `FETCHCONTENT_BASE_DIR` is ever set or the fetch is renamed/vendored, the check goes vacuously green forever. A one-liner makes it self-testing: also assert the path *does* exist in the adjacent tests-ON build.

**Smaller items:**

3. `cmake/FetchStimTestOracle.cmake:35` uses single-arg `FetchContent_Populate`, deprecated since CMake 3.30 (CMP0169) — it already warns on ubuntu-24.04's CMake and hard-fails under `CMP0169=NEW`. Since this file is new, the modern `FetchContent_Declare(... EXCLUDE_FROM_ALL)` + `MakeAvailable` form seems worth adopting now.
4. The guard's `cmake --build` (`ci.yml` ~129) adds no signal — the stim-absence answer is fully decided at configure time — but costs a cold full Debug compile in the gating job on every PR. Configure alone gives the same verdict.
5. Two pure Stim-oracle tests (`stim::Tableau` identity/Hadamard, `then()`/`inverse()` composition) stayed in `tests/test_hir.cc` under `[hir]` tags. They cover `prepend_H_XZ`/`prepend_ZCX`, which `test_stim_contract.cc` doesn't assert yet but the frontend differential helpers use — so moving (not deleting) them into the contract file would complete the localization this PR sets up.
6. `FetchStimTestOracle.cmake:12`: `STIM_BUILD_PYTHON`/`STIM_BUILD_TESTS` are no-ops — Stim v1.15.0 defines no such options; exclusion comes entirely from `EXCLUDE_FROM_ALL`. Can be deleted.
7. `tests/stim_mask.h:23`: the hand-rolled trailing-word clear could be `dst.zero_out()` + copy, mirroring the sibling's `dst.clear()` shape. Cosmetic.

None of this blocks the PR's purpose — nice measured build-time win, and the stack structure made this easy to review.

cmake/FetchStimTestOracle.cmake
```diff
@@ -0,0 +1,37 @@
+# Fetch Stim v1.15.0 only for the independent C++ test oracle.
```
bachase, line 1: Do w really need stim  C++ long term for oracle? Can we just do the testing in python, or use a different C++ oracle, to simplify the build dependencies?

.github/workflows/ci.yml
```diff
@@ -126,6 +126,16 @@ jobs:
           enable-cache: true
           python-version: "3.12"
 
+      - name: Build production C++ without test dependencies
```
bachase, line 129: Do we really need this test long term? I get its useful in the interim, but once stim is gona ... whats the point of testing that it is still gone?

docs/getting-started/installation.md
```diff
@@ -39,9 +39,8 @@ uv pip install -e .
 uv run python -c "import clifft; print(clifft.version())"
 ```
 
-!!! info "First build takes 10-15 minutes"
-    The initial build compiles Stim (a dependency) from source, which has many files.
-    Subsequent incremental builds are fast.
+Source installation builds Clifft's native core directly; Stim is not fetched
```
bachase, line 42: As a go forward comment, we don't need this detail since stim dependency is no longer there.

bachase:
Thanks for the detailed review. We took the broader route and now remove the C++ Stim dependency entirely, while retaining Python Stim as an end-to-end dev/test oracle.

The follow-up is split into three reviewable layers:

- [#397](https://github.com/unitaryfoundation/clifft/pull/397) adds scalar Clifford and Pauli-channel references beside Stim, proves three-way equivalence at packed-word boundaries, and freezes the benchmark fixtures against the Stim generator.
- [#400](https://github.com/unitaryfoundation/clifft/pull/400) migrates every C++ consumer to those validated references and fixtures, then deletes the adapter and Stim contract tests.
- [#401](https://github.com/unitaryfoundation/clifft/pull/401) mechanically deletes the C++ FetchContent module/link and the temporary negative CI guard.

That changes the disposition of the findings:

1. and 2. We are not adding a wheel symbol audit or positive control. With #401 there is no C++ Stim fetch, target, include, adapter, or consumer to audit; the special no-Stim guard is removed as well.
3., 4., 6., and 7. The affected FetchContent file, guard, and adapter are deleted rather than revised.
5. The two scattered HIR Stim tests are removed with the other C++ Stim contract tests. Their relevant composition and inverse-state semantics are covered by the scalar reference that #397 first checks against both native Clifft and Stim, then #400 keeps as the C++ oracle.

A fresh #401 configure contains only Catch2 and FastFloat under `_deps`; the clean C++ test build succeeds and all 857 non-benchmark cases pass.

PR #395, opened 2026-08-24
feat(hip): add retained sampling and experimental Python workflow

bachase:
Retain uploaded HIP programs and precision-specific bounded workspaces across
calls, preserving global-shot RNG streams and survivor ordering across batches.
Reject overlapping sampler calls while keeping ROCm out of the ordinary Python
extension.

Expose compilation, plan inspection, FP32 and FP64 sampling, survivor sampling,
and forced replay through `clifft.experimental.hip`. Add GPU-free conformance
and packaging coverage plus developer documentation for lowering, packed data,
coefficient evolution, workspace layout, launch policy, and GPU targets.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,7 @@ jobs:
         run: cmake --build build-hip --target clifft_hip_tests --parallel 2
 
       - name: Run GPU-free HIP conformance tests
-        run: ctest --test-dir build-hip --output-on-failure -R HIP
+        run: ctest --test-dir build-hip --output-on-failure --no-tests=error -R HIP
       - name: Build editable HIP Python package
         env:
           CMAKE_ARGS: >-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,15 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Setup uv and Python
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: true
+          python-version: "3.12"
+
+      - name: Install Python binding build dependencies
+        run: uv sync --frozen --only-group dev --no-install-project
+
       - name: Configure gfx942 HIP build
         run: |
           cmake -S . -B build-hip -G Ninja \
@@ -115,6 +124,20 @@ jobs:
 
       - name: Run GPU-free HIP conformance tests
         run: ctest --test-dir build-hip --output-on-failure -R HIP
+
+      - name: Configure gfx942 HIP Python extension
+        run: |
+          cmake -S . -B build-python-hip -G Ninja \
+            -DSKBUILD=ON \
+            -DPython_EXECUTABLE=${{ github.workspace }}/.venv/bin/python \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCLIFFT_ENABLE_HIP=ON \
+            -DCMAKE_HIP_ARCHITECTURES=gfx942 \
+            -DCMAKE_HIP_COMPILER=/opt/rocm/llvm/bin/clang++ \
+            -DCMAKE_HIP_COMPILER_ROCM_ROOT=/opt/rocm
+
+      - name: Compile optional HIP Python extension
+        run: cmake --build build-python-hip --target _clifft_hip --parallel 2
 
   build-and-test:
     timeout-minutes: 45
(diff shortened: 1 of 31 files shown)


FarLab:
Reviewed with Claude Code (findings adversarially verified against the PR head; none of the scary-looking candidates was refuted outright, but several were downgraded after verification). **The batching refactor itself checks out**: seeded RNG streams are batch-invariant (shot streams depend only on root + global index, so results match the old single-launch code for any batch size), survivor compaction preserves row order and counts, validation still precedes device access, and a cross-file trace of every signature change found zero broken call sites.

1. **The non-overlap contract is documented but unenforced, and violations corrupt silently** (`hip_bindings.cc:129`). The bindings release the GIL around one shared mutable workspace with no lock or busy-flag — two Python threads sharing a `Sampler` interleave batches and staging reads and both get plausible-looking mixed rows, no exception. Cheapest fix that keeps the hot path clean: an atomic busy-flag that throws on overlapping entry, making the documented rule ("calls on one instance must not overlap") loud instead of silent.
2. **Every `import clifft` in a HIP build eagerly dlopens the ROCm extension** (`__init__.py:366` → `hip.py:24`). The experimental subpackage is wired via a bottom-of-file import (guarded only by a file-wide `noqa: E402`, no comment), and `hip.py` imports `_clifft_hip` at module scope — so non-HIP users of a HIP wheel pay ROCm load cost on import, and any import reordering breaks `import clifft` with nothing to catch it. A lazy module `__getattr__` (or importing `_clifft_hip` inside the functions) removes both hazards. (Graceful-degradation on broken ROCm is already handled nicely.)
3. **`ctest -R HIP` needs `--no-tests=error`** (`ci.yml:126`) — otherwise a discovery or naming regression turns the conformance step vacuously green. One flag.

**Smaller items:**

4. `hip.compile()` re-implements the compile pipeline in Python (parse→trace→passes→normalize) while `clifft.compile` delegates to C++ `prepare_hir_for_lowering`, plus a duplicate private `_DefaultPasses` sentinel and already-divergent error text. Since the backend's conformance premise is "the CPU plan is the oracle," the two compiles drifting apart is the failure mode to avoid — worth sharing the front half.
5. Download/copy efficiency cluster in `sampler.hip`: `download_rows` unconditionally downloads arrays dead on the current path (`survived` on `sample()`; records/detectors/exp_vals on `sample_survivors(keep_records=false)` — this extends the O(shots) issue noted on #384 into the batched form); `sample()` stages detectors/observables/exp_vals through host vectors then re-copies though `download` accepts arbitrary spans; and the per-shot record copy degenerates to one bulk copy when there are no hidden records.
6. Python facade hardcodes `65_536`/`256`, always overriding the nanobind defaults wired to `kDefaultMaxBatchShots`/`kDefaultBlockSize` — retuning C++ silently strands Python callers. `None`-sentinels falling through to native defaults keep one source of truth.
7. Host `coefficient_elements_per_shot` restates the kernel's inline stride formula (match today; divergence = OOB device writes). A shared host/device constexpr in `device_program.h` ties them — same shared-helpers theme as the #384 comment.
8. Binding-layer duplication: `vec_to_numpy` copied verbatim from `bindings.cc`, and the 8-positional-arg `SampleResult` construction now at five call sites in two TUs; also the `Executable` is retained three times per Python `Sampler` (Program + BoundSampler copy + Impl copy — a `const Executable&` accessor removes one), and `Program` exposes no way to learn the visible+hidden record count `replay_shot` demands.
9. Test/hardening notes: the fp32/fp64 dual-buffer one-always-empty invariant could be a single byte buffer; `utils_hip`'s forced-record helper conflates visible vs total record widths (latent — fails loudly on the first reset-containing circuit); its binomial tolerance is a second convention alongside conftest's (σ=6 vs 5); and `DeviceBuffer::download`'s loosened size check trades an invariant for prefix semantics — a separate `download_prefix` would keep both.

The workflow/packaging parts of the PR (uv wiring, `CMAKE_ARGS` folding, SKIP_RETURN_CODE plumbing, STABLE_ABI) all verified clean.

bachase:
@FarLab Thanks — addressed the pre-merge findings at 2b73743c:

- overlapping calls now fail through an atomic binding guard before the GIL is released;
- clifft.experimental and the HIP extension load lazily, while omitted batch/block values inherit native defaults;
- CPU and HIP compilation now share the same C++ HIR preparation helper;
- Python exposes total visible-plus-hidden record width, and forced replay now includes a reset-hidden-record fixture;
- host/device coefficient stride is one constexpr contract, partial transfers use an explicit download_prefix operation, and CTest uses --no-tests=error.

The earlier 4852229f already removed unused aggregate-only record, detector, and EXP_VAL downloads. I am leaving direct-to-result copy tuning, broader binding deduplication, the single coefficient-buffer idea, and tolerance consolidation for hardware-profiled or housekeeping follow-ups.

PR #403, opened 2026-08-25
feat(sampling): add packed cross-shot execution

### LINKED ISSUE #313: Add packed cross-shot execution to the symbolic sampler
## Goal

Add a single-threaded batched execution mode that processes multiple shots through the same symbolic plan while packing the classical symbolic layer across shots.

This is a future performance enhancement. It does not block the symbolic sampler cutover in #247 or the executable-plan work in discussion #298.

## Proposed direction

- Represent independent symbol assignments, measurement records, branch bits, detector bits, observable bits, and affine-expression results as packed columns across a batch.
- Reuse the same plan traversal and symbolic-expression work for every shot in the batch.
- Retain a separate dense active-state vector for each live shot, initially using shot-major storage and the existing contiguous kernels for each shot.
- Choose batch size from the planned peak active width and an explicit cache or memory-footprint budget instead of using one fixed batch size.
- Track rejected shots with a live mask and compact survivors only when a cost model predicts that moving them is cheaper than carrying inactive lanes.
- Keep a single-shot fallback for small shot counts, large active states, continuations, or other cases where batching does not pay.
- Preserve allocation-free and exception-free hot execution by preparing all batch, record, symbolic, state, and compaction storage before dispatch.
- Preserve deterministic RNG behavior, shot order, records, survivor semantics, replay behavior, and postselection results.

Batching here means processing several shots together within one worker thread. Cross-shot worker parallelism and intra-shot kernel parallelism are separate concerns.

## Evaluation

Benchmark at least:

- pure-Clifford and active-width-zero workloads where symbolic evaluation dominates;
- small-active-width noisy and postselected workloads;
- representative cultivation and distillation workloads;
- larger active widths where cache pressure may eliminate the benefit;
- low shot counts where the single-shot path should remain preferable.

Report throughput, memory use, selected batch sizes, masking or compaction behavior, and time divided between symbolic evaluation and dense kernels.

## Reference

SymFT uses packed columns for the classical symbolic layer while retaining one dense active-state vector per live shot, with an adaptive batch size based on active-state footprint:

https://arxiv.org/abs/2607.28600

bachase:
Closes #313.

## Summary

- add a CPU `BatchExecutor` that traverses one immutable executable plan for packed shot lanes
- bit-pack symbols, expression registers, records, detectors, observables, readout faults, and live masks while retaining one existing dense `State` and RNG stream per lane
- preserve seeded shot identity, stable survivor order, expectation sidecars, fixed-k conditioning, postselection, and cross-shot worker composition
- add cost-based live-lane compaction with fixed preallocated storage
- expose `batch_size="auto" | int` on `sample`, `sample_survivors`, `sample_k`, and `sample_k_survivors`
- add profiler control through `CLIFFT_PROFILE_BATCH_SIZE`

## Policy and limits

Automatic mode currently selects up to 512 lanes only for noiseless, peak-active-width-zero CPU plans with at least 64 shots per worker. Benchmarks showed that noisy presampling and action-major dense-state traversal do not yet reliably beat the scalar executor, so those plans conservatively resolve to one lane. Explicit positive capacities force the packed path up to 2048 lanes for profiling and expert use.

Packed execution does not support transition instruments, traps, continuations, record replay, externally supplied presampled symbols, WebAssembly, or combination with intra-shot workers. The existing scalar and trajectory paths remain unchanged.

## Validation

- 876 native test cases, 2,320,358 assertions: pass
- sampling and importance Python suites, 141 tests: pass
- repository pre-commit suite: pass
- profiler target: builds
- forced capacities 2, 63, 64, and 65 match scalar rows for ordinary, survivor, fixed-k, and fixed-k-survivor sampling
- dedicated regression verifies expectation columns computed before rejection remain aligned after compaction

## Benchmark note

On the current VM, automatic batching reduced median time for the existing 50-qubit deep-Clifford 100k-shot workload from 0.0754 s to 0.0587 s, about 22 percent. Noisy surface-code and active-width-10 cultivation workloads resolve to scalar and remained within run-to-run noise.

### DIFF
--- a/docs/guide/importance-sampling.md
+++ b/docs/guide/importance-sampling.md
@@ -331,7 +331,7 @@ postselection. Key observations:
 
 ## API Reference
 
-### `clifft.sample_k(program, shots, k, seed=None, threads=1, thread_layout=None, intra_shot_min_active_width=None)`
+### `clifft.sample_k(program, shots, k, seed=None, threads=1, thread_layout=None, intra_shot_min_active_width=None, batch_size="auto")`
 
 Sample with exactly `k` forced faults per shot. Returns
 a `SampleResult`, just like `clifft.sample()`. Results must be weighted by $P(K=k)$ for correct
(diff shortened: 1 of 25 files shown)


bachase:

PR #404, opened 2026-08-25
perf(sampling): specialize packed hot paths

bachase:
Stacked on #403. This keeps the initial packed executor review separate from the data-driven hot-path follow-ups.\n\n## Summary\n\n- add aggregate-survivor profiling controls and postselect-all fixture support\n- keep counts-only packed sampling out of row materialization and final survivor compaction\n- compile direct-rotation runs of 8 to 32 actions and execute each run shot-major to reuse per-lane state\n- retain ordinary dispatch for shorter runs so surface-code, cultivation, and distillation plans stay neutral\n\n## Performance\n\nOn the development VM, 100k-shot coherent d3r3 sampling at an explicit 2048-lane capacity improved from about 333 ms to 273 ms (1.22x throughput). Counts-only postselection also avoids storage and compaction that callers do not consume.\n\n## Validation\n\n- full native suite: 2,321,267 assertions across 884 cases\n- explicit capacities through multi-word and partial batches\n- repository pre-commit suite\n\nNo clifft-bench update is included, and this PR should remain stacked until the performance series is ready.

### DIFF
--- a/src/clifft/sampling/batch_executor.cc
+++ b/src/clifft/sampling/batch_executor.cc
@@ -111,8 +111,10 @@ uint32_t resolve_batch_capacity(const ExecutablePlan& plan, uint32_t shots, uint
 #endif
 }
 
-BatchExecutor::BatchExecutor(const ExecutablePlan& plan, uint32_t lane_capacity)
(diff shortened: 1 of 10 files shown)


bachase:

PR #405, opened 2026-08-25
perf(sampling): share presampled batch expressions

bachase:
Stacked on #404. This closes the main symbolic-layer gap identified in the SymFT comparison without expanding the product surface.\n\n## Summary\n\n- split each affine register into its presampled contribution and causal residual during executable preparation\n- intern repeated presampled partials and compile an exact parent/delta/copy program by dependency level\n- score candidate parents through a symbol inverted index, avoiding a quadratic preparation scan\n- sort each level by source symbol for locality and retain the original dependency path when a conservative cost model rejects the program\n- stabilize the x86 packed hot-loop layout against unrelated builder code-size changes\n\nThe runtime performs no topology discovery or allocation; all sharing and scheduling metadata is immutable and precomputed.\n\n## Performance\n\nFor 100k aggregate, postselect-all shots with an explicit 2048-lane batch on the development VM:\n\n- surface d7: about 211 ms with the planner disabled to 152.5 ms enabled; scalar is about 234 ms\n- cultivation d5: about 1,138 ms on the parent to 1,029 ms here in an alternating comparison\n- distillation and small coherent plans reject the expression program and retain the existing path\n\nExecutable preparation rises from roughly 1.6 to 12.8 ms on surface d7 and 2.1 to 16.5 ms on cultivation d5. This is a one-time compile cost, and the inverted index replaces the earlier 100+ ms quadratic prototype.\n\n## Validation\n\n- dedicated scalar-vs-packed test forces parent deltas, constant inversion, and duplicate copies across 2, 63, 64, 65, and 128 lanes\n- full native suite: 2,321,272 assertions across 885 cases\n- repository pre-commit suite\n- final profiler build and checksum comparison\n\nNo clifft-bench changes are included, and this PR should remain stacked until the complete performance series is ready.

### DIFF
--- a/src/clifft/sampling/batch_bits.cc
+++ b/src/clifft/sampling/batch_bits.cc
@@ -4,6 +4,7 @@
 #include <bit>
 #include <limits>
 #include <stdexcept>
+#include <utility>
 
 namespace clifft::sampling {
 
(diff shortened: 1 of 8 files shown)


bachase:

PR #406, opened 2026-08-25
perf(sampling): select profitable packed plans automatically

bachase:
Stacked on #405. This turns the measured wins into a conservative automatic policy while retaining scalar execution for the observed losing regimes.\n\n## Summary\n\n- raise the automatic ceiling from 512 to 2048 lanes within the existing 768 KiB dense-state budget\n- automatically select large packed batches when executable preparation found profitable shared presampled expressions\n- make selection output-aware so counts-only postselected plans through peak active width 5 can use the packed aggregate path\n- require at least 512 retained lanes for these expanded regimes and keep readout-heavy, wider, or unprepared dense plans scalar\n- preserve the existing five-argument capacity resolver ABI and document the policy\n\n## Performance\n\nMedian 100k-shot results on the development VM:\n\n- surface d7 aggregate postselect-all, auto: 234 ms scalar to 156 ms (1.50x throughput)\n- surface d7 fixed rows, auto: 569 ms scalar to 494 ms (1.15x)\n- distillation aggregate postselect-all, auto: 182 ms scalar to 107 ms (1.71x)\n- 50-qubit deep Clifford: 65.6 ms scalar, 59.0 ms at 512 lanes, 57.4 ms at 2048 lanes\n\nA rotation-heavy width-5 control measured 555 ms scalar versus 560 ms at its 1024-lane footprint capacity, so it remains scalar. Width-6, coherent width-7, and cultivation width-10 controls likewise remain outside the expanded policy.\n\n## Validation\n\n- policy tests cover zero-width, prepared-expression, row versus aggregate, minimum-amortization, state-footprint, explicit override, and intra-shot exclusions\n- full native suite: 2,321,279 assertions across 885 cases\n- repository pre-commit suite\n- automatic-mode profiler checksums match their scalar and explicit-capacity controls\n\nNo clifft-bench changes are included, and this PR should remain stacked until the full series is ready.

### DIFF
--- a/docs/guide/simulation.md
+++ b/docs/guide/simulation.md
@@ -269,12 +269,13 @@ Within each cross-shot worker, the symbolic sampler normally processes shots
 in an adaptive packed batch. Classical symbols, expression results, records,
 detectors, and observables are bit-packed across the batch, while each live
 shot retains its own dense active state and random stream. The automatic
-capacity is bounded by the worker's shot count, a maximum of 512 lanes, and a
-768 KiB dense-state footprint budget. Automatic mode currently selects packing
-only for noiseless peak-active-width-zero plans with at least 64 shots per
-worker. Benchmarks show that per-lane noise sampling and action-major traversal
-of dense states still favor the scalar executor, so noisy and active-state
-plans conservatively fall back to it.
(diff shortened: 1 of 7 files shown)


bachase:

PR #407, opened 2026-08-25
feat(api)!: add high-throughput packed batch sampling

### LINKED ISSUE #313: Add packed cross-shot execution to the symbolic sampler

bachase:
Closes #313.

## Summary

- add a basis-major, shot-interleaved state and packed fixed-plan executor
- precompute categorical-noise factorization, shared affine-expression initialization, record-parity sidecars, and compact packed noise-carrier mappings during lowering
- specialize uniform sparse noise, symmetric readout noise, dormant measurements, survivor aggregation, and stable live-lane compaction
- expose `batch_size="auto" | 1 | int` on `sample`, `sample_survivors`, `sample_k`, and `sample_k_survivors`
- preserve exact seeded replay within the same execution configuration; scalar and packed configurations are statistically equivalent but intentionally use separate RNG streams
- preserve observable record snapshots across later readout rewrites, and validate packed parity sidecars against their affine outcomes
- choose automatic lane capacity independently of thread count, then cap effective packed workers to the aggregate memory budget
- allocate retained symbol and fixed-fault readout columns only in the execution modes that use them
- prepare fixed-k conditioning tables once per call and share them across scalar or packed workers

Automatic batching is conservative: at least 64 shots, peak active width at most 5, about 8 MiB per packed worker, and at most 2048 lanes. Lane capacity depends on the plan, shots, output mode, and requested batch setting, but not the resolved thread count. The sampler separately caps the effective worker count by the number of batches and the 64 MiB aggregate retained-worker budget. The estimate covers executor storage, survivor counters, and fixed-k selection scratch. Fixed-k's potentially large immutable DP preparation is retained once per sampling call rather than once per worker. Packed ranges remain capacity-aligned, so `threads=1`, explicit thread counts, and `threads="auto"` preserve the same seeded rows. Explicit capacities remain available for profiling but are rejected when the aligned dense state would exceed 64 MiB. Packed executors are scoped to a sampling call and are not retained in the compiled program between calls.

Instruments/continuations, intra-shot worker combinations, WebAssembly, and exact record-probability replay remain on their existing paths. Product-state factorization is deliberately deferred; it primarily affects the coherent-noise cases outside this PR's target set.

## Performance

The comparison uses the pure-Clifford surface-code circuits from the SymFT paper repository with detector postselection disabled, matching the normalized clifft-bench configuration. SymFT is the paper-release implementation (`0.1.0`, source-matched to commit `9ec5790`) with AVX-512 enabled. Both tools used one pinned CPU, one thread, one-million-shot calls, wall-clock timing, and five repeats of at least five seconds. The table reports the median throughput. Each tool used its selected capacity: Clifft 2048/1024 lanes for d7/d9 and SymFT 2048 lanes for both.

| paper circuit | Clifft | SymFT | Clifft throughput advantage |
|---|---:|---:|---:|
| surface d7 r7 | 4.085 Mshots/s | 3.333 Mshots/s | 22.6% |
| surface d9 r9 | 1.663 Mshots/s | 1.360 Mshots/s | 22.3% |

The compact noise-carrier mapping was the material final algorithmic improvement: it avoids clearing the much larger sparse global-symbol matrix in prepared mode and improved both paper circuits by about 11-12%. On d9, hardware counters moved from 1.70 to 1.89 IPC while the L1 data-cache miss rate fell from 13.46% to 8.68%. A sparse dirty-column clear was also profiled and rejected because it regressed d9 by about 6%.

The final retained-state cleanup improved local 100,000-shot sampling by another 12.3% at one thread and 19.2% at 16 threads on d7, and 13.2% at one thread and 8.8% at 16 threads on d11, relative to the preceding PR head. It removes the unused full global-symbol matrix in compact-carrier mode, omits fixed-fault readout storage for ordinary sampling, reuses compaction scratch, and stops compacting columns that are dead before action dispatch. The SymFT table above predates this cleanup and is therefore conservative.

The final fixed-k audit also removed per-worker duplication of the nonuniform conditioning table. In a deliberately preparation-dominated local profile with 1,024 nonuniform sites, `k=512`, 16 shots, eight scalar workers, and explicit `batch_size=1`, median call time fell from about 14.3 ms to 0.7 ms. A 4,096-site, `k=2,048`, eight-worker call now retained about 119 MiB RSS rather than multiplying a greater-than-64-MiB table by every worker. Row scaling also prevents central Poisson-binomial coefficients from overflowing; the old Release path collapsed that case onto the final 2,048 sites, while the corrected seeded sample selected 1,019 sites from the first half and 1,029 from the second.

Earlier production-workload profiling at 100,000 shots and 2048 lanes also showed runtime reductions versus the same SymFT release of 31.5% on cultivation d3, 32.1% on color-code distillation, and 6.7% on surface d7 before the compact-carrier optimization. A readout-heavy stress profile showed automatic packed execution at about 14x scalar throughput for 0.1% symmetric readout noise, 1.3x at 10% symmetric readout noise, and neutral/slightly faster for asymmetric readout noise.

## Compile-time cost

The feature necessarily adds compile-time preparation for plans that benefit from packed execution. This revision hoists the cheap expression-size rejection ahead of GF(2) factorization and removes the duplicate full-plan validation while retaining validation in every executable constructor. The local surface-d7 planning-plus-construction median is about 11.3 ms, down from 11.8 ms at the preceding PR head; the remaining increase over main is the compiler-side packed expression, parity, and presampling program construction that keeps runtime execution allocation-free and free of topology analysis.

## Validation

- full Debug C++ suite: 906/906 tests passed
- focused packed replay, capacity-policy, and presampled-expression suites
- rebuilt Python extension: 1,344 passed, 6 skipped
- executable documentation examples: 33 passed, 8 skipped
- WebAssembly build and smoke suite

The new coverage includes exact same-mode replay across batch capacities and worker schedules, true multi-batch/multi-worker execution with partial final batches, scalar/packed statistical equivalence, detectors and observables with expected-bit corrections, historical record snapshots across readout rewrites, an analytic categorical-noise oracle for compact carriers, sparse and certain noise, symmetric readout endpoints and rates, active/dormant measurements, expectations, postselection compaction, aggregate survivor counts, shared and numerically stable fixed-k sampling, complete-worker automatic memory budgets, and aligned explicit-state limits.

The executor-independent syndrome value representation cleanup discussed during final review is tracked separately in #411 because it changes plan semantics and lowering across scalar CPU, packed CPU, and HIP rather than the packed executor alone.


bachase:
## Re-review of 45832726 + 3df93068

Verified the two fix commits empirically: rebuilt Debug at 3df93068, re-ran the original divergence repro, and ran the full `ctest -E Bench` suite (all pass).

### Fix verification

| Item | Status |
|---|---|
| Observable snapshot bug | **Fixed.** Record-generation tracking correctly drops the parity sidecar when a record is rewritten after inclusion. Repro (`M 0 / OBSERVABLE_INCLUDE(0) rec[-1] / READOUT_NOISE(0.5) rec[-1]`) now gives scalar = packed = 0 observable ones with measurements still ~50% noisy. Both flavors (post-include noise, straddled double-include) covered by the new planner test and the exact scalar==packed differential across 5 capacities. |
| Explicit `batch_size` memory bound | Fixed — 64 MiB explicit-state budget with a clear diagnostic; boundary math checks out (width-18×2 allowed at ~50 MB, width-19×2 rejected). |
| Profiler timing | Fixed — checksum back outside the measured window. |
| `validate()` parity↔outcome cross-check + stale-readout-source check | Added, both negative-tested. |
| Inspection sidecar rendering | Added in both formats, digest updated. |
| Differential/auto-replay test gaps | Closed. |
| Hot-path items (double classification, span hoist, live-word parity bound, compaction locality, `rng_.next_double()`, wasm builder gate) | All fixed; verified the stale scratch tail after the partial parity fill is safe (every consumer masks with `live_words_`, whose tail is zeroed). |

### New finding: plan-compile latency

Median `plan_sampling` + `ExecutablePlan` construction on `surface_d7_r7` (Release, 25 reps, this machine):

| build | median |
|---|---:|
| main (50f2cfa1) | 6.55 ms |
| PR head before fixes (69453d5c) | 10.52 ms (+61%) |
| PR head now (3df93068) | 12.14 ms (+85%) |

The +61% is the batch preparation (record parities, GF(2) factorization, presampled tape); the extra +15% is the new semantic `validate()` cross-check. This largely undoes the #67 compile-perf win (14 → 6.4 ms), which was motivated by workflows that compile many plans. Possibly an acceptable trade, but worth an explicit decision. Cheapest mitigations:

1. `validate()` runs twice per compile (end of `plan_sampling` and again in `ExecutablePlanBuilder::compile`); the semantic parity cross-check could run in only one of them, or be `NDEBUG`-gated (structural checks stay).
2. The `original_presampled_terms < kMinBatchExpressionTerms` half of the cost bail in `prepare_batch_expression_initialization` can be hoisted before the GF(2) factorization — the count is already known there (helps small plans).
3. The per-outcome `residual` allocation and the O(blocks × shared-symbol fan-out) parent scan in the same function remain as further compile-time levers.

### Still open (cleanup, non-blocking — fine as a follow-up PR)

- `resolve_batch_capacity` still exists twice (header wasm inline vs native .cc) and the copies have drifted further (native gained the 64 MiB check); plus the four per-entry-point `#if !defined(__EMSCRIPTEN__)` blocks in `sampler.cc`. One shared policy function + a single availability constant would collapse all of it.
- `sampler.cc`: ~20-line lane→row copy block duplicated between `sample_fixed_batches`/`sample_surviving_batches`; `checked_size` lambda ×4.
- Static/dynamic fused-orbit templates ~90 lines near-identical (dedup via a matrix-provider parameter; verify codegen).
- Presampled finalize if/else now at 5 sites → `finalize_presampled_symbols()`.
- `BatchOption`/`parse_batch_option` duplicate the thread-option parser; the 3-line OpenMP-SIMD pragma guard ×11 wants a `CLIFFT_OMP_SIMD` macro; `first_effect`/`xor_effects` re-derive `mask_view.h` primitives; 2×2 batch worker structs restate the scalar quartet.
- Dead API: `BatchExecutor::dust_clamps()/compactions()/attempted_shots()/lane_capacity()` have no callers; `count_lane_bits` is public but assert-only.
- Low: cumulative-clamp final-outcome edge for hand-built plans; near-capacity-multiple worker imbalance (docs-only fix realistic given the replay contract).

**Bottom line:** both blockers are properly fixed with good coverage. I'd merge after an explicit call on the compile-latency regression (ideally taking mitigation 1 and/or 2); the cleanup cluster can follow separately.

---
*Review assisted by Claude (Opus 5).*

src/clifft/sampling/batch/executor.cc
```diff
@@ -0,0 +1,817 @@
...
+        }
+        return capacity;
+    }
+    if (shots < kDefaultMinAutoBatchShots) {
+        return 1;
+    }
+    if (plan.peak_active_width() > 5) {
+        return 1;
+    }
+    const uint64_t state_bytes_per_lane = interleaved_state_bytes_per_lane(plan);
```
bachase: Automatic selection budgets only coefficient and scratch storage, but each worker immediately allocates lane-scaled symbol, expression, record, detector, observable, readout, and expectation storage. At 2,048 lanes, every packed bit column costs 256 bytes, so long low-active-width plans can exceed the nominal conservative budget by orders of magnitude, multiplied by the shot-worker count. The included surface_d11 fixture already needs roughly 1.4 MiB of row-mode expression/record/detector columns before counting symbols. Please estimate the complete retained worker footprint when resolving automatic capacity and add a sidecar-heavy regression. The per-worker memory documentation should describe this packed scaling as well.

tests/test_sampling_executor.cc
```diff
@@ -1352,6 +1352,210 @@ TEST_CASE("Threaded fixed-row sampling preserves seeded shot order") {
     }
 }
 
+TEST_CASE("Explicit batch capacities replay seeded fixed rows") {
```
bachase, line 1355: These capacity replay checks always use one thread. Conversely, the threaded tests use 257 shots with automatic capacity 257, so sample_fixed_batches resolves batch_workers to one and never exercises parallel packed scheduling. Please compare threads=1 with threads>1 using an explicit small capacity and enough shots for several batches, including a partial final batch. The same gap exists for survivor compaction; covering sample_k as well would protect all documented replay guarantees.

CMakeLists.txt
```diff
@@ -258,5 +258,11 @@ else()
             ${CMAKE_CURRENT_SOURCE_DIR}/src
             ${CMAKE_CURRENT_BINARY_DIR}/generated
         )
+        add_executable(profile_interleaved tools/profile/profile_interleaved.cpp)
```
bachase: This adds a fourth profiling executable, but tools/profile/README.md still describes three harnesses and its documented build command, along with just profile-build, only builds profile_compile, profile_probability, and profile_sample. Please either integrate and document profile_interleaved or remove the standalone 210-line harness now that profile_sample exercises the production batching path.

bachase (commented):
Overall, the packed execution design looks sound: it preserves preallocation and exception-free hot execution, keeps topology planning in the compiler, uses deterministic RNG domains, and has strong lane-by-lane scalar references plus Stim statistical and deterministic-error oracles. I found no quantum-semantic mismatch in the kernels, measurement collapse, noise factorization, record snapshots, survivor compaction, or observable routing.

Local validation passed: 899/899 C++ tests; 1,341 Python tests passed with 7 skipped and 1 expected failure; all pre-commit checks; and a manual 5,000-shot, capacity-65 comparison produced identical rows with 1 and 4 workers. GitHub CI is also green.

I would address the packed-worker memory budgeting issue and add a true multi-batch/multi-worker replay regression before merging. I have also left one cleanup comment about the profiling harness. Further non-blocking simplifications worth considering are consolidating the duplicated measurement-dust classifier, auditing the now-one-line sample_records wrapper, and avoiding full batch-expression factorization work before its cost threshold rejects the optimization.

src/clifft/sampling/batch/executor.cc
bachase: Addressed in 15996612 and extended in dbe857d3. Automatic selection now estimates the complete retained BatchExecutor footprint, including coefficient state, global symbols, compact noise carriers, expression and record sidecars, outputs, readout storage, expectations, and per-lane/word scratch. It also multiplies by the packed worker count and enforces both an 8 MiB per-worker and 64 MiB aggregate budget. The d11 sidecar-heavy regression checks the resulting capacity, row/count modes, and worker-count sensitivity, and the guide now documents the full packed scaling.

tests/test_sampling_executor.cc
bachase: Addressed in 396fb1cb. The explicit-capacity replay suite now compares threads=1 and threads=4 at capacities 2, 63, 64, and 65 with 257 shots, forcing several packed batches plus a partial final batch. It covers sample, survivor compaction, sample_k, and sample_k_survivors.

CMakeLists.txt
bachase: Addressed in 324ec145. The redundant profile_interleaved target and 210-line harness were removed; profile_sample remains the single production batching profiler and the documented profile-build target is accurate again.

bachase:
Follow-up to the compile-latency re-review and final performance pass:

- `4b4e2004` takes both suggested low-complexity compile mitigations: the expression-size rejection now happens before GF(2) factorization, and the duplicate `plan.validate()` scan was removed from `plan_sampling`. CPU and HIP executable construction still perform full validation before execution. Surface-d7 planning plus construction is now about 11.5 ms median here versus the reviewed 12.1 ms; the remaining delta from main is the packed runtime program preparation.
- I profiled a parity-workspace reuse change for the remaining allocation concern; it produced no measurable compile improvement, so I reverted it rather than add complexity.
- `dbe857d3` maps sparse presampled noise symbols onto compact carrier columns. Prepared resets now clear those compact columns instead of the global symbol matrix. This improved the paper d7/d9 circuits by roughly 11-12%; d9 IPC rose from 1.70 to 1.89 and its L1D miss rate fell from 13.46% to 8.68%. A dirty-column sparse-clear alternative regressed d9 by about 6% and was rejected.
- Fresh wall-clock, one-thread, pinned-CPU medians against the paper-release SymFT source with postselection disabled are Clifft 4.085 vs SymFT 3.333 Mshots/s on d7 (+22.6%), and 1.663 vs 1.360 Mshots/s on d9 (+22.3%).

The PR body now records the design, limitations, compile-time tradeoff, and full measurement method. The larger cleanup cluster from the review remains intentionally out of scope for this performance PR.

bachase:
Addressed the worker-count-dependent automatic capacity concern in `4a6e1206`.

The policy is now explicitly two-stage:

1. select lane capacity from the plan, shots, output mode, and requested batch setting, independent of the resolved thread count;
2. cap the effective packed worker count by the number of batches and the 64 MiB aggregate retained-worker budget.

Because packed ranges remain capacity-aligned and batch RNG seeding depends on the batch's first shot rather than worker identity, `threads=1`, explicit threaded counts, and `threads="auto"` now keep the same seeded rows. The budget-bound d7 regression that previously selected 2048 lanes with one worker and 1024 with 16 now selects 2048 in both cases while reducing the effective worker count. A Python integration test exercises 32,768 d7 shots across 1 and 16 requested workers and compares measurements, detectors, and observables exactly.

I also took the two organization/documentation tweaks discussed alongside the fix:

- the eight new packed implementation files now live under `src/clifft/sampling/batch/` with non-redundant filenames;
- every batch policy constant has a concise rationale comment.

Validation: full optimized C++ suite (901 cases, 2.34M assertions), full Python suite (1,344 passed / 6 skipped), frozen pre-commit suite, WebAssembly build, and WebAssembly smoke tests. On this 8-CPU VM, d7 automatic throughput was 4.03 Mshots/s at one thread and 16.7 Mshots/s at eight; the aggregate cap does not reduce the usable eight-worker configuration.

bachase:
Addressed the final review and the adjacent pre-merge audit in three commits:

- `6765d6b5` prepares nonuniform fixed-k conditioning once per sampling call and shares the immutable table across scalar and packed workers. Per-worker selection scratch and survivor counters are included in automatic worker budgeting. The same change scales large DP rows before overflow; a 4,096-site, k=2,048 case that previously collapsed onto the final 2,048 sites now produces a mixed, exact-size conditioned sample.
- `48161555` keeps tentative categorical outcomes and assignments builder-local and moves them into `ExecutablePlan` only after the cost heuristic accepts the packed program, so rejected factorization capacity is released.
- `777b482f` replaces compact-carrier self-replay with an analytic categorical-noise oracle, asserts that compact carriers were selected, and checks scalar plus five packed capacities. It also removes the unnecessary lambda capture.

The broader audit also found that survivor counter vectors were outside the aggregate worker estimate and that the importance-sampling guide still showed the pre-batching function signatures; both are corrected.

Local validation at `777b482f`: 906/906 Debug C++ tests, 1,344 Python tests passed with 6 skipped, 33 documentation examples passed with 8 skipped, full pre-commit, WebAssembly build, and WebAssembly smoke tests. GitHub Actions remains unavailable, so no remote-check result is available yet.

Targeted fixed-k profile: 1,024 nonuniform sites, k=512, 16 shots, eight scalar workers, explicit batch size 1 fell from about 14.3 ms to 0.7 ms median because the DP table is no longer built eight times. The 4,096-site, k=2,048, eight-worker check peaked at about 119 MiB RSS with one shared table.

bachase:
Final cleanup and wide-audit update at `9509d9ad`:

- `ae2b913f` fixes the heterogeneous fixed-k regression. The prior max-odds normalization could underflow a reachable suffix coefficient and silently bias the sample. Preparation now produces direct conditional-selection probabilities and has independent coverage from an exhaustive subset oracle plus a grouped analytic high-stratum oracle.
- `9509d9ad` keeps the common preparation path fast: it uses the scaled multiply-add recurrence only when every reachable coefficient and non-forced branch remains finite and strictly representable, otherwise rebuilding in log space. On the prior 1,024-site, k=512, 16-shot, 8-worker profile, always-log setup measured about 5.9 ms median; the guarded path is about 1.0 ms while the adversarial case still takes the robust fallback.
- `b46452df` moves the policy contract into `batch/policy.h`, resolves complete fixed-k worker budgeting once per nonempty call, and removes the dead packed symbol indirection and state accessor.

I re-audited the full PR across plan semantics, lowering, worker construction, allocation policy, hot execution, scheduling/RNG, output ownership, and tests. The retained boundaries are deliberate: immutable `ExecutablePlan`, construction-time execution modes, conditional per-worker sidecars, one shared fixed-k distribution, per-worker mutable scratch, and call-owned outputs. I found no further redundant retained matrix, mode-dependent allocation, mutate/rollback preparation, worker-dependent capacity, hot allocation/exception, or self-replay-only correctness gap. The one intentional semantic duplication is the affine-outcome/record-parity representation already captured in #411; changing it crosses CPU scalar, packed CPU, and HIP and remains the right next architectural PR. Separate interleaved kernels and the carrier/expression optimizer remain justified by layout and measured performance.

Final local validation: 908/908 optimized C++ tests; 1,344 Python tests passed with 6 skipped; full frozen pre-commit; fresh WebAssembly build and smoke tests. The pushed head is clean and CI has restarted.

src/python/bindings.cc
```diff
@@ -41,21 +41,31 @@ namespace nb = nanobind;
...
     const int64_t count = std::get<int64_t>(option);
     if (count <= 0 || static_cast<uint64_t>(count) > std::numeric_limits<uint32_t>::max()) {
-        throw std::invalid_argument("threads must be a positive integer or 'auto'");
+        throw std::invalid_argument(std::string(field) + " must be a positive integer or 'auto'");
     }
     return static_cast<uint32_t>(count);
 }
 
+uint32_t parse_thread_option(const ThreadOption& option) {
+    return parse_positive_uint32_option(option, "threads").value_or(0);
```
bachase, line 64: Why is this 0 if unset but BatchOption parsing below isnt?

docs/guide/simulation.md
```diff
@@ -240,8 +240,75 @@ assert (r1.measurements == r2.measurements).all()  # Identical
 ```
 
 If `seed` is omitted or set to `None`, Clifft uses hardware entropy from the operating system.
-Each shot derives an independent random stream from the call seed and its shot
-index. Seeded results therefore do not depend on how shots are scheduled.
+Seeded results replay exactly when the program and execution configuration are
```
bachase: Let's rework this entire situation a little bit. 

First lets make it clear that the seed is meant for reproducibility in debugging and testing. Simulations and scientific results should rely on hardware entropy.

Second, this new section is still way too much detail. I would just say simply that this reproducibility is only within a sampling mode, e.g. random streams can differ if using batch sampling versus non batch sampling, but the same batch sample configuration reproduces the same stream. But the resulting samples should be statistically equivalent across all sampling modes.

docs/guide/simulation.md
```diff
@@ -240,8 +240,75 @@ assert (r1.measurements == r2.measurements).all()  # Identical
...
+Native automatic mode can select packed execution while WebAssembly always
+uses scalar execution. Consequently, identical calls using `batch_size="auto"`
+are not guaranteed to produce the same individual rows across native and
+WebAssembly builds. Upgrading from a release that predates packed automatic
+sampling can likewise change seeded rows. Repeating the call with the same
+program, seed, and execution configuration remains exactly reproducible.
+
+## Packed Batch Sampling
+
+The four fixed-plan sampling functions accept `batch_size`. Its default value,
```
bachase: I'd introduce batching a bit more clearly at the outside. "In a batched mode, clifft samples several shots at the same time using by packing their state data together to benefit from single-core SIMD instructions and cachce locality. This is in contrast to multi-threaded sampling which uses multiple CPU threads (ideally cores) to sample shots in parallel.  The two modes can be combined."

Not that verbatim, but a bit more context.

docs/guide/simulation.md
```diff
@@ -240,8 +240,75 @@ assert (r1.measurements == r2.measurements).all()  # Identical
...
+```python
+import clifft
+
+program = clifft.compile("M 0")
+result = clifft.sample(program, 100_000, seed=42, batch_size="auto")
+scalar = clifft.sample(program, 100_000, seed=42, batch_size=1)
+packed = clifft.sample(program, 100_000, seed=42, batch_size=1024)
+```
+
+An explicit positive integer requests a packed lane-capacity limit and is mainly
```
bachase: Let's also simplify this section. Its important to retain what circuit features/oeprations aren't supported, but I'd more clearly frame this limits and caps on size. So something like "As the maximum active_width k or the batch size grows, it will exceed the size of useful CPU caches and result in poor performance. `clifft`s auto mode attempts to select the ideal batch size based on the expected memory usage of teh particuarl sampling mode. Users can also explicitly set a size if (reasons). batch mode is disabled by default?"

src/clifft/util/shot_parallel.h
```diff
@@ -50,12 +50,21 @@ inline uint32_t resolve_shot_worker_count(uint32_t shots, uint32_t requested_thr
     return std::min(resolve_thread_budget(requested_threads), shots);
 }
 
-inline uint32_t shot_chunk_size(uint32_t shots, uint32_t workers) noexcept {
-    assert(workers != 0 && "shot chunking requires at least one worker");
+inline uint32_t shot_chunk_size(uint32_t shots, uint32_t workers,
+                                uint32_t minimum_chunk_size = 1) noexcept {
+    assert(workers != 0 && minimum_chunk_size != 0 &&
+           "shot chunking requires workers and a positive minimum chunk");
     constexpr uint64_t kTargetChunksPerWorker = 8;
```
bachase, line 59: this predates this PR but this parameter is a bit of a magic number .. I suppose its fine but how did we motivate 8 versus something else?

src/clifft/sampling/executable_plan.h
```diff
@@ -348,6 +397,14 @@ class ExecutablePlan {
     std::vector<PreparedNoiseOutcome> noise_outcomes_;
     std::vector<PreparedNoiseSite> noise_sites_;
     std::vector<double> noise_hazards_;
+    std::optional<double> uniform_noise_inverse_hazard_;
+    // Batch-only categorical factorization. It is either empty or parallel to
+    // noise_outcomes_; each assignment names a compact carrier column whose
+    // packed bit is set when that outcome is selected.
```
bachase: This might be better for the followup PR -- but consider reordering all these private members (not just the new ones) so its clearer for readers; right now they are a bit scattered and the comments make it confusing to follow. Be sure this doesn't impact locality/speed adversely.

src/clifft/sampling/executable_plan.h
```diff
@@ -361,6 +418,13 @@ class ExecutablePlan {
     // Optional debug sidecar parallel to actions_. It stays empty for ordinary
     // compilation and therefore adds no per-action production storage.
     std::vector<PlanActionRange> action_plan_ranges_;
+    // Optional batch-only tape for shared presampled expression prefixes.
```
bachase: For these presampled data, and the logic to build them in executable_plan_builder, would it maek sense to create a dedicated class/file (.h and .cc potentially) so they are isolated and easier to understand? Then you could locally document what they are/why they exist, and say include an example for future readers

src/clifft/sampling/executable_plan.h
```diff
@@ -348,6 +397,14 @@ class ExecutablePlan {
     std::vector<PreparedNoiseOutcome> noise_outcomes_;
     std::vector<PreparedNoiseSite> noise_sites_;
     std::vector<double> noise_hazards_;
+    std::optional<double> uniform_noise_inverse_hazard_;
+    // Batch-only categorical factorization. It is either empty or parallel to
```
bachase: Either here or above with the struct BatchNoiseOutcome, can we concisely document what these are doing in a less jargony way, maybe even with an example.

src/clifft/sampling/batch/bits.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <span>
+#include <vector>
+
+namespace clifft::sampling {
+
+// Fixed-capacity symbol-major bit columns used by packed cross-shot execution.
```
bachase: Can you make this less jargon/more standalone, or perhaps use an example (or ascci diagram of the storage)? But keep it concise

For example, is it essential or incidental that this happens to store symbols versus records? Or is this really just a compact bit matrix?

src/clifft/sampling/batch/policy.h
```diff
@@ -0,0 +1,108 @@
...
+#if defined(__EMSCRIPTEN__)
+// WebAssembly retains the scalar executor to minimize its binary footprint.
+inline constexpr bool kPackedBatchExecutionAvailable = false;
+#else
+// Native builds include packed execution and its interleaved kernels.
+inline constexpr bool kPackedBatchExecutionAvailable = true;
+#endif
+
+struct BatchExecutionPolicy {
+    // Stable number of shot lanes assigned to each packed batch.
```
bachase: Drop "stable" -- its clear this is just the fixed number of shot lanes.

src/clifft/sampling/batch/interleaved_state.h
```diff
@@ -0,0 +1,68 @@
...
+namespace clifft::sampling {
+
+// Return the SIMD-aligned lane stride used by interleaved allocations.
+[[nodiscard]] uint32_t interleaved_batch_lane_pitch(uint32_t lane_capacity);
+
+// Return the complete coefficient and measurement-scratch allocation size.
+[[nodiscard]] size_t interleaved_batch_state_bytes(uint32_t max_active_width,
+                                                   uint32_t lane_capacity);
+
+// Dense active-coordinate storage whose innermost dimension is the shot lane.
```
bachase: How about "state vector" or "amplitude" storage? That will make it clearer what this is for in a less jargon way.

src/clifft/sampling/batch/interleaved_state.h
```diff
@@ -0,0 +1,68 @@
...
+// Every shot in a batch follows the same prepared active-width transitions, so
```
bachase: I'm not sure this or the next comment is that helpful. Maybe just have the formula for the layout, e.g. where real/complex coefficient c of shot s in the batch are stored in the contiguous array, which is allocated at construction to store the full batch at the maximum active width.

src/clifft/sampling/batch/interleaved_state.h
```diff
@@ -0,0 +1,68 @@
...
+    void compact_lanes(std::span<const uint32_t> source_lanes) noexcept;
+
+    [[nodiscard]] uint32_t active_width() const noexcept { return active_width_; }
+    [[nodiscard]] uint32_t max_active_width() const noexcept { return max_active_width_; }
+    [[nodiscard]] uint32_t active_lanes() const noexcept { return active_lanes_; }
+    [[nodiscard]] uint32_t lane_capacity() const noexcept { return lane_capacity_; }
+    [[nodiscard]] uint32_t lane_pitch() const noexcept { return lane_pitch_; }
+    [[nodiscard]] uint64_t size() const noexcept { return uint64_t{1} << active_width_; }
+    [[nodiscard]] uint64_t capacity() const noexcept { return coefficient_capacity_; }
+
```
bachase, line 49: Should we define here at least once (or in the updated comment to the class) what basis means?

src/clifft/sampling/batch/interleaved_state.h
```diff
@@ -0,0 +1,68 @@
...
+    [[nodiscard]] double* real_basis(uint64_t basis) noexcept;
+    [[nodiscard]] const double* real_basis(uint64_t basis) const noexcept;
```
bachase: Should these implementations be in the header so they can be inlined?

src/python/bindings.cc
bachase: The distinction is intentional: the older C++ thread scheduler uses `0` as its internal auto sentinel, while batch selection needs `std::nullopt` to distinguish automatic policy from an explicitly requested lane capacity. I kept the public parsing shared and added a concise comment at the conversion point in c3ac3f96.

src/clifft/sampling/executable_plan.h
bachase: Addressed in 06a9ab24. `BatchPresampledProgram` now lives in dedicated `.h` and `.cc` files, owns both packed categorical factorization and shared-prefix expression evaluation, and is an optional value so absence represents the ordinary fallback by construction.

src/clifft/sampling/batch/bits.h
bachase: Addressed in c3ac3f96. `PackedBitColumns` is now documented as a generic dense bit matrix, with a small column/lane diagram and the exact word index formula. Symbols, records, and outputs are incidental users of the same storage.

docs/guide/simulation.md
bachase: Addressed in c3ac3f96. The guide now explains exponential active-width and linear lane scaling, the auto policy and its per-worker/aggregate memory targets, explicit override use, and a concise list of unsupported combinations. `auto` remains the default; `batch_size=1` explicitly disables batching.

docs/guide/simulation.md
bachase: Addressed in c3ac3f96. The section now opens by distinguishing packed single-core SIMD/cache parallelism from multithreaded worker parallelism and explains that Clifft can combine them.

src/clifft/sampling/batch/interleaved_state.h
bachase: Addressed in c3ac3f96. The class comment defines basis `b` as the computational-basis bit pattern over the current active coordinates.

src/clifft/sampling/batch/interleaved_state.h
bachase: Addressed in c3ac3f96. The class comment now gives the concrete layout formulas `real_[b * lane_pitch_ + s]` and `imag_[b * lane_pitch_ + s]`, then explains that construction reserves the maximum active width.

src/clifft/sampling/executable_plan.h
bachase: Addressed with the new `BatchPresampledProgram` header in 06a9ab24. Its class comment gives concrete A/B downstream-effect and shared-expression-prefix examples, explains compact carriers in terms of XOR effects, and states the fallback semantics.

docs/guide/simulation.md
bachase: Addressed in c3ac3f96. The guide now frames explicit seeds as a debugging/testing replay aid, recommends OS entropy for independent production simulations, and limits the guarantee to the same sampling mode and configuration while stating statistical equivalence across modes.

src/clifft/util/shot_parallel.h
bachase: I kept the value because this is a bounded scheduling heuristic rather than part of the packed algorithm. The new comment records the tradeoff: several claims per worker redistribute uneven shot costs without paying one atomic claim per shot. Eight targets enough redistribution headroom at low scheduler overhead; changing it should be driven by a separate cross-machine scheduler benchmark.

src/clifft/sampling/batch/interleaved_state.h
bachase: Addressed in c3ac3f96. The trivial basis accessors are now defined in the header, preserving their assertions while allowing normal inlining.

src/clifft/sampling/executable_plan.h
bachase: Addressed in c3ac3f96. The members are now grouped as metadata, affine registers, presampled inputs, packed output parity storage, instrument state, and the hot action stream with its prepared payloads. The hot action data remains adjacent, so this is a readability regrouping rather than a hot-layout change.

src/clifft/sampling/batch/interleaved_state.h
bachase: Addressed in c3ac3f96. The class is now described as dense complex state-vector storage for a batch of shots, and its arrays as amplitude and measurement-scratch storage.

src/clifft/sampling/batch/policy.h
bachase: Addressed in c3ac3f96: `stable` was removed. The comment now says this is the number of shot lanes in a full packed batch.

src/clifft/sampling/batch/presampled_program.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace clifft::sampling {
+
+class BatchExecutor;
+class ExecutablePlan;
+class ExecutablePlanBuilder;
+class SamplingPlan;
+
+// Immutable packed-only program for evaluating presampled affine inputs.
+// Categorical outcomes set XOR combinations of compact carrier columns based
```
bachase: This is a nice example, but still kind of jargony. Can you reframe it in terms of operations or actions that clifft supports?

PR #410, opened 2026-08-26
feat(api): add native unitary OpenQASM 2 input

bachase:
## Summary

- add a dependency-free native OpenQASM 2.0 parser and lowerer for the documented unitary subset
- expose explicit `parse_qasm2()`, `parse_qasm2_file()`, and `compile(..., input_format="qasm2")` Python entry points without format sniffing
- preserve OpenQASM-to-Clifft Euler-gate phase corrections in a `Qasm2Import.global_phase_half_turns` sidecar, while keeping ordinary `Circuit`, HIR, plans, and executor state phase-free
- preserve declared register widths, implement register broadcasting and Qiskit-compatible constant-expression precedence, and tolerate UTF-8 in comments
- document the Qiskit phase convention selected where the OpenQASM 2 specification is ambiguous
- validate the complete gate vocabulary against Qiskit and confirm all 196 current ABSTRACTS QASM circuits parse successfully

The advertised contract is OpenQASM 2 only. OpenQASM 3 is intentionally deferred: its `stdgates.inc` semantics, `U` phase convention, gate modifiers, `gphase`, declarations, and dynamic-circuit surface deserve a separately specified subset instead of an accidental compatibility claim.

Measurements, resets, classical conditions, and custom/opaque gate declarations are rejected initially. Measurement/reset import for Clifft sampling can be added independently; this PR does not assign an implicit meaning to non-unitary amplitudes. A bounded custom-gate macro expander, including an embedded source definition of the remaining unitary `qelib1.inc` vocabulary, is the designated unitary-format follow-up.

The Release fast-math hardening was reproduced with Apple Clang 17 on arm64. Finite and probability checks now use inline integer validation behind a zero-instruction optimizer barrier on native Clang/GCC, with a volatile fallback for other toolchains. This preserves nonfinite validation without adding a function-call boundary to existing code paths.

## Test plan

- [x] Debug C++ tests pass (`ctest --test-dir build -E Bench --output-on-failure`): 876 passed
- [x] Release OpenQASM C++ tests pass with `-ffast-math`: 11 passed
- [x] Release numeric validation tests pass with `-ffast-math`: 2 passed
- [x] Python tests pass (`uv run pytest tests/python/ -q`): 1,347 passed, 7 skipped, 1 xfailed
- [x] Qiskit componentwise phase reconstruction passes for `U`, `u1`, `u2`, `u3`, and register broadcast without global-phase alignment
- [x] Doc examples pass (`uv run pytest --codeblocks docs/ README.md -v`): 34 passed, 8 skipped
- [x] Strict MkDocs build passes
- [x] All 196 QASM files in `mjsutcliffe99/ABSTRACTS` parse with the native importer

## Contributor agreement

- [x] I confirm this contribution is my original work (or I have the right to submit it) and I license it under the project's Apache-2.0 license.
- [ ] If this contribution was AI-assisted, I have reviewed the generated code and included an `Assisted-by:` git trailer in the commit message.

The required `Assisted-by:` trailer is present. The remaining checkbox is left for human review before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,10 @@
 
 Built and maintained by the [Unitary Foundation](https://unitary.foundation).
 
-Clifft accepts Stim-format circuits, extends them with non-Clifford gates, and
-compiles them into symbolic-coordinate sampling plans. It is designed for
-circuits whose dominant structure is Clifford, but whose behavior depends on
-localized non-Clifford operations.
+Clifft accepts Stim-format circuits with non-Clifford extensions and a native
+unitary OpenQASM 2 subset, then compiles them into symbolic-coordinate sampling
+plans. It is designed for circuits whose dominant structure is Clifford, but
+whose behavior depends on localized non-Clifford operations.
 
 The dense active state has `2^k` amplitudes, where `k` is its active width. The
 main simulation cost therefore scales with `2^k`, rather than directly with the
@@ -51,8 +51,9 @@ benchmarks.
 
 ## Why Clifft?
 
-- **Stim-compatible format and API**: parse Stim-format circuits with noise,
-  detectors, observables, and repeat blocks, plus non-Clifford extensions.
+- **Native circuit formats**: parse Stim-format circuits with noise, detectors,
+  observables, and repeat blocks, plus unitary OpenQASM 2 input and non-Clifford
+  extensions.
 - **Exact near-Clifford simulation**: simulate localized non-Clifford effects
   without approximating the quantum state.
 - **Optimizing compiler pipeline**: resolve Clifford coordinates and symbolic
--- a/README.md
+++ b/README.md
@@ -23,10 +23,10 @@
 
 Built and maintained by the [Unitary Foundation](https://unitary.foundation).
 
-Clifft accepts Stim-format circuits with non-Clifford extensions and a native
-unitary OpenQASM 2 subset, then compiles them into symbolic-coordinate sampling
-plans. It is designed for circuits whose dominant structure is Clifford, but
-whose behavior depends on localized non-Clifford operations.
+Clifft accepts Stim-format circuits, extends them with non-Clifford gates, and
+compiles them into symbolic-coordinate sampling plans. It is designed for
+circuits whose dominant structure is Clifford, but whose behavior depends on
+localized non-Clifford operations.
 
 The dense active state has `2^k` amplitudes, where `k` is its active width. The
 main simulation cost therefore scales with `2^k`, rather than directly with the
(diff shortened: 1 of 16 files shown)


docs/getting-started/openqasm2.md
bachase, line 1: DO we really need this whole page? Or if we do, can we make it appear as a subpage of front-end integrations? It just feels confusingly placed.

docs/getting-started/openqasm2.md
```diff
@@ -0,0 +1,89 @@
+# Native OpenQASM 2 Input
+
+Clifft can natively parse and compile a unitary subset of OpenQASM 2.0. This
+path has no Qiskit runtime dependency and is intended for unitary simulation
+and exact-query workflows such as the
```
bachase: We don't need to link to other tools/motivations. The point is your first sentence or two -- clifft can parse subset of openqasm2.

docs/getting-started/openqasm2.md
```diff
@@ -0,0 +1,89 @@
...
+
+Declared register width is preserved even when some qubits are unused.
+Register-valued gate operands are expanded using OpenQASM broadcasting rules.
+
+The initial contract rejects classical registers, measurements, resets,
+classical conditions, custom gate declarations, opaque declarations, and
+nonstandard include files. These statements fail during parsing rather than
+being dropped or assigned approximate semantics.
+
+Custom gate declarations are the intended next unitary-format extension. A
```
bachase: We don't need to specifiy what is next (or even openQASM3 below). We can just have the sentence above about what is and isn't supported, and then a comment that support may expand in the future.

docs/getting-started/openqasm2.md
```diff
@@ -0,0 +1,89 @@
...
+
+source = "OPENQASM 2.0; qreg q[1]; U(0, 0, 0) q[0];"
+imported = clifft.parse_qasm2(source)
+print(imported.circuit)
+print(imported.global_phase_half_turns)
+```
+
+`global_phase_half_turns` is a value `t` representing the correction
+`exp(1j * pi * t)`. OpenQASM 2's Euler gates and Clifft's internal `U3`
+representation can differ by this phase. Keeping it beside the ordinary AST
```
bachase: I'd just frame this as, "clifft is generally insensitive to global phase, but we retain it here for poential future use cases"

docs/getting-started/openqasm2.md
```diff
@@ -0,0 +1,89 @@
...
+lets future phase-sensitive compilation consume the exact source convention
+without adding scalar bookkeeping to sampling HIR, plans, or executor state.
+
+For the ambiguity between the OpenQASM 2 specification's displayed `U` matrix,
```
bachase: Just say what we do without the preamble -- "For gate phase convention, Clifft follows Qiskits ...."

docs/getting-started/openqasm2.md
bachase: Or consider taking the last paragraph where you make that point and moving up here.

README.md
bachase, line 1: Don't elevate QASM@ to the README.md here. If anything it falls under the "other supported frontends" type approach that we have for cirq or qiskit from a functionality perspective. IF that means we don't mention it in the README, that is fine.

PR #3, opened 2026-04-16
fix: playground link serves docs page instead of SPA

bachase:
## Summary

- Remove `docs/playground.md` — the standalone docs page about the playground is redundant since the SPA itself is self-explanatory
- Fix `cp -r` in `docs.yml` and `pr_preview.yml` to replace the MkDocs-generated `site/playground/` before copying the SPA, so the playground `index.html` isn't shadowed by the docs page
- Remove the nav entry from `mkdocs.yml`

The header pill button continues to link to `playground/`, which now correctly serves the SPA.

## Test plan

- [x] PR preview playground link loads the interactive SPA, not a docs page
- [ ] After merge, verify `unitaryfoundation.github.io/clifft/playground/` serves the SPA

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #412, opened 2026-08-26
feat: add target-aware basis amplitude queries

### LINKED ISSUE #408: Add phase-aware exact computational-basis amplitude queries
## Problem or motivation

Clifft exposes exact computational-basis probabilities through
`basis_probabilities()` and exact measurement-record probabilities through
`record_probabilities()`. It does not expose the corresponding complex
computational-basis amplitudes.

`get_statevector()` is not a substitute for this query:

- it returns a state representative only up to global phase;
- it materializes the full physical `2^n` statevector;
- it cannot use a requested output basis state as a terminal effect during
  planning.

Complex amplitudes are useful for circuit verification, scalar tensor-network
contractions, composition with other exact tools, and amplitude-oriented strong
simulation benchmarks. In particular, ABSTRACTS asks simulators to return
`<0...0|C|0...0>` rather than only its squared magnitude:
https://github.com/mjsutcliffe99/ABSTRACTS

## Proposed solution

Add a public exact query for selected computational-basis amplitudes. The most
natural user-facing name is `basis_amplitudes`, parallel to
`basis_probabilities`:

```python
amplitudes = clifft.basis_amplitudes(program, ["000", "111"])
```

The result should be a one-dimensional `numpy.complex128` array in query order.
Input validation and bitstring conventions should match `basis_probabilities()`.

Because a known output bitstring can materially reduce active width, the design
should also provide a query-aware compilation route rather than requiring every
amplitude to use a plan compiled without its terminal effect. The exact API can
be settled during implementation; one possible shape is:

```python
query = clifft.compile_basis_amplitude(circuit_text, output_bits="0" * n)
print(query.peak_active_width)
amplitude = clifft.evaluate_amplitude(query)
```

A single-target query plan is sufficient initially. A batch convenience API can
reuse an existing plan where appropriate or compile target-specific plans when
that reduces width.

## Semantic contract

Initial support should match the pure-unitary eligibility of
`basis_probabilities()`:

- the initial state is `|0...0>`;
- the requested output is a complete computational-basis effect;
- measurements, resets, feedback, and noise are rejected;
- every accepted unitary gate has a documented canonical matrix convention,
  including global phase;
- the returned value is the complex scalar `<x|U|0...0>` under those
  conventions.

For every supported query:

```python
abs(clifft.basis_amplitudes(program, [x])[0]) ** 2
```

should agree with:

```python
clifft.basis_probabilities(program, [x])[0]
```

within the documented floating-point tolerance.

## Implementation expectations

- Track the required complex scalar through Clifford absorption, active-coordinate
  basis changes, canonicalization, and coordinate elimination.
- Preserve phase when a query-aware terminal effect forces projections or removes
  completed coordinates.
- Reuse the selected-basis query machinery where practical; do not expand the
  full physical statevector.
- Perform tableau evolution, effect propagation, coordinate planning, and phase
  dependency analysis during compilation/planning, not in executor kernels.
- Preserve allocation-free and exception-free ordinary hot execution.
- Expose `peak_active_width` for a query-specific plan before evaluation.
- Leave the behavior and return types of `basis_probabilities()`,
  `record_probabilities()`, and `get_statevector()` unchanged.

## Validation

- Analytic tests covering real, imaginary, negative, and zero amplitudes.
- Componentwise comparison against Qiskit Aer for small random Clifford+T
  circuits; fidelity-only comparisons are insufficient.
- Tests for canonical global phases of Clifford gates, T/T-dagger, square-root
  gates, and supported Pauli rotations.
- Randomized checks that `abs(amplitude) ** 2` agrees with
  `basis_probabilities()`.
- Regression tests for phase changes introduced by Clifford-frame updates,
  coordinate changes, and terminal-effect elimination.
- Tests showing that query-aware terminal effects can reduce
  `peak_active_width` without changing the returned amplitude.
- CPU backend consistency tests where multiple kernel implementations apply.

## Acceptance criteria

- A documented public Python API returns exact selected basis amplitudes as
  `complex128` values.
- Canonical gate/global-phase conventions are documented and independently
  validated.
- The implementation does not require full-statevector materialization.
- A query-specific plan can use the output effect during planning and reports its
  `peak_active_width`.
- Existing exact probability and statevector APIs retain their current contracts.
- Relevant C++, Python, documentation, and pre-commit checks pass.

## Non-goals

- ZX-calculus or phase-polynomial preprocessing.
- Adding PyZX as a Clifft dependency.
- Noisy-channel amplitudes, arbitrary Kraus operators, or POVMs.
- Defining scalar amplitudes for partial measurement records with unclosed quantum
  outputs.
- Symbolic amplitudes.
- Changing `get_statevector()` from its current up-to-global-phase contract.

## Alternatives considered

Return a signed or complex-valued variant of `basis_probabilities()`. This would
mix two different mathematical objects: probabilities are non-negative real
numbers, while amplitudes are complex and phase-sensitive. A separate
`basis_amplitudes` query makes the contract explicit.

Require callers to use `get_statevector()` and select one coefficient. This loses
the circuit's global phase contract and requires `O(2^n)` output even when only a
single scalar is requested.

bachase:
## Summary

- add `basis_amplitude()` as the one-call API for one exact complex computational-basis amplitude, with `compile_basis_amplitude()` plus `query.evaluate()` for feasibility-sensitive callers
- compile for the selected complete output bitstring, allowing `StatevectorSqueezePass` to provide the same target-aware active-width reduction as terminal-measurement `record_probabilities()`
- retain canonical global phase through Clifford absorption, target-aware measurement branches, coordinate changes, discarded scalar rotations, and the OpenQASM 2 phase sidecar
- remove the previous forward/adjoint orientation selection and the now-unused whole-state selected-amplitude walk
- fix an existing `basis_probabilities()` stale-sign bug during Z-block elimination, which could swap probabilities after a later pivot updated an earlier row
- keep the existing probability, statevector, planner API, and ordinary hot executor paths unchanged; phase-aware compilation and replay are query-private and opt-in

Internally, the selected output is enforced with terminal computational-basis measurements. Exact replay preserves small positive amplitudes below the ordinary sampling dust threshold, while the target-aware exact Clifford frame composes each forced branch before later coordinate changes. ZX/phase-polynomial preprocessing and a PyZX dependency remain out of scope.

Closes #408.

## Performance check

Matched local `RelWithDebInfo` builds were compared against the then-current `origin/main` with alternating run order:

- QV20 sampling at active width 20: +0.19% median
- surface-code d7 sampling at active width 0 over 10,000 shots: +0.13% median
- end-to-end existing compilation: approximately +0.1% to +1.5% across three fixtures

For the terminal-measurement revision, a second matched comparison against the previous PR head kept the same 20-qubit QV plan, 3,600 actions, and deterministic checksum. Alternating 30-run samples differed by about +0.3% in the means, within host variation. After the forced-branch phase fix, the optimized ordinary executor and kernel disassemblies remain byte-for-byte identical to the prior revision after normalizing the object header. Exact replay is a separate query-private executor mode.

## Test plan

- [x] Full C++ suite on the current head: 951 passed, excluding benchmark cases
- [x] Full rebuilt Python suite: 1,421 passed, 6 skipped, 1 expected failure
- [x] Focused Python amplitude tests: 55 passed
- [x] Selected amplitude and OpenQASM documentation examples: 10 passed
- [x] 300 randomized OpenQASM 2 circuits with arbitrary rotations match Qiskit componentwise without phase alignment
- [x] Deterministic regressions cover persistent three-qubit entanglement and anti-correlated Bell outputs across the 64-bit mask boundary
- [x] Optimized profiler build completes without a new release warning
- [x] All fixed Clifford matrices match Stim componentwise, including global phase
- [x] Regressions cover forced measurement branches, target-driven coordinate reactivation, multiple planner coordinate changes, and amplitudes whose squared magnitude is below the sampling dust threshold

## Contributor agreement

- [ ] I confirm this contribution is my original work (or I have the right to
  submit it) and I license this contribution under the
- [ ] If this contribution was AI-assisted, I have reviewed the generated code


bachase (commented):
Review of `0327ac29`. Findings below were independently verified against the PR head: full Debug C++ suite, a Debug Python wheel, a wasm docker build, ~4000 randomized differential circuits against dense statevector references, plus targeted probes and mutation tests. The core CH-form algebra, planner frame composition, and executor replay all checked out clean under that fuzzing — everything below is what did *not* check out, ordered by severity.

## Correctness (reproduced)

**1. Unreduced rotation angle feeds `std::polar`, producing wrong phases for large Clifford-classified angles.**
`ExactTraceState::multiply_source_scalar` (`src/clifft/frontend/frontend.cc:43`) computes the scalar from the raw half-turn angle, while `classify_clifford_rotation` reduces it mod 2 with exact `fmod`. The two disagree once the angle is large enough that `pi * alpha / 2` loses precision. Reproduced: `compile_basis_amplitude` on `H 0` / `R_Z(1000000000000001) 0`, target `0`, returns `(-0.0424, -0.7058)` instead of the exact `-i/√2 = (0, -0.7071)` — a 0.06 rad phase error (already 6.5e-8 at `alpha=1000000001`). The same unreduced `std::polar` appears in the planner's discarded-scalar paths (`src/clifft/sampling/planner.cc:399` and `:915`; repro: `R_Z(1000000000000000.5)` returns `(0.808, -0.590)` instead of `(0.707, -0.707)`). Fix: reduce half-turns with exact `std::fmod(alpha, 4.0)` before `std::polar` at all three sites.

**2. Duplicate two-qubit targets reach the CH form unchecked; `SQRT_XX 0 0` silently returns non-normalized amplitudes.**
clifft's parser accepts duplicate pair targets (Stim rejects them), and nothing between the parser and `StabilizerChForm` checks distinctness. In `apply_pair_rotation` (`src/clifft/tableau/stabilizer_ch_form.cc:465`), `set_pauli` is called twice on the same qubit, degenerating the axis: on a Release wheel, `H 0` / `SQRT_XX 0 0` returns `0.3536+0.3536j` for **both** |0⟩ and |1⟩ — |a|² sums to 0.5, and disagrees with `basis_probabilities`' `[0.5, 0.5]` for the identical circuit. `H 0` / `CZ 0 0` instead leaks the internal message `CH-form CZ requires two distinct in-range qubits` from the amplitude API only. (Debug builds abort earlier in a pre-existing trace assert, so CI never sees the Release behavior.) Duplicate-target rejection belongs at the parser/trace level so every API surface agrees.

**3. `exp(0.5 * log_probability)` throws away dyadic exactness and eventually underflows to an ambiguous exact zero.**
`evaluate()` (`src/clifft/api/basis_amplitudes.cc:109`) reconstructs the magnitude in log space, but every dormant terminal measurement contributes an exactly-dyadic `kLogHalf`. Measured on all-H circuits, target |0…0⟩ (exact amplitude 2^(−n/2)): relative error 4.2e-15 at n=64, 7.2e-12 at n=1024, 3.5e-11 at n=2048 — past the ~1e-12 precision the API demonstrates elsewhere — and at n=2200 the result is exactly `(0,0)`, indistinguishable from the `!replay.reachable` return, with no diagnostic and no log-magnitude accessor. Counting the halvings in an integer and finishing with `std::ldexp` would be exact; exposing `replay.log_probability` (analogous to `return_log=True` on `basis_probabilities`) would cover the underflow regime.

**4. The new raw-`Circuit` validator is inconsistent — it rejects some malformed shapes but segfaults or misreports on others.**
`append_terminal_measurements` (`src/clifft/api/basis_amplitudes.cc:38`) rejects PAIR/TRIPLE arity mismatches and non-unitary gates, but (a) a hand-built `AstNode{.gate=R_Z, .targets={qubit(0)}, .args={}}` SIGSEGVs in `trace_impl` (`frontend.cc:460` reads `node.args[0]` unchecked), and (b) `{H 0; CX rec(0) 1}` passes the `is_unitary` check and dies later with the confusing planner message `sampling planner operation 0 reads record 0 before assignment` instead of the intended pure-unitary error. Parsed circuits are safe; the C++ entry point isn't. Since the new arity check also papers over `trace`'s silent odd-target truncation only for this one API, these checks would be better placed at the `Circuit`/trace level where all consumers benefit.

## Test gap (mutation-proven)

**5. The sub-dust branch retention — the entire reason `force_active_effect_branch` exists — has zero coverage.**
Replacing the `force_active_effect_branch` call at `src/clifft/sampling/executor.cc:461` with plain `force_active_branch` still passes the whole Debug suite (936 cases, 2.35M assertions). The comment at `tests/test_basis_amplitudes.cc:235` claiming to cover it sits on a test whose `peak_active_width()` is 0 — its 2^-30 amplitude comes entirely from 60 dormant `kLogHalf` additions and never touches dust classification. A future regression folding `ReplayEffects` back onto the dust-clamping path would silently return 0 for any amplitude whose active branch mass is below 1e-18 of total, with the suite green.

## Robustness

**6. The active-width contract in `evaluate()` is assert-only, then coefficient `[0]` is read unconditionally.**
`src/clifft/api/basis_amplitudes.cc:106` — shipped wheels are NDEBUG, so if any future squeeze-pass/planner change leaves `active_width > 0` after effect replay, Release builds return `phase * sqrt(p) * state[0]` (one coordinate of a 2^w-wide state) as the exact amplitude with no diagnostic, while Debug CI aborts. The invariant currently holds (thousands of randomized queries never tripped it), but this is a cold path — a thrown check would fail loudly in both build types.

## Maintainability / performance

**7.** `src/clifft/frontend/frontend.cc:578` — `PhaseAwareCliffordFrame::compose_input` front-inserts into `operations_`, once per terminal measurement / dormant promotion / forced branch, each shifting the entire recorded op log: O(n·(G+n²)) variant moves (hundreds of millions at ~2000 qubits × 200k gates), stalling compile before `evaluate()` runs. `tableau()` also pays a full `then(Tableau::from_pauli_rotation(...))` product — O(n³/64) plus an allocation — per recorded rotation, though a reverse walk with the existing O(n²/64) `prepend_named_gate`/`prepend_pauli_rotation` builds the same tableau. Append to a back-growing input log replayed in reverse; the staging `input` vector inside `compose_input` is also unnecessary (insert the span directly).

**8.** `src/clifft/sampling/planner.cc:591` — `plan_sampling_impl` threads five all-or-nothing nullable side-channel parameters through `process_rotation`/`process_measurement`, with the pairing invariant enforced only by an assert inside the `discarded_scalar` branch (a some-but-not-all caller dereferences null in Release, only when a symbol-dependent identity rotation occurs). This PR already solves the identical optional-observer problem in `frontend.cc` with the zero-cost `TraceState` policy template — the same pattern here would make the contract structural and remove the new per-op null tests from the ordinary compile path.

**9.** `src/clifft/sampling/planner.cc:35` — the `compose_*` gate lists must byte-match the `CoordinateFrame` kernels in `planner_frame.cc` (swap order, pivot, H/S placement), and the plan-time forced branch at `planner.cc:448` must stay in lock-step with the executor's correction convention built in `executable_plan_builder.cc:260` — all tied together only by comments. A change to either side silently produces wrong phases or diverged branch selection with nothing structural prompting the mirror update.

**10.** `src/clifft/sampling/executor.cc:457` — `ReplayEffects` differs from `ReplayRecords` by exactly one call, but is wired as a `Mode == ReplayRecords || Mode == ReplayEffects` disjunction repeated at six dispatch sites. A future handler testing only `ReplayRecords` compiles cleanly and silently routes effect replay into the sampling branch (consuming RNG → nondeterministic "amplitude"). A single `is_forced_replay(Mode)` predicate — or making the dust policy a parameter instead of a fifth mode — removes the hazard.

**11.** `src/clifft/api/basis_probabilities.cc:419` — the `Output`/`CoefficientAt`/`Finish` template generalization of the probability walk is leftover scaffolding: its only other consumer was deleted later in this same PR, leaving one instantiation (`double` + `std::norm`) and a stale comment at line 631 justifying `clifford_row_phase`'s `conj()` via a walk the amplitude query never uses (it gets its scalar from `Executor::replay_effect`). Reverting to the concrete form restores the name matching `docs/theory/basis_probabilities.md`.

**12.** `src/clifft/sampling/planner.cc:62` — the `if (promoted.sign())` Z-correction arm of `dormant_promotion_operations` is unreachable: `resolve_pauli` unconditionally clears the body sign and both call sites pass `resolved.body` (instrumented full-suite run: 0 hits, while a neighboring branch hit 73 times). It makes signed promoted axes look handled-and-tested here when the sign actually lives in the runtime `AffineBool`. Drop it or exercise it.

**13.** Duplication: the ceil(n/64)-words + no-unused-high-bits basis validator is restated three times in this PR (`src/clifft/api/basis_amplitudes.cc:22`, `frontend.cc:606`, `stabilizer_ch_form.cc:703`) with three different error strings, while `api/basis_probabilities.cc:55/75` already has `basis_word_count`/`mask_has_only_qubit_bits` and `util/mask_view.h` has `bit_get`/`bit_set`/`popcount`; `stabilizer_ch_form.cc:16` also adds a third private `i_pow` and a fifth `kInvSqrt2`. One shared validator + `i_pow` next to `MaskView` would do.

**14.** `src/clifft/frontend/phase_aware_frontend.h:57` — `trace_phase_aware` (the reject-all-measurements variant) is exported public API with zero callers and zero test coverage; only `trace_phase_aware_terminal_measurements` is used. Delete until needed, or cover it.

**15.** Both new C++ test files re-implement `check_complex` and `kInvSqrt2` from `tests/test_helpers.h` (losing the per-component `WithinAbs` diagnostics), and `tests/python/test_basis_amplitudes.py:96` hand-builds the stim→Qiskit mirror and AerSimulator boilerplate that `tests/python/utils_qiskit.py` provides — and that the sibling `test_basis_probabilities.py` already imports.

## Suggested additional tests

- **Sub-dust retention** (closes #5): a circuit with an *active* measurement whose selected branch mass is below `kMeasurementDustEpsilon` relative to total — e.g. drive a qubit to amplitude ~1e-10 with non-Clifford rotations, measure it, and query the amplitude through the tiny branch. Assert the amplitude is nonzero; this fails under the `force_active_branch` mutation.
- **Large-angle phases** (closes #1): amplitude checks for `R_Z(4k+1)` and half-integer angles with `k` around 1e15, compared to the reduced-angle equivalent — exact equality after the `fmod` fix.
- **Wide-circuit dyadic exactness** (closes #3): all-H at n ∈ {64, 512, 2048}, asserting the amplitude equals `2^(-n/2)` exactly (`ldexp`-constructed), and an explicit-behavior test for the underflow regime once a log accessor exists. The current Aer comparison only exercises n ≤ 4.
- **Duplicate-target rejection** (closes #2): once rejection lands, `SQRT_XX 0 0` / `CZ 0 0` / `ISWAP 0 0` should raise the same parse/validation error from *both* `basis_probabilities` and `compile_basis_amplitude`.
- **Cross-surface consistency fuzz**: randomized circuits asserting `|compile_basis_amplitude(b)|²` matches `basis_probabilities`' value for the same `b` — this would have caught #2 (and guards #6) cheaply on every CI run.

bachase:
Addressed in 3b003093.

What changed:

- Reduced half-turn angles modulo 4 before multiplying by pi in the phase-aware frontend, planner scalar paths, and prepared rotation/promotion coefficients. This fixes both very large Clifford-classified angles and large non-absorbed rotations.
- Moved raw Circuit structural validation to the common trace boundary. Incomplete arities, duplicate pair/triple or Pauli targets, missing rotation arguments, invalid record placement/references, parser-only nodes, and out-of-range targets now fail before tracing. The ordinary compile and amplitude-query surfaces therefore reject the same malformed circuits.
- Kept dormant 1/2 effect factors as an integer count and applies them with ldexp, so representable dyadic amplitudes remain exact. Query preparation now also verifies in Release builds that terminal effects eliminate every active coordinate.
- Made PhaseAwareCliffordFrame input composition append-only and reconstructs the operator with reverse prepend operations. This removes repeated front insertion and the per-rotation full tableau product.
- Replaced the planner five-pointer side channel with one phase-aware context, centralized the forced-replay mode predicate, removed the unreachable promoted-sign correction, deleted the unused trace_phase_aware entry point, and restored the probability walk to its concrete probability-only form.
- Consolidated mask-width/high-bit checks and powers of i, reused the existing C++ and Qiskit test helpers, and made CH-form logical-bit comparison robust to padding bits.

Added coverage for:

- very large integer and half-integer rotations;
- duplicate SQRT_XX, CZ, and ISWAP targets across both strong-simulation APIs;
- malformed raw Circuit nodes and future record references;
- an actually active sub-dust effect branch;
- bit-exact all-H amplitudes at 64, 512, and 2048 qubits;
- phase-frame input composition order;
- randomized componentwise Aer comparisons plus abs(amplitude)^2 versus basis_probabilities on every generated circuit;
- an opt-in 256-qubit phase-aware compilation benchmark.

Local verification:

- 946 non-benchmark C++ tests passed;
- 1,418 Python tests passed, with 6 skipped and 1 expected failure;
- optimized Release build completed without new warnings from these changes;
- all pre-commit formatting, lint, typing, and repository checks passed;
- the new compile benchmark measured about 70 ms at 256 qubits on this machine.

Deliberately not included:

- A public log-amplitude or log-magnitude API. The integer dyadic fix removes avoidable precision loss for every representable result, but binary64 must still underflow for sufficiently tiny nonzero amplitudes. A log-valued API needs its own naming/return-contract decision and is better handled as a separate feature.
- A policy-templated duplicate of the whole planner. The single optional context makes the all-or-nothing contract structural with less code and keeps ordinary planning as one implementation.
- A shared implementation between CoordinateFrame kernels and the phase-frame mirror. Unifying those representations would be a much larger topology-planning change; focused composition-order and randomized cross-surface tests provide protection without adding a second abstraction.
- The additional CH-form and dual-orientation calibration micro-optimizations from the review. They are independent performance work, while the concrete asymptotic front-insertion issue is fixed and now has an opt-in benchmark.

bachase:
Addressed the independent follow-up review in f5d467d7.

What changed:

- Restored the public ReplayResult aggregate to its original two fields. Effect-only dyadic metadata now stays in a private Executor::ExecutionResult, and a two-element structured-binding regression test protects the source contract.
- Replaced the batch-oriented Clifford-row calibration with a query-private selected-effect phase lookup. The amplitude query already knows the terminal output basis and terminal-effect planning leaves the zero virtual basis as the sole survivor, so it obtains the exact row phase directly from that known output. This also removes PhaseAwareSamplingPlan::final_tableau and the independently mutable duplicate state.
- Replaced per-operation target vectors with a compact validated two-slot representation. Unsupported gates and wrong arities now fail at construction.
- Added a singular Python BasisBitstring contract. compile_basis_amplitude now accepts a string or one-dimensional bool/uint8 NumPy array and rejects batch-shaped input at the Python boundary.
- Expanded the opt-in compile benchmark to 256, 1024, and 2048 qubits, and added coverage for unreachable selected effects, invalid compact operations, singular NumPy inputs, and ReplayResult structured binding.

The direct selected-effect path removed the cubic calibration and improved same-machine Release compile measurements from about 1.46 to 1.14 ms at 256 qubits, 28.1 to 22.1 ms at 1024, and 152.6 to 128.7 ms at 2048. The remaining scaling is in the rest of query compilation, not a second batch amplitude-structure construction.

I also prototyped exact inverse-operation cancellation in the phase frame. It did not produce a measurable improvement on this workload, so I removed it rather than retain another optimization mechanism and its invariants. I likewise did not add a general phase-propagation subsystem: the selected-effect anchor is narrower, faster, and uses information this query already owns.

Local verification on the final tree:

- 947/947 non-benchmark C++ tests passed;
- the focused Python amplitude suite passed 55/55, and the earlier full Python run passed 1,421 tests with 6 skipped and 1 expected failure;
- all formatting, lint, typing, workflow-lint, and repository pre-commit checks passed;
- git diff --check passed.

bachase:
Follow-up cleanup in d43b8598:

- Removed the now-unused PhaseAwareCliffordFrame::tableau reconstruction. The composition-order test now exercises inverse_on_basis, the production amplitude path, across all physical and virtual three-qubit basis states.
- Removed the dead allow_terminal_measurements mode and its unreachable error branch; the remaining query-private tracer has one contract.
- Added a deterministic entangled relative-phase interference test: create a Bell pair, apply T while entangled, uncompute, then interfere with H. It checks both analytic complex survivors and both exact-zero outputs.
- Added boundary coverage for the zero-qubit identity, a selected bit above the first mask word, padding-bit rejection, and wrong-word-count rejection.

Validation on this head:

- 949/949 non-benchmark C++ tests passed;
- 55/55 focused Python amplitude tests passed;
- all pre-commit formatting, lint, typing, workflow-lint, and repository checks passed;
- git diff --check passed.

bachase:
Added the two requested deterministic regressions in c734748b:

- a persistent three-qubit GHZ correlation with a T phase on the chained terminal branch, checking 000, 111, and all six impossible outputs analytically;
- a persistent anti-correlated Bell pair across qubits 0 and 69, checking both phased survivors and both impossible equal-bit outputs across the mask-word boundary.

No implementation changes were needed. Validation on this head:

- 951/951 non-benchmark C++ tests passed;
- both new focused fixtures passed;
- all pre-commit formatting, lint, typing, workflow-lint, and repository checks passed;
- git diff --check passed.

docs/guide/strong-simulation.md
```diff
@@ -1,6 +1,10 @@
-# Strong Simulation: Exact Probabilities
+# Strong Simulation: Exact Amplitudes and Probabilities
 
-Clifft provides two exact-probability APIs covering complementary regimes:
+Clifft provides an exact complex-amplitude query and two exact-probability APIs:
+
+- **`clifft.compile_basis_amplitude(circuit, output_bits)`** followed by
```
bachase: Should query show up as the `x` and `output_buts` as the zero in the example? Otherwise it might be unclear how this maps from API to math.

docs/guide/strong-simulation.md
```diff
@@ -1,6 +1,10 @@
-# Strong Simulation: Exact Probabilities
+# Strong Simulation: Exact Amplitudes and Probabilities
 
-Clifft provides two exact-probability APIs covering complementary regimes:
+Clifft provides an exact complex-amplitude query and two exact-probability APIs:
```
bachase: I agree we should document this new feature, but should we put it at the end of the list? The API is much narrower <x|U|0> ... which makes it feel unbalanced with the other probability apis that are more general? Or is there a better API design to consider for amplitudes?

docs/guide/strong-simulation.md
```diff
@@ -9,17 +13,60 @@ Clifft provides two exact-probability APIs covering complementary regimes:
...
-substantially — see
+You can convert a basis-probability problem into a record-probability problem
+by adding explicit terminal measurements to a unitary program, and the two
+distributions match on the bitstrings the records encode. The runtime cost can
+differ substantially — see
 [Performance on overlapping circuits](#performance-on-overlapping-circuits).
 
+## Exact selected amplitudes
+
+An amplitude query is compiled for one output effect. Compilation reports its
```
bachase: "one output effect" is an odd phrasing. Maybe say one output bitstring (assuming terminal measurements)?

docs/guide/strong-simulation.md
```diff
@@ -9,17 +13,60 @@ Clifft provides two exact-probability APIs covering complementary regimes:
...
+"""
+
+query = clifft.compile_basis_amplitude(circuit, "11")
+print(query.peak_active_width)
+
+amplitude = clifft.evaluate_amplitude(query)
+print(amplitude)
+```
+
+The compiler represents the selected bra as query-private terminal
```
bachase: Rephrase and shorten this section, matching parameter names that you might change in response to the above, e.g. query parameter by bra in the prose.

docs/guide/strong-simulation.md
```diff
@@ -9,17 +13,60 @@ Clifft provides two exact-probability APIs covering complementary regimes:
...
+representative. Native-format `U3` uses Clifft's determinant-one
+`R_Z(phi) R_Y(theta) R_Z(lambda)` convention; select `input_format="qasm2"`
+when the source QASM scalar is part of the desired amplitude.
+
 ## When to use which
 
 | Your circuit… | Use |
 |---|---|
-| has no measurements | `basis_probabilities()` |
+| needs one phase-sensitive basis amplitude | `compile_basis_amplitude()` |
```
bachase: Is it really `compile_basis_amplitude` and `evaluate_amplitude`? Aside from wanting to know the max `k` before calling the method, should this just be one API call instead of two?

docs/guide/strong-simulation.md
```diff
@@ -355,6 +402,12 @@ hidden measurement slots (from `R` / reset gates lowered to
 measure-then-feedback) are also rejected. Recompile without resets, or use
 `sample()` to marginalize over the hidden outcomes.
 
+Amplitude-query compilation rejects measurements, resets, feedback, noise,
+detectors, observables, expectation probes, and postselection. It intentionally
+does not accept a precompiled `Program` or user-selected HIR passes: exact phase
```
bachase: I'd reorder the last sentence to be more like "Because the amplitude query extracts the exact phase, it does not accept the phase insensitive `Program` or user selected HIR passes.

src/clifft/circuit/gate_data.h
```diff
@@ -387,4 +387,8 @@ inline constexpr std::string_view gate_name(GateType g) {
 // Returns GateType::UNKNOWN for unrecognized names.
 GateType parse_gate_name(std::string_view name);
 
+// Returns the named inverse of a fixed Clifford gate. Parameterized and
```
bachase, line 390: Putting argument inversion on the caller seems fragile. Is there a better approach that remains fast and less prone to error?

src/clifft/frontend/frontend.cc
```diff
@@ -1,105 +1,73 @@
...
-        case GateType::XCY:
-        case GateType::XCZ:
-        case GateType::YCX:
-        case GateType::YCY:
-        case GateType::YCZ:
-            return gate;
-        default:
-            throw std::invalid_argument("Gate does not have an explicit Clifford inverse: " +
-                                        std::string(gate_name(gate)));
+struct ProjectiveTraceState {
```
bachase, line 25: Worth a concise comment above that although the main HIR and most representations are phase insensitive, some use cases like the basis amplitude requires tracking phase, so this and class below enable doing so in the frontend?

src/clifft/frontend/frontend.cc
bachase, line 1: These front end changes are a bit invasive; is it better to localize them to its own file and then hook into the frontend? 

What happens for workflows that don't require phase? Are they impacted?

src/clifft/sampling/executor.h
```diff
@@ -135,49 +155,49 @@ class Executor {
                                          uint32_t symbol_prefix_size) noexcept;
 
     template <ShotMode Mode>
-    [[nodiscard]] ReplayResult execute_actions_for_backend(std::span<const uint8_t> forced_records,
-                                                           uint32_t begin = 0) noexcept;
+    [[nodiscard]] ExecutionResult execute_actions_for_backend(
```
bachase: Its confusing to have both EXecutionResult and ReplayResult still. Are they both needed?

src/clifft/sampling/executor.h
```diff
@@ -109,16 +111,34 @@ class Executor {
...
+
+    struct ExecutionResult {
+        bool reachable = true;
+        double log_probability = 0.0;
+        uint32_t exact_half_probability_factors = 0;
+    };
+
+    // Terminal effects preserve small positive branch mass that ordinary
+    // sampling intentionally classifies as numerical dust.
+    [[nodiscard]] ExecutionResult replay_effect(std::span<const uint8_t> forced_records) noexcept;
```
bachase: Not in the code, but can you explain a bit more of why this is necessary?

src/clifft/sampling/planner.cc
```diff
@@ -723,11 +896,30 @@ SamplingPlan plan_sampling(const HirModule& hir, SamplingPlanOptions options) {
...
     // externally assembled plans before execution. Repeating that full scan
     // here makes the normal compile path validate the same plan twice.
     return plan;
 }
 
+SamplingPlan plan_sampling(const HirModule& hir, SamplingPlanOptions options) {
+    return plan_sampling_impl(hir, options, nullptr);
+}
+
+PhaseAwareSamplingPlan plan_sampling_phase_aware(const HirModule& hir,
```
bachase, line 909: Why is this not in its own cc file?

docs/guide/strong-simulation.md
bachase: Addressed in 5c1908d7. The guide now states explicitly that output_bits is x in <x|U|0...0>, both in the API list and the selected-amplitude section.

docs/guide/strong-simulation.md
bachase: Addressed in 5c1908d7. I reordered the guide around the two broader probability APIs and put the narrower selected-amplitude API afterward. The public convenience path is now basis_amplitude(circuit, output_bits); compile_basis_amplitude remains the feasibility-sensitive escape hatch.

docs/guide/strong-simulation.md
bachase: Addressed in 5c1908d7. The user-facing prose now says one complete output bitstring and avoids the internal output-effect terminology.

docs/guide/strong-simulation.md
bachase: Addressed in 5c1908d7. I shortened the section substantially, aligned it with output_bits and x, and removed the implementation-heavy phase-ledger and terminal-effect discussion.

docs/guide/strong-simulation.md
bachase: Addressed in 5c1908d7. The normal API is now one call: basis_amplitude(...). Advanced callers can use compile_basis_amplitude(...), inspect peak_active_width, and then call query.evaluate(). The redundant top-level evaluate_amplitude function was removed.

docs/guide/strong-simulation.md
bachase: Addressed in 5c1908d7. The limitation now leads with why source text is required: ordinary Program objects discard global phase, and target-aware compilation depends on the requested output bitstring.

src/clifft/circuit/gate_data.h
bachase: Addressed in 5c1908d7 by renaming the helper to inverse_fixed_clifford_gate, making its fixed, argument-free contract explicit at every call site. Parameterized rotations continue through their existing argument-aware inverse path; a generic gate-plus-arguments abstraction here would add machinery without removing a live ambiguous call site.

src/clifft/frontend/frontend.cc
bachase: Addressed in 5c1908d7. A concise comment now explains that ordinary HIR intentionally discards global phase while the amplitude compiler uses the exact query-private trace policy.

src/clifft/frontend/frontend.cc
bachase: Partially localized in 5c1908d7: PhaseAwareCliffordFrame implementation moved to phase_aware_frontend.cc. I kept the small trace policy and entry hook beside the shared templated tracer; moving those would require exposing or duplicating frontend internals. Non-phase workflows instantiate the no-op projective policy, so the compiler optimizes the hooks away, and the runtime executor path is unchanged.

src/clifft/sampling/executor.h
bachase: Addressed in 5c1908d7. The internal type is now private DispatchResult : ReplayResult, containing only the additional query-private counter. ReplayResult remains the stable public two-field result, including its structured-binding behavior.

src/clifft/sampling/executor.h
bachase: Expanded in 5c1908d7. Effect replay counts exact 1/2 branches separately so amplitude reconstruction can apply ldexp instead of losing the dyadic factor through log conversion or underflow. It also preserves small positive branch mass that ordinary stochastic sampling intentionally treats as numerical dust.

src/clifft/sampling/planner.cc
bachase: I kept this in planner.cc deliberately. plan_sampling_phase_aware shares the private plan_sampling_impl and its planning context; moving only the wrapper would force a new cross-file internal interface or duplicate planning logic. The phase-aware data types remain separated in their header, but the implementation stays next to the planner machinery it extends.

bachase:
Deferring this for now pending a clearer need

PR #413, opened 2026-08-26
refactor(sampling): unify syndrome value representation

### LINKED ISSUE #411: Unify syndrome value representation in SamplingPlan
## Context

PR #407 added an optional record-parity representation alongside the affine outcome on detector and observable write actions. The packed executor can use record parity directly, but carrying both values duplicates generation tracking, equivalence validation, inspection formatting, lowering state, and fallback logic in an otherwise executor-independent plan.

## Proposed direction

Represent each syndrome write with one backend-neutral selected value: either an affine Boolean expression or a record parity.

- Detectors can use record parity.
- Observables can use record parity while the referenced record snapshots remain current, and fall back to an affine value otherwise.
- Scalar CPU, packed CPU, and HIP lowering should each consume the selected representation from the plan.
- Preserve inspection output, structural validation, and exact scalar-versus-packed differential coverage.

This is intentionally separate from #407 because it changes the semantic plan representation across every backend. It is an architectural cleanup rather than a prerequisite for the packed execution performance or correctness.

bachase:
## Stack

This PR is stacked on #407. Its base is intentionally `codex/issue-313-interleaved-batches`; review commits `c3ac3f96..d30130e3` here.

## Summary

- replace the duplicated affine outcome plus optional packed-only record-parity sidecar with one backend-neutral output representation
- make detector values record parity by construction; only observables can select between record parity and an affine value for a historical snapshot
- lower the selected representation independently in the scalar CPU, packed CPU, and HIP backends
- stop preparing, validating, formatting, storing, and propagating affine expression registers that a record-parity output supersedes
- keep record-parity terms in one compact tape and allocate packed record columns only when a selected output actually reads them
- represent an active expectation as one projection-plus-sign value, so a zero probe cannot retain an irrelevant sign expression
- remove redundant plan counts and flags that are derivable from owned vectors or actions
- store only each symbol's kind; defining actions and noise-site membership remain owned by the action stream and noise distributions instead of duplicated metadata
- page-align packed bit matrices and account for their exact retained footprint, avoiding allocator-size-dependent cache behavior without inflating large carrier matrices to huge-page boundaries

## Correctness coverage

- canonicalizes record parity on construction and validates the remaining cross-object invariants
- characterizes planner selection around readout mutations and historical observable snapshots
- preserves exact scalar-versus-packed rows across capacities through mixed affine and record-parity outputs
- covers missing and duplicate symbol definitions, action-kind disagreement, use before definition, and duplicate noise binding without constructible metadata disagreements
- validates raw detector targets before XOR cancellation and canonicalizes observable snapshots before freshness selection
- checks HIP term-domain selection and CPU-differential results for historical affine and current record-parity observables in both coefficient precisions
- tests aligned allocation sizing and verifies packed memory-policy accounting uses the actual retained footprint

## Validation

Validation was rerun at `d30130e3`:

- pre-commit: passed
- Debug native CTest: 923/923 non-benchmark entries passed
- Python oracle, structured, statistical, and integration tests: 1,365 passed, 6 HIP-hardware skips
- executable documentation examples: 35 passed, 8 intentional skips
- WebAssembly Release build and smoke tests: passed
- earlier offline HIP `gfx942` build: `clifft_tests` and `clifft_hip_tests` compiled

## Performance

Same-VM A/B against the exact stacked parent `c3ac3f96`, with identical Release/native-CPU/OpenMP builds, one worker pinned to CPU 2, automatic batch selection, and identical seeded checksums. Sampling compilation is outside the timed interval. Lower is better.

| Workload | Parent | This PR | Change |
|---|---:|---:|---:|
| surface d7, 1M shots, aggregate output, median | 250.605 ms | 228.655 ms | -8.8% |
| surface d9, 1M shots, aggregate output, median | 604.055 ms | 545.175 ms | -9.7% |
| generated non-Clifford width 5/depth 20, 100k shots, median | 412.348 ms | 411.845 ms | -0.1% (flat) |
| surface d7 compile, 100-run median total | 12.897 ms | 11.188 ms | -13.3% |

The parent Clifford measurements were repeated after the current run and the table uses the faster parent median, making the reported gains conservative. The packed alignment fix matters because removing redundant expression registers moved some matrices across the system allocator's size threshold; explicit page alignment preserves the intended locality independent of allocator behavior.

Closes #411

### DIFF
--- a/src/clifft/sampling/plan.h
+++ b/src/clifft/sampling/plan.h
@@ -197,29 +197,32 @@ struct ApplyReadoutNoise {
     double prob_one_to_zero = 0.0;
 };
 
-// Batch execution can evaluate detector and observable parities from packed
-// record columns instead of expanding every record's full affine history.
-struct BatchRecordParity {
+// A parity over circuit record slots at the point where a syndrome value is
+// written. Record slots are canonicalized in strictly increasing order.
+struct RecordParity {
     bool constant = false;
     std::vector<RecordSlot> records;
 };
 
+// Syndrome writes retain the representation whose dependencies are still
+// available at execution time. Record parity avoids expanding current record
+// snapshots; affine form preserves historical values after a record changes.
+using SyndromeValue = std::variant<AffineBool, RecordParity>;
+
 // Writes a detector parity after the planner has XORed its expected reference
-// parity into the expression. A nonzero postselected detector rejects the shot
-// immediately; later actions and outputs are irrelevant for it.
+// parity into the selected value. A nonzero postselected detector rejects the
+// shot immediately; later actions and outputs are irrelevant for it.
 struct WriteDetector {
-    AffineBool outcome;
+    SyndromeValue outcome;
     DetectorSlot detector{};
     bool postselected = false;
-    std::optional<BatchRecordParity> batch_parity;
 };
 
 // Writes one fully accumulated logical observable after the planner has XORed
-// its expected reference parity into the expression.
+// its expected reference parity into the selected value.
 struct WriteObservable {
-    AffineBool outcome;
+    SyndromeValue outcome;
     ObservableSlot observable{};
-    std::optional<BatchRecordParity> batch_parity;
 };
 
 // Writes a non-destructive Pauli expectation probe. The planner retains only
--- a/src/clifft/sampling/plan.h
+++ b/src/clifft/sampling/plan.h
@@ -300,21 +300,6 @@ struct PlannedAction {
     SamplingAction action;
 };
 
-// Describes how one symbol is populated. SamplingPlan::validate checks each
-// kind's legal fields and its agreement with the referenced defining action.
-struct SymbolInfo {
-    SymbolKind kind = SymbolKind::Unused;
-
-    // Index of the action that assigns this symbol. Required for Derived,
-    // Branch, Readout, and Instrument; absent for Presampled and Unused.
-    std::optional<uint32_t> defining_action;
-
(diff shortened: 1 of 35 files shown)


src/clifft/sampling/plan.h
```diff
@@ -197,37 +197,59 @@ struct ApplyReadoutNoise {
...
+    [[nodiscard]] bool is_canonical() const;
+
+    friend bool operator==(const RecordParity&, const RecordParity&) = default;
+
+  private:
+    bool constant_ = false;
+    std::vector<RecordSlot> records_;
 };
 
+// Observable writes retain the representation whose dependencies are still
```
bachase, line 226: Consider adding a brief summary or table  here or at the top of this section in a comment that concisely describes the situation

e.g. "DETECTOR and OBSERVABLES are parity (XOR) of records. In the absence of readout noise or later modification of records, these can be calculated immediately, by XORing the slots identified in RecordParity. For cases like OBSERVABLE_INCLUDE, that build up its parity expression each time it is called but are evaluated at the end of the circuit, intervening readout noise means the final rec value might differe from the one at the callsite. In that case, we need to evaluate against the original symbols (which is the AffineBool) variant below. These cases can be determined entirely at planning time, and then represented as these different ObservableValue variants to execute at sample time."

Or a simpler/concise version of that.

bachase: Addressed in d30130e3. The overview now explains that outputs are XOR parities of measurement-record snapshots, why detectors always see current slots, why accumulated observables sometimes require historical `AffineBool` values, and that executors consume the planner-selected representation without repeating freshness analysis.

Issue #415, opened 2026-08-27
perf(sampling): make automatic packed batching postselection- and cache-aware


bachase:
## Problem

Automatic packed batching can regress sharply when survivor sampling uses detector postselection, even on a circuit where the same packed executor is substantially faster without postselection.

On an M4 Pro, Release build, one thread, 100,000 shots of `pure_surface_d7_r7_p1e-3.stim` through aggregate `sample_survivors`:

| mode | `batch_size=1` | `batch_size="auto"` |
|---|---:|---:|
| no detector postselection | 249.5 ms | 16.0 ms |
| all detectors postselected | 132.7 ms | 1187.1 ms |

Thus postselection changes automatic batching from about a 15.6x win to about a 9x regression. Explicit packed capacities from 64 through 2048 lanes were also slower than the scalar path in the postselected case, so merely selecting a smaller capacity did not resolve this workload.

This is independent of the Apple NEON per-shot backend: the packed path bypasses that backend, and forcing scalar versus NEON ISA does not change packed execution time.

PR #407's published d7/d9 comparison intentionally disabled detector postselection. The current `clifft-bench` QEC workload contract instead postselects every detector. That left an important production mode outside the policy's performance validation.

## Likely interaction to investigate

The scalar executor exits an individual shot at its first failed postselected detector. A packed executor must retain the remaining live lanes, carry dead lanes until compaction is profitable, and compact retained state and sidecar columns when selected. Low or progressively falling survival can therefore reverse the packed/scalar cost balance.

Please profile before choosing a fix. In particular, separate time spent in detector parity evaluation, live-mask maintenance, lane compaction, packed-column compaction, state compaction, and work performed after most lanes have died.

## Scope

Audit automatic packed selection across execution modes that were not all represented in the original throughput study:

- fixed-row sampling versus survivor sampling;
- no postselection, partial postselection, and all-detector postselection;
- aggregate survivor counts versus retained output rows;
- ordinary sampling versus fixed-fault `sample_k` and `sample_k_survivors`;
- `keep_records=false` and `keep_records=true`;
- sparse, dense, symmetric, asymmetric, and categorical noise where their packed paths differ;
- low, medium, and high survival rates, including circuits where failures happen early versus late;
- partial final batches, explicit capacities, one worker, and multiple shot workers.

[...]

PR #418, opened 2026-08-27
perf(sampling): make packed lane policy cost aware

bachase:
Builds on #417, now merged. That PR owns the packed executor lifetime fix; this PR owns automatic policy, cost calibration, profiling, and user guidance.

## Summary

- precompute approximate per-lane coefficient work and annotate postselected detectors with their remaining work
- compare saved dead-lane work with the cost of compacting live coefficient state and packed sidecars
- calibrate packed sidecar compaction against the broader AMD, M4, and Intel evidence
- use a conservative portable default: automatic batching stays scalar for postselected plans, sustained width-5 work, and widths above 5
- retain automatic packed execution for eligible non-postselected plans, subject to shot, work, state, worker-memory, and thread-layout limits
- continue honoring explicit batch capacities, including for postselected plans, so expert users can benchmark and override the default
- document exactly when automatic batching applies and recommend testing representative explicit capacities on the target workload and CPU
- expose cost and effective-policy data through executable inspection and the profiling tools

## Relationship to #417

#417 fixes how explicitly packed postselected execution manages dead-lane lifetimes, including the known aggregate-survivor regression. This PR builds on that implementation without broadening its scope: postselected `batch_size="auto"` is deliberately scalar, while an explicit packed request uses the improved executor from #417.

Future workload-specific tuning, potentially including a Python interface, is tracked separately in unitaryfoundation/clifft-bench#37 without prescribing an implementation.

## Evidence and policy decision

- AMD EPYC: 70/70 wins in the eligible repository corpus; calibrated random width-5 study produced 175 wins, 5 neutral cases, and 0 regressions. Width 6 was mixed, while widths 7 and 8 consistently regressed.
- Apple M4 Pro: short width-5 packed execution delivered large wins, and explicit batch 256 also improved the sustained width-5 fixture.
- Intel C7i: short non-postselected width-5 execution improved by roughly 46-74%, but postselected and sustained cases showed that the profitable capacity depends on survivor lifetime and architecture. Sustained explicit batching regressed substantially.

The conservative default keeps the repeatable cross-platform wins without encoding survival-rate or machine-specific assumptions from these circuits. Explicit capacities preserve the platform-specific opportunities.

## Validation

- focused packed policy and survivor tests
- `ctest --test-dir build-profile --output-on-failure` (931/931)
- `uv run pytest tests/python/test_sample.py tests/python/test_importance_sampling.py -q` (130/130)
- `just docs-build`

Relates to #415.


bachase:
## Broader benchmark update

I froze policy changes and expanded the study on an AMD EPYC 9554P VM. Timings below use medians, a 5% win/regression threshold, and CPU pinning for the single-thread randomized runs.

| Study | Result |
|---|---|
| Repository circuits eligible for automatic batching | 70 wins, 0 neutral, 0 regressions across seven circuits and ten sampling-mode combinations |
| Initial deterministic random corpus, active widths 0-5 | 212 wins, 1 neutral, 3 regressions across 216 cases; all three regressions came from one sustained width-5 plan |
| Expanded deterministic random width-5 corpus, explicit 1024 lanes | 172 wins, 7 neutral, 1 regression across 180 cases |
| Same width-5 corpus under the current 16,384-work automatic cutoff | 136 wins, 7 neutral, 1 regression among 144 eligible cases; all 36 excluded high-work cases happened to be wins, so the cutoff is conservative and misses real opportunities |
| Wider random plans, explicit 1024 lanes | actual width 6: 18 wins, 9 neutral, 3 regressions; width 7: 30/30 regressions; width 8: 36/36 regressions |

The random plans span transient, mixed, and sustained active-state lifetimes; randomized rotation/Clifford action mixes; detector counts and error probabilities; ordinary and fixed-k sampling; aggregate and retained survivor output; and alternating postselection.

### Candidate and rejected refinements

- Charging packed-column compaction more accurately (isolated commit `26f3ba1a`, not in this PR) improved the expanded width-5 result to 175 wins, 5 neutral, 0 regressions at 1024 lanes. A separate width-1-to-4 corpus was 144/144 wins. This is the candidate for cross-architecture validation.
- A universal 256-lane width-5 policy was rejected: even with the improved compaction cost it produced two >5% single-thread regressions in 180 cases.
- A postselection-exposure cutoff was rejected: it disabled 32 additional wins and did not eliminate the remaining regression.
- Eight-worker runs exposed a portability/threading risk: some width-5 retained-survivor paths regressed at 1024 lanes and improved at 256, but making automatic capacity thread-dependent would change automatic batch boundaries and the universal 256 rule is not safe.

### Interpretation

This supports the packed-lane lifetime mechanism, but not a claim that the current automatic eligibility cutoff is universally optimal. Peak width above 6 is a robust negative signal on this host; width 6 is mixed; width 5 depends on lifetime, action mix, output mode, and threading. The main remaining risk is microarchitecture dependence: Apple NEON, Intel SIMD, cache topology, and memory bandwidth can move both compaction and batching crossovers.

I am therefore leaving this PR in draft and keeping the empirical compaction calibration isolated until the exact baseline/candidate pair has been run on Apple M4 and Intel hosts.

bachase:
## Apple M4 cross-host update

The committed evidence is on `bench/issue-415-macos-m4-pro` at `7e88de8a` (`aa09d3c8` contains the broader initial sweep; `7e88de8a` adds the fixed alternating-postselection follow-up). I independently reproduced the committed classifications and configuration/checksum checks.

For the isolated packed-column cost calibration at `26f3ba1a` versus this PR at `fba10d75`:

- focused alternating study, all 64 comparisons: 6 wins / 54 neutral / 4 regressions at a 5% threshold
- comparisons with both medians at least 5 ms: 1 win / 46 neutral / 1 regression
- packed comparisons with both medians at least 5 ms: **1 win / 31 neutral / 0 regressions**
- ordinary packed `sample_survivors`: 1 win / 19 neutral / 0 regressions
- the only reliable packed win was transient, eight workers, retained survivor rows, `auto=1024`: 6.877 -> 6.403 ms (-6.9%)
- the only reliable regression was a scalar control; the compaction change cannot affect it

The fixed-k percentages are not decision evidence in these fixtures: every fixed-k survivor case retained zero shots and all sustained fixed-k medians were below 5 ms. Their large relative swings are scheduler/timer noise over small absolute differences.

### Architecture-dependent width-five crossover

The larger finding is that the current width-five work cutoff is conservative but not portable-optimal:

- M4 Pro, sustained width-five fixture: explicit 256 lanes improved the ordinary/non-microbenchmark paths by roughly 25-38% versus scalar, while `auto` selected scalar because work=25,304 exceeds the 16,384 cutoff.
- Matched AMD EPYC 9554P rerun, 100,000 shots and 11 repetitions: explicit 256 lanes was 7-13% slower serially on the same ordinary paths; with eight workers it was mostly neutral (one +6.7% case, the rest within 5%).

Therefore relaxing the width-five cutoff globally would recover a substantial Apple win but reintroduce AMD regressions. Conversely, the isolated `4 -> 16` packed-column cost calibration looks safe on M4 and improved the randomized AMD width-five corpus, but it should remain isolated until the Intel run is available.

This reinforces keeping architecture/cache-aware eligibility tuning separate from the packed-lane lifetime mechanism.

docs/guide/simulation.md
```diff
@@ -278,13 +278,35 @@ the memory required by the selected sampling and output mode, targets at most
...
 the requested thread count, preserving deterministic batch boundaries.
 
-Automatic selection currently requires at least 64 shots and a peak active
-width of at most 5, and it considers capacities up to 2048 lanes. Set
-`batch_size=1` to disable batching. An explicit positive integer requests a
-capacity up to 2048 and is mainly useful for profiling or overriding the
-conservative automatic policy. Explicit mode still rejects a dense packed
-state above 64 MiB per worker, but plans retaining many symbols, expressions,
-or records can use additional memory beyond that state-vector limit.
+Automatic selection is conservative. It requires at least 64 shots, no
```
bachase: Way to much detail -- briefly say the defaults (up to peak active width 5 and determines optimal number of lanes, but falls back to non-batched for long width 5 circuits). Plans with postselection do not batch in automated mode because  ....

Users can set an explicit batch_size, include 1 to force scalr execution. Since batching benefit varies based on the specific circuit , we recommend testing a few sizes to see what gives the best performance. For postselected sampling, note that reject lanes are marked to avoid unnecessary work, but lane compaction may be deffered dpeending on the workload"

(OR simila rcomment -- do we need to briefly define lane compaction or use a different word?)

bachase: Updated in 6ce78ee1. I condensed this to the automatic eligibility/default, explicit overrides with a recommendation to benchmark representative sizes, and a short postselection explanation. The user-facing text avoids the term lane compaction and says that rejected lanes stop executing immediately while Clifft may wait to repack the remaining live lanes until repacking is worthwhile.

bachase:
Addressed the two follow-up performance findings in 1a8b5e82:

- batch work is now stored as common work plus row-output-only work, so aggregate-survivor policy and compaction omit discarded expectation-value probes; aggregate compaction cost also omits expectation storage
- all-width and width-five work are now tracked separately, and the sustained-width-five cutoff uses only mode-specific width-five work
- added regressions for a width-five plan dominated by discarded expectation outputs and for a brief width-five prefix followed by substantial width-one work

Validation: 931/931 optimized C++ tests and the full required pre-commit suite pass.

bachase:
Follow-up in 3932264c removes the tracker overhead from ordinary plan construction: plan-wide work now accumulates during lowering, non-postselected plans allocate no per-action work vector, and postselected plans retain only 16 bytes per source action for the reverse compaction pass (down from 32). Postselection detection reuses the existing storage-estimation scan, so this adds no extra plan pass. The 931-test optimized suite passed with assertions forced on, along with the full pre-commit suite.

PR #419, opened 2026-08-27
perf(amplitudes): select lower-width commuting schedule

bachase:
## Summary

- expose the reverse ordering of mutually commuting non-Clifford expansions as a second legal squeeze schedule
- prepare both schedules for selected basis-amplitude queries and retain the one with lower peak active width
- keep the default squeeze behavior and ordinary sampling execution path unchanged

This is stacked on #412.

## Results

On the 196-circuit ABSTRACTS corpus, direct Clifft planning improves 105 circuits and worsens none. Median peak active width falls from 11 to 10 and maximum width falls from 28 to 25. The largest individual reduction is 28 to 16.

Planning both schedules raises median amplitude-query preparation from approximately 4.2 ms to 8.1 ms. The tradeoff is query-local and targets the exponential execution parameter.

## Validation

- 947 native tests
- 1,415 Python tests, 6 skipped, 1 expected failure
- 300 randomized QASM circuits against Qiskit
- 114 published Qiskit corpus amplitudes match componentwise within 8.2e-15
- all 196 published PyZX corpus magnitudes match within 8.2e-16
- full pre-commit suite

### DIFF
--- a/src/clifft/api/basis_amplitudes.cc
+++ b/src/clifft/api/basis_amplitudes.cc
@@ -60,6 +60,25 @@ Circuit append_terminal_measurements(const Circuit& source) {
     return result;
 }
 
+struct PlannedAmplitude {
+    SamplingPlan plan;
+    std::complex<double> phase;
+};
+
+PlannedAmplitude plan_amplitude(HirModule hir, PhaseAwareCliffordFrame final_clifford_frame,
+                                std::span<const uint8_t> output_records,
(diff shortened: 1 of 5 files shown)


bachase:
Superseded by #424, which applies the lower-width squeeze scheduling improvement once in the general production HIR pipeline instead of selecting a second amplitude-specific schedule. The basis-amplitude API work remains independently tracked in #412.

Issue #409, opened 2026-08-26
Improve documentation on sampling and backend modes


bachase:
clifft has evolved to support CPU and an experimental GPU backend, different CPU modes (inter-shot and intra-shot threading and batching), alongside the original importance sampling and instrument sampling approaches. The documentation here has gotten quite long. We should consider rewriting the documentation to help users understand when they would pick certain modes, and separate high-level usage guides from more expert options where users will want more control of settings. This should also work well as we add additional GPU backends.

Also consider how frontends now show up -- just having the QASM2 sub page makes it look like that is the only one we support.

bachase:
The documentation work is ready as an ordered stacked review series:

1. #440 -> #446 (base: `main`) — workflow chooser and shared terminology
2. #441 -> #447 (base: #446 branch) — beginner entry points and circuit inputs
3. #442 -> #448 (base: #447 branch) — task-focused simulation workflows
4. #443 -> #449 (base: #448 branch) — advanced CPU execution and tuning
5. #444 -> #450 (base: #449 branch) — experimental HIP guide and backend capability matrix
6. #445 -> #452 (base: #450 branch) — validated workflow/backend contracts and executable examples

Please review and merge in that order. Each PR closes its child issue when merged; #409 remains the parent tracking issue.

Scope decisions reflected throughout the stack:
- no `llms.txt` and no separate AI-only documentation path;
- the same structured, explicit, copyable docs serve humans and integration tools;
- HIP is visible but consistently marked experimental and source-build only;
- CUDA is documented as planned, not implemented, with no implied current API.

All six PRs are currently mergeable and their CI gates are green.

bachase:
Closed as part of #455

PR #446, opened 2026-08-31
docs: add workflow chooser

### LINKED ISSUE #440: docs: add a workflow chooser and shared terminology
Parent: #409

Establish one conceptual entry point for choosing how to use Clifft.

Scope:
- Add a Choose a workflow page that maps user goals and circuit constraints to the appropriate API and result type.
- Define and consistently distinguish circuit input, simulation workflow, CPU execution strategy, and hardware backend; avoid using mode as a catch-all.
- Present stable CPU defaults first and label experimental paths explicitly.
- Link the chooser from the quick start, simulation guide, and API reference.

Out of scope:
- Detailed navigation rewrites, CPU tuning guidance, and GPU setup, which are covered by sibling issues.

Done when a new user can select the correct supported API without first understanding Clifft execution internals.

bachase:
## Summary

- add a central workflow chooser organized by user result and circuit constraints
- distinguish circuit inputs, simulation workflows, CPU execution strategies, and hardware backends
- make stable CPU defaults and the explicit experimental HIP boundary clear
- link the chooser from Getting Started, Simulation, and the Python API reference

## Validation

- `.venv/bin/pytest --codeblocks docs/getting-started/choosing-a-workflow.md -v`
- `.venv/bin/mkdocs build --strict`

Closes #440
Parent: #409

### DIFF
--- a/docs/getting-started/choosing-a-workflow.md
+++ b/docs/getting-started/choosing-a-workflow.md
@@ -1,89 +1,91 @@
 # Choose a Workflow
 
-Choose the result you need before choosing performance settings. Circuit input,
-simulation workflow, CPU execution strategy, and hardware backend are separate
-decisions in Clifft.
+Clifft offers several ways to provide a circuit, obtain results, and tune
+performance. This page walks through those three choices so you can select a
+workflow from the result you need rather than from Clifft's implementation
+details.
 
-If you are unsure, start with the ordinary CPU workflow:
+!!! tip "First time using Clifft?"
+    Start with the [Quick Start](quickstart.md) to compile and sample your first
+    circuit, then return here when you need a different result or execution
+    strategy.
 
-```python
-import clifft
+## 1. Provide a circuit
 
-program = clifft.compile("H 0\nT 0\nM 0")
-result = clifft.sample(program, shots=1000)
-print(result.measurements[:5])
-```
+For most users, the starting point is Stim-compatible circuit text passed to
+`clifft.compile()`. This is Clifft's native input and supports the complete
+circuit workflow.
 
-This uses the stable CPU backend and its default execution settings. Change the
-workflow only when the result or circuit semantics require it. Tune execution
-only after measuring a representative workload.
+If your circuit already uses another format:
 
-## Choose the result you need
+- pass supported unitary OpenQASM 2 text with `input_format="qasm2"`
+- use the separately released `clifft-qiskit` adapter for supported Qiskit
+  circuits
+- use the separately released `clifft-cirq` converter or sampler for supported
+  Cirq circuits
 
-| Goal or circuit requirement | Use | Result and important constraints |
-|---|---|---|
-| Draw measurement outcomes, detector events, observables, or expectation values | `clifft.compile()` then `clifft.sample()` | Returns one row per requested shot. The compiled program must not have a post-selection mask. |
-| Discard shots when selected detectors fire | Compile with `postselection_mask`, then use `clifft.sample_survivors()` | Returns survivor counts. Pass `keep_records=True` when per-survivor rows are also needed. |
-| Query exact computational-basis probabilities for a unitary circuit with no measurements | `clifft.basis_probabilities()` | Queries selected bitstrings without constructing every output probability. Noise, measurements, detectors, observables, and post-selection are not supported. |
-| Query exact probabilities of measurement records in a noiseless circuit | `clifft.record_probabilities()` | Supports measurement records and classical feedback. Noise, detectors, observables, post-selection, and hidden reset records are not supported. |
-| Inspect the complete final pure state of a small unitary circuit | `clifft.get_statevector()` | Debugging and validation path. Expands all physical qubits and is currently limited to 10 qubits. |
-| Estimate rare events by conditioning on exactly `k` faults | `clifft.sample_k()` or `clifft.sample_k_survivors()` | Advanced statistical workflow. Results are conditional on `K = k` and must be combined with the corresponding fault-count probabilities. |
-| Model leakage, loss, or other supported noncomputational transitions | `clifft.noncomp.sample()` | Experimental trajectory workflow. It accepts a circuit and model together and compiles continuations internally. |
+These alternatives have format-specific restrictions. See
+[Circuit Inputs and Integrations](integrations.md) before using one. The input
+path does not select the simulation workflow or execution backend.
 
-Use the [Simulation guide](../guide/simulation.md) for ordinary and
-post-selected sampling. The
-[Strong Simulation tutorial](../guide/strong-simulation.md) covers exact
-queries, the [Importance Sampling tutorial](../guide/importance-sampling.md)
-covers fixed-fault strata, and the
-[Leakage and Loss guide](../guide/leakage-and-loss.md) covers noncomputational
-trajectories.
+## 2. What do you want to know?
 
-## Choose the circuit input separately
+### Sample circuit outcomes
 
-The input path does not select the simulation workflow or execution backend.
+- **I need an output row for every requested shot.** Use `clifft.compile()` and
+  `clifft.sample()` to obtain measurement, detector, observable, and expectation
+  value results. Start with the [Simulation guide](../guide/simulation.md).
+- **I need samples conditioned on post-selection.** Compile with a
+  `postselection_mask`, then use `clifft.sample_survivors()`. It returns survivor
+  counts and can optionally retain each survivor's records. See the
+  [post-selection workflow](../guide/simulation.md#post-selection-survivor-sampling).
+- **I need rare-event estimates conditioned on exactly `k` faults.** Use
+  `clifft.sample_k()` or `clifft.sample_k_survivors()`, then combine the
+  conditional results with the corresponding fault-count probabilities. This
+  is an advanced workflow covered by the
+  [Importance Sampling tutorial](../guide/importance-sampling.md).
 
-| Starting point | Path |
-|---|---|
-| Stim-compatible text with Clifft extensions | Pass the text to `clifft.compile()`. This is the native path and supports the full stable circuit workflow. |
-| Unitary OpenQASM 2 text | Pass `input_format="qasm2"` to `clifft.compile()`. Only the documented unitary subset is supported. |
-| Qiskit `QuantumCircuit` | Use the separately released `clifft-qiskit` adapter for its supported terminal-measurement workflow. |
-| Cirq `cirq.Circuit` | Use the separately released `clifft-cirq` converter or sampler for its supported circuit subset. |
+### Calculate exact probabilities
 
-See [Circuit Inputs and Integrations](integrations.md) for installation,
-examples, and current limitations.
+- **My unitary circuit has no measurements.** Use
+  `clifft.basis_probabilities()` to query selected computational-basis outcomes
+  without constructing every output probability.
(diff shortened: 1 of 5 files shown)


docs/getting-started/quickstart.md
bachase, line 1: As a ride-along to this PR (or to come in a later PR in this stack) -- the quickstart mentions state vector access, which is really more a debug tool. I think we should drop it from the quick start and instead focus more on sampling to get measurement outputs

docs/guide/simulation.md
```diff
@@ -1,5 +1,10 @@
 # Simulation
 
+Not sure which API applies to your circuit? Start with
+[Choose a Workflow](../getting-started/choosing-a-workflow.md). This page
```
bachase: Does "This page" mean this simulation page, or the choosing a workfow page?

If the former, I'd switch the "Choose A workflow" to a callout style box instead.

docs/reference/python-api.md
```diff
@@ -1,8 +1,12 @@
 # Python API Reference
 
-`clifft.compile()` returns a reusable `Program` consumed by the sampling and
-exact-query functions below. All public sampling functions use that program
-directly; there is no sampler-selection module or backend switch.
+The stable CPU workflow starts with `clifft.compile()`, which returns a reusable
```
bachase: I'd just drop this paragraph and have this just be the python API listings.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
+# Choose a Workflow
+
+Choose the result you need before choosing performance settings. Circuit input,
```
bachase: I'd write in a more easy to read framing. Something like

"Clifft attempts to offer lots of choices to work with a variety of use cases and performance requirements. By working backwards from your specific goals, this page helps you select the appropriate clifft workflow" 

basically help ease in and setup why we even have this page in user-oriented terms.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+simulation workflow, CPU execution strategy, and hardware backend are separate
+decisions in Clifft.
+
+If you are unsure, start with the ordinary CPU workflow:
```
bachase: I'd drop this and just say "first time? go checkout the quickstart page" or similar framing.

docs/getting-started/choosing-a-workflow.md
bachase: This can also setup the three choices in the sections below.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+|---|---|---|
+| Draw measurement outcomes, detector events, observables, or expectation values | `clifft.compile()` then `clifft.sample()` | Returns one row per requested shot. The compiled program must not have a post-selection mask. |
+| Discard shots when selected detectors fire | Compile with `postselection_mask`, then use `clifft.sample_survivors()` | Returns survivor counts. Pass `keep_records=True` when per-survivor rows are also needed. |
+| Query exact computational-basis probabilities for a unitary circuit with no measurements | `clifft.basis_probabilities()` | Queries selected bitstrings without constructing every output probability. Noise, measurements, detectors, observables, and post-selection are not supported. |
+| Query exact probabilities of measurement records in a noiseless circuit | `clifft.record_probabilities()` | Supports measurement records and classical feedback. Noise, detectors, observables, post-selection, and hidden reset records are not supported. |
+| Inspect the complete final pure state of a small unitary circuit | `clifft.get_statevector()` | Debugging and validation path. Expands all physical qubits and is currently limited to 10 qubits. |
```
bachase: I'd drop get_statevector -- we can have it documente din a few spots but I'd not put it out as a major use case.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+This uses the stable CPU backend and its default execution settings. Change the
+workflow only when the result or circuit semantics require it. Tune execution
+only after measuring a representative workload.
+
+## Choose the result you need
+
+| Goal or circuit requirement | Use | Result and important constraints |
+|---|---|---|
```
bachase: This seems a little text heavy, is there a shorter/more concise version?

Should we have anything more hierarchical/decision tree oriented? (Maybe not, just considering) Tihngs like post selection and importance sampling seem to be a subset/specialization of regular sampling; noncomputational seems its own tree, and then records + probs seems another.

Should we have a more relevant example for each? Like link to either a tutorial page if we want to go in depth? Or thats overkill?

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+| Draw measurement outcomes, detector events, observables, or expectation values | `clifft.compile()` then `clifft.sample()` | Returns one row per requested shot. The compiled program must not have a post-selection mask. |
+| Discard shots when selected detectors fire | Compile with `postselection_mask`, then use `clifft.sample_survivors()` | Returns survivor counts. Pass `keep_records=True` when per-survivor rows are also needed. |
```
bachase: I'd flip this around -- its more about "My results are post-selected"

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+program = clifft.compile("H 0\nT 0\nM 0")
+result = clifft.sample(program, shots=1000)
+print(result.measurements[:5])
+```
+
+This uses the stable CPU backend and its default execution settings. Change the
+workflow only when the result or circuit semantics require it. Tune execution
+only after measuring a representative workload.
+
+## Choose the result you need
```
bachase: Rather than "choose the result" can this be framed more problem / use case phrasing, e.g. "What are you measuring" or "What do you want to know"?

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+| Query exact probabilities of measurement records in a noiseless circuit | `clifft.record_probabilities()` | Supports measurement records and classical feedback. Noise, detectors, observables, post-selection, and hidden reset records are not supported. |
+| Inspect the complete final pure state of a small unitary circuit | `clifft.get_statevector()` | Debugging and validation path. Expands all physical qubits and is currently limited to 10 qubits. |
+| Estimate rare events by conditioning on exactly `k` faults | `clifft.sample_k()` or `clifft.sample_k_survivors()` | Advanced statistical workflow. Results are conditional on `K = k` and must be combined with the corresponding fault-count probabilities. |
+| Model leakage, loss, or other supported noncomputational transitions | `clifft.noncomp.sample()` | Experimental trajectory workflow. It accepts a circuit and model together and compiles continuations internally. |
+
+Use the [Simulation guide](../guide/simulation.md) for ordinary and
```
bachase: As you consider the alternative design -- these should live more as inline links with the appropriate section versus readers having to read the table and then jump down and read this to find where to go to learn more.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+post-selected sampling. The
+[Strong Simulation tutorial](../guide/strong-simulation.md) covers exact
+queries, the [Importance Sampling tutorial](../guide/importance-sampling.md)
+covers fixed-fault strata, and the
+[Leakage and Loss guide](../guide/leakage-and-loss.md) covers noncomputational
+trajectories.
+
+## Choose the circuit input separately
```
bachase: Should this come first? More of a "Provide a circuit"?

Although we do support these 4, stim is by far the primary input format and others have restrictions. Wondering how front and center we should have this

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+|---|---|
+| Stim-compatible text with Clifft extensions | Pass the text to `clifft.compile()`. This is the native path and supports the full stable circuit workflow. |
+| Unitary OpenQASM 2 text | Pass `input_format="qasm2"` to `clifft.compile()`. Only the documented unitary subset is supported. |
+| Qiskit `QuantumCircuit` | Use the separately released `clifft-qiskit` adapter for its supported terminal-measurement workflow. |
+| Cirq `cirq.Circuit` | Use the separately released `clifft-cirq` converter or sampler for its supported circuit subset. |
+
+See [Circuit Inputs and Integrations](integrations.md) for installation,
+examples, and current limitations.
+
+## Choose execution only after the workflow
```
bachase: Similar to the above, should this be more a decision tree, or menu of options? Or more a bulleted list to at least have some visual structure if options still compose.

I'd frame this as "Performance options", then CPU vs GPU, then CPU batch or thread (but note things have decent defaults) and then within GPU mention the experimental approach.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+[Parallel Sampling](../guide/simulation.md#parallel-sampling).
+
+GPU execution is experimental and is never selected automatically. The current
+HIP backend uses the separate `clifft.experimental.hip` API, supports only its
+documented subset, and requires an explicit HIP-enabled source build. See the
+[Experimental HIP Sampling Backend](../development/hip-backend.md) before using
+it. Additional GPU backends will follow the same explicit experimental boundary
+until their support contracts are promoted.
+
+## Terminology
```
bachase: If we have a terminology section, that means this page is too complex. It is meant to be a very easy onboarding page.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,89 @@
...
+| Unitary OpenQASM 2 text | Pass `input_format="qasm2"` to `clifft.compile()`. Only the documented unitary subset is supported. |
+| Qiskit `QuantumCircuit` | Use the separately released `clifft-qiskit` adapter for its supported terminal-measurement workflow. |
+| Cirq `cirq.Circuit` | Use the separately released `clifft-cirq` converter or sampler for its supported circuit subset. |
+
+See [Circuit Inputs and Integrations](integrations.md) for installation,
+examples, and current limitations.
+
+## Choose execution only after the workflow
+
+The stable `clifft.compile()` and sampling APIs execute on the CPU. Automatic
```
bachase: Don't use the word stable as it might imply there are some non-stable APIs. I'd leave all CPU sections without this type of "experimental" or "stable" designation, and its just GPU that gets experimental.

docs/getting-started/choosing-a-workflow.md
```diff
@@ -0,0 +1,91 @@
...
+  `clifft.record_probabilities()` to query exact probabilities of selected
+  measurement records.
+
+Both APIs have circuit restrictions. The
+[Strong Simulation tutorial](../guide/strong-simulation.md) explains when to
+use each one.
+
+### Model leakage or loss
+
+Use the experimental `clifft.noncomp.sample()` trajectory workflow for
```
bachase, line 63: Lets drop the experimental here at least. I think the warning on the leak/loss page is fine but we don't need to overdo it.

Visually, shoul dthis match the other section styles of the "My circuit has leakage or loss"? as a bolded list item?

bachase:
Superseded by #455, which consolidates the full documentation review into one final-state PR and preview.

Issue #458, opened 2026-09-01
Active-width scheduling: production integration and default-pipeline decision


bachase:
## Goal

Deliver compile-time active-width scheduling that improves sampling performance by reducing peak active width or time spent at high width, while preserving sampling distributions. Evaluate compilation cost together with sampling cost, and resolve whether the pass should run by default.

This issue tracks production integration through #507. Exact search and optimality certificates are separate research work, as recorded in the [scope decision](https://github.com/unitaryfoundation/clifft/issues/458#issuecomment-5513907353).

## Current status as of 2026-09-24

- **Implementation available, not merged:** #507 consolidates logical-position noise signs, structural width analysis, legal scheduling, the scheduling pass, Python inspection, documentation, and measurement tools. It replaces #460, #471, #472, and #502; the earlier six-PR stack is historical.
- **Candidate behavior:** #507 enables `ActiveWidthSchedulePass` after peephole fusion and squeeze in the default pipeline. Trajectory compilation continues to exclude it. This default change remains subject to the performance decision below.
- **Local validation reported in #507:** 1,016 Release CTest checks, 1,015 Debug CTest checks, and 2,218 Python tests passed, with 84 Python skips. Coverage includes independent Aer/Stim checks, noisy scheduling equivalence, structural reference checks across word boundaries, and the expanded shared CPU sampling tests. Strict documentation and pre-commit checks also passed.
- **Current CI:** the [CI gate](https://github.com/unitaryfoundation/clifft/actions/runs/35909655053) passes on PR head `bf2276c4`. The PR description's pending-rerun item is stale. The [advisory performance canary](https://github.com/unitaryfoundation/clifft/pull/507#issuecomment-5782379529) could not produce a comparison, so it supplies no performance evidence for this head.
- **Remaining:** review the consolidated implementation, resolve the default-pipeline policy against the evidence, and merge the production work with matching documentation and validation.

## Production scope

1. **Structural active-width analysis.** Track the dormant stabilizer subspace and predict the planner's width trace without amplitude simulation. Expose `active_width_trace` and useful scheduling statistics for inspection.
2. **Sound reordering across Pauli noise.** Preserve logical noise positions and compute the appropriate symbolic sign corrections when operations cross noise sites. Validate metadata and legal schedules before execution; preserve classical dependencies, source maps, and the expectation-value/instrument barriers. Noise events must be available before their first use, including when sampling is deferred.
[...]

PR #460, opened 2026-09-01
feat(planner): resolve noise signs from logical positions

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`.

Pauli noise is presampled: every `NOISE` channel gets a `SymbolKind::Presampled` symbol whose value exists before the action stream, so a `T_GATE`, `PHASE_ROTATION`, or `MEASURE` moved across a `NOISE` op only needs its sign corrected by an affine function of that site's symbols (`R_p E = E R_{+-p}`, `E Pi_r(p) = Pi_{r xor [E,p]}(p) E`); it does not need the noise itself resampled at a different point. `HirModule::logical_noise_prefix` is the bookkeeping that makes this expressible: `logical_noise_prefix[i]`, when non-empty, is the number of `NOISE` ops that logically precede `ops[i]` in the original circuit. It is a parallel side vector in the style of `source_map` -- empty means schedule semantics (today's behavior, unchanged), and passes that reorder, delete, or strip noise from ops keep it consistent the same way they already keep `source_map` consistent. Noise sites themselves never move, so a logical position is always a plain site count.

The planner change is the enabling step. `plan_sampling` validates the vector up front (size, in-range entries, and that only the three movable op types may disagree with their schedule count -- `NOISE` itself and every barrier must match it exactly, which is also what proves their correction interval is always empty). A symbol prepass precomputes, for each `NOISE` site, the first symbol id its nonzero channels receive under the unchanged schedule-order allocation, so plans without the vector remain byte-for-byte identical. After resolving an operation's sign the usual way, the planner XORs in the channel symbols of every site between its schedule and logical positions: such a site's frame contribution is either missing (the operation moved earlier than the site) or spurious (it moved later), and XORing the same site back in fixes both directions. `StatevectorSqueezePass` and `PeepholeFusionPass` carry entries through their swaps, rotates, and deletions; `PeepholeFusionPass` additionally refuses to run at all when the vector disagrees with schedule position anywhere, because its virtual S/Pauli absorption conjugates later masks in schedule order and could move a noise-site mask into a different Clifford frame than a stale logical position was recorded against. `RemoveNoisePass` and `DropNonUnitaryPass` clear the vector outright, since both strip every `NOISE` op unconditionally.

## Validation

- Inertness: materializing `logical_noise_prefix` (schedule counts made explicit) never changes `plan_sampling(hir).inspect()`, checked on every `tests/fixtures/*.stim` fixture (raw, peephole-only, and production-pipeline HIR) and on 200 seeded random noisy circuits.
- Exact sign correction: four hand-built cases (a measurement and a rotation each moved earlier than their noise site, a rotation moved later, and a `DEPOLARIZE1` site whose two anticommuting channels must both fold in) check the corrected plan's sign expression against the original's, and against a naive vector-cleared plan that is shown to get it wrong.
- Sampling equivalence: 30 seeded random noisy circuits and the `coherent_d3_r3` fixture are reordered by a test-only walk that moves movable ops across `NOISE` sites (bypassing today's overly conservative `can_swap` noise clause) or across ordinary `can_swap`-legal neighbors, then sampled 20000 shots before and after with different seeds; every record column mean, every detector column mean, and every consecutive-pair record parity agree within a 6-sigma cross-binomial tolerance.
- Legality oracle: for 30 seeded random noisy circuits, a fixed noise realization is built by replacing each fired noise line with its explicit Pauli and dropping the rest; the same reordering, mapped onto the resulting noise-free HIR by skipping the removed `NOISE` ops, is checked to satisfy `can_swap` for every pair it inverted relative to original order -- i.e. it is a legal reordering of the noise-free circuit, which is the soundness argument from the issue turned into a test.
- Pass maintenance and validation: `StatevectorSqueezePass` keeps the vector parallel and attached to its op through a real reorder; `PeepholeFusionPass` fuses normally on a consistent vector and leaves the HIR untouched on an inconsistent one; `RemoveNoisePass` and `DropNonUnitaryPass` clear it; malformed vectors (wrong size, a barrier entry that disagrees with its schedule count, an entry above the noise-site count) throw `std::invalid_argument`.
- As a sanity check on the test suite itself, I temporarily disabled the correction and reran: exactly the four hand-built sign-correction tests and the sampling-equivalence test failed (33 assertions), while inertness, legality-oracle, pass-maintenance, and validation tests kept passing, as expected since they don't depend on the correction being active.

## Test commands

- `cmake -B build-debug -DCMAKE_BUILD_TYPE=Debug -G Ninja && cmake --build build-debug -j8` -- succeeds, no warnings.
- `ctest --test-dir build-debug -j8` -- `100% tests passed, 0 tests failed out of 947`.
- `SKBUILD_CMAKE_BUILD_TYPE=Debug uv pip install -e . --no-cache --reinstall-package clifft --python .venv/bin/python` -- installs a Debug extension (`with debug_info, not stripped`), version stamp matches the pushed tip (`0.9.1.dev33+gee4325cf5`, non-dirty).
- `SKBUILD_CMAKE_BUILD_TYPE=Debug uv run --python .venv/bin/python pytest tests/python -x -q` -- `1385 passed, 7 skipped`.
- `uv run --frozen --only-group dev pre-commit run --all-files --show-diff-on-failure` -- all hooks pass.

### DIFF
--- a/src/clifft/frontend/hir.h
+++ b/src/clifft/frontend/hir.h
@@ -460,6 +460,33 @@ struct HirModule {
     /// invalidated the map for that op.
     std::vector<std::vector<uint32_t>> source_map;
 
+    // Every NOISE channel is presampled: its symbol's value exists before
+    // the action stream, not at the schedule position where the planner
+    // happens to consume it. That is what lets an operation cross a NOISE
+    // op at all -- the crossing only changes which presampled symbols the
+    // planner must fold into the operation's sign, not when noise is
+    // sampled. logical_noise_prefix[i], when non-empty, is the number of
+    // NOISE ops that logically precede ops[i] in the original circuit, and
+    // the planner resolves ops[i]'s noise-dependent sign from that count
+    // instead of from ops[i]'s position among the NOISE ops before it in
+    // `ops`. Noise sites themselves never move, so a logical position is
+    // always expressible as a plain site count.
+    //
+    // Empty means schedule semantics: every operation's logical position
+    // equals its schedule position, i.e. today's behavior with nothing
+    // reordered across noise.
+    //
+    // Contract for passes, mirroring source_map: an entry travels with its
+    // operation, and a pass that deletes an operation drops its entry. A
+    // pass that removes NOISE ops must clear the vector entirely, because
+    // once the sites are gone there is nothing left for a stale entry to
+    // correct for. A pass that absorbs a virtual Clifford gate into later
+    // operation and noise-site masks must not run while an entry disagrees
+    // with its schedule position, because the correction it implies
+    // compares an operation's mask against noise-site masks that the frame
+    // change may have moved into a different basis.
+    std::vector<uint32_t> logical_noise_prefix;
+
     std::optional<Tableau> final_tableau;
 
     // Hidden measurement slot trace() assigned to the requested node's
@@ -480,6 +507,31 @@ struct HirModule {
             }
         }
         return true;
+    }
+
+    /// True when logical_noise_prefix is materialized and parallel to ops.
+    /// An empty ops list is never considered materialized, since an empty
+    /// vector cannot then be distinguished from the schedule-semantics
+    /// sentinel.
+    [[nodiscard]] bool has_logical_noise_prefix() const {
+        return !ops.empty() && logical_noise_prefix.size() == ops.size();
+    }
+
+    /// Fills logical_noise_prefix with each operation's schedule-order
+    /// noise count, making today's implicit schedule semantics explicit.
+    /// A no-op when the vector is already non-empty.
+    void materialize_logical_noise_prefix() {
+        if (!logical_noise_prefix.empty()) {
+            return;
+        }
+        logical_noise_prefix.reserve(ops.size());
+        uint32_t schedule_count = 0;
+        for (const HeisenbergOp& op : ops) {
+            logical_noise_prefix.push_back(schedule_count);
+            if (op.op_type() == OpType::NOISE) {
+                ++schedule_count;
+            }
+        }
     }
 
     // --- Mask accessors ---
--- a/src/clifft/optimizer/commutation.h
+++ b/src/clifft/optimizer/commutation.h
@@ -2,23 +2,12 @@
 
 #include "clifft/frontend/hir.h"
 #include "clifft/util/mask_view.h"
+#include "clifft/util/symplectic.h"
 
 #include <cassert>
 #include <cstdint>
 
 namespace clifft {
-
-/// Symplectic inner product over runtime-width mask views. Returns true if
-/// the two Pauli strings anti-commute. All four views must share num_words().
-inline bool anti_commute(MaskView x1, MaskView z1, MaskView x2, MaskView z2) {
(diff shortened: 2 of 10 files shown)


bachase:
Reviewed against `planner.cc` and verified independently at `bb566b17`.

- The correction interval `[min(c, m), max(c, m))` is right in both directions: a site the operation moved ahead of is missing from the frame and gets XORed in; a site it moved behind is spurious in the frame and the same XOR removes it. Both sides use the initial-frame body and the site's unsigned channel masks, so they cancel exactly.
- Symbol bases from the prepass mirror the schedule-order allocation (one per `MEASURE`, `READOUT_NOISE`, `INSTRUMENT`; one per nonzero channel per `NOISE`), and the assert in the `NOISE` case turns any future drift into a Debug failure rather than a silent sign error.
- Validation pins every non-movable operation to its schedule position, so barriers and classical ops can never receive a correction, and `has_logical_noise_prefix()` plus the empty-vector sentinel keep the default path byte-for-byte unchanged (the inertness tests confirm this on every fixture and pipeline variant).
- Peephole refusing to run on an inconsistent vector is the right conservative choice for now; iterating fusion after scheduling is a later concern.
- Local Debug build in a separate worktree: `ctest` 947/947 at the new head, `[logical_noise_prefix]` all 17 cases passing. The teeth check (disabling the correction fails exactly the sign and sampling-equivalence tests) is the kind of evidence these tests should carry.

The shared `util/symplectic.h` resolves the one duplication I asked about. No further changes requested.

bachase:
Follow-ups in progress from the external review: the inertness helper will catch `std::overflow_error` only instead of every exception, and two exact sign-correction cases are being added for a multi-qubit site (`DEPOLARIZE2`) and a site with a zero-probability channel (`PAULI_CHANNEL_1` with a zero entry), plus random-circuit sampling equivalence with both.

bachase:
Verified the review follow-ups at `31fb245f`: local Debug `ctest` 950/950. The narrowed catch surfaced no hidden exception on any fixture, the DEPOLARIZE2 case checks the correction against the fifteen channels' anticommutation term for term, and the PAULI_CHANNEL_1 case confirms zero-probability channels get no symbol and the site's symbol base indexes only the nonzero ones. No further changes requested.

bachase:
Follow-up in progress from the fresh review: an explicit fast path when `logical_noise_prefix` is absent, so the default compile path skips the symbol-base computation and the per-operation correction entirely; the canary's 7.4 percent `compile_plan_cultivation_d5` regression is the target, with base, before, and after medians to be posted.

bachase:
Verified the fast path at `5098a1b3`: local Debug `ctest` 950/950. The correction is skipped whenever the logical and schedule noise positions agree and the symbol-base vector is only built when a logical prefix is materialized, so the default compile path does the work it did before this PR. The one-millisecond effect is below this host's noise floor, so the canary on this PR's next run is the arbiter for the 7.4 percent regression it reported.

bachase-agent:
Superseded by #507, which consolidates the final active-width implementation onto main in five ordered review commits. Please continue review at https://github.com/unitaryfoundation/clifft/pull/507; this PR remains available as development history.

PR #459, opened 2026-09-01
feat(optimizer): add structural active-width analysis

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`.

## Summary

Implements PR 1 of #458: a non-mutating structural active-width analysis for the Heisenberg IR.

- `DormantSubspace` (`src/clifft/optimizer/active_width_analysis.h/.cc`) tracks the unsigned isotropic subspace `S` of dormant stabilizer generators directly in HIR (initial-frame) coordinates as a bit-packed GF(2) generator list, with `commutes_with_all`, `contains`, `apply_rotation`, and `apply_measurement`. Anticommute/promote is O(dim S) bit-parallel generator scans; `contains()` membership uses a GF(2) echelon basis that is rebuilt lazily and only when a change has actually invalidated it, so a run of promotions between two membership queries costs nothing extra.
- `analyze_active_width(const HirModule&)` walks the HIR once, classifying every op into a `WidthEffect` (`None`, `RotationStabilizer`, `RotationNeutral`, `RotationPromote`, `MeasureClassical`, `MeasureDormantRandom`, `MeasureActive`, `InstrumentClassical`, `InstrumentActive`, `InstrumentActivate`, `InstrumentDormantTrap`) and returns an `ActiveWidthTrace` (`initial_width`, `peak_width`, `final_width`, one `WidthTransition` per op). It never mutates the HIR, never builds a coordinate frame, and never throws on width overflow (unlike the planner, which enforces `kDenseActiveWidthLimit`; this is a structural predictor, not a plan).
- The instrument branch mirrors `process_instrument` exactly, including the `neglect_instrument_damping` / equal-`p_fire` trap condition.

## Validation

A differential test (`tests/test_active_width_analysis.cc`) compares `analyze_active_width` against `clifft::sampling::plan_sampling` on the same HIR: peak width, the full sequence of width changes in op order (from `PlannedAction::active_before/active_after` vs. the trace's transitions), and per-effect action counts (`PromoteDormantRotation`/`RotationPromote`, `MeasureActivePauli`/`MeasureActive`, `MeasureDormantRandom`/`MeasureDormantRandom`, `RecordClassical`/`MeasureClassical`, `RotateActivePauli`/`RotationNeutral`).

[...]

PR #461, opened 2026-09-01
feat(optimizer): schedule dependence relation and apply_schedule

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`. Stacked on #459 and #460; the diff includes them until they merge.

## Semantics

`ScheduleDependence::build(hir, options)` derives a conservative DAG over op indices `0..N-1` with edges `i -> j` for `i < j`. Only `T_GATE`, `PHASE_ROTATION`, and `MEASURE` ever move; every other op type keeps its original relative order via a chain over consecutive fixed ops, regardless of what `can_swap` would say about any one pair (this keeps every `NOISE`, `DETECTOR`, `OBSERVABLE`, `READOUT_NOISE`, `CONDITIONAL_PAULI`, `EXP_VAL`, and `INSTRUMENT` op fixed, matching what the sampling planner already requires). For every remaining pair, an edge is added unless `can_swap(ops[i], ops[j], hir)` allows it, or, when `noise_transparent` is set, one of the two is a `NOISE` op and the other is movable. `is_linear_extension` checks a candidate order directly against the CSR adjacency (every op once, every predecessor before its successor); legal schedules are exactly the linear extensions, i.e. the Mazurkiewicz trace class of the original order under this relation, since `can_swap` is symmetric.

Noise transparency is sound because every `NOISE` channel is presampled (its symbol exists before the action stream runs) and because the planner (#460) resolves an operation's noise-dependent sign from its logical position via `HirModule::logical_noise_prefix` rather than its schedule position. `apply_schedule` materializes that vector before permuting only when the relation is noise-transparent: without transparency, the only way a movable op crosses a `NOISE` site is because it already commutes with every channel there (a zero-correction crossing), so materializing would just leave behind a vector `PeepholeFusionPass` reads as schedule-inconsistent and refuses to fuse across, for no benefit -- skipping it keeps that pass usable. `apply_schedule` throws `std::invalid_argument` unless `order` is a linear extension (including a defensive check that `dependence` was built against an HIR with the same op count as the one being mutated, beyond the literal "not an extension" case), and permutes `ops`, `source_map`, and `logical_noise_prefix` together so every side-table entry travels with its op.

## Other deliverables

- `DormantSubspace::generators()` (`active_width_analysis.h/.cc`): unsigned-body generator list, needed by the confluence test to compare two subspaces reached via different linear extensions.
- Registered `schedule_dependence.cc` in both `src/clifft/CMakeLists.txt` and `src/wasm/CMakeLists.txt`.
- `tests/sampling_equivalence_helpers.h`: the task allowed extracting `test_logical_noise_prefix.cc`'s random noisy-circuit generator and cross-binomial sampling-equivalence checker into a shared header rather than duplicating them "if you can do so without changing that test file's behaviour" -- did so; `test_logical_noise_prefix.cc` now pulls both in via `clifft::test` and its own tests are unchanged (same assertions, same seeds, same pass/fail).
- `tests/test_schedule_dependence.cc`: see Validation below.

## Validation

- **Relation matches `can_swap`**: 200 seeded random noisy circuits (4-10 qubits, T/T_DAG/R_Z, M/MX/MR/R, X_ERROR/DEPOLARIZE1, DETECTORs), checked pairwise against `can_swap` under both `noise_transparent` settings, plus the fixed-op consecutive chain checked separately.
- **Random linear extensions**: a seeded Kahn's-algorithm-with-random-tie-break draws a linear extension; checked accepted, fixed ops stay in relative order, and a deliberately inverted dependent pair is rejected.
- **Confluence**: 100 seeded random circuits x 2 random linear extensions x both relation options; the final `DormantSubspace` (replayed the same way `analyze_active_width` does) agrees in active width and generator containment both ways. This never failed during development -- no theory bug to report.
- **Four-operation regression** (from the theory writeup): `R_XX(0.3), R_ZY(0.3), M_YY, M_YI` reorders to `[1, 2, 0, 3]`, dropping peak active width from 2 to 1; `analyze_active_width` and `plan_sampling` agree on both peaks.
- **`apply_schedule`**: rejects a non-extension and a dependence/HIR op-count mismatch; carries `source_map` and `logical_noise_prefix` with their ops through a random extension; a noise-transparent random extension is sampling-equivalent to the original (20000 shots, 6-sigma cross-binomial tolerance, distinct seeds per program) on 20 random circuits plus `coherent_d3_r3.stim`.

## Test commands and results

```
# 100% tests passed, 0 tests failed out of 960

./build-debug/tests/clifft_tests "[schedule_dependence]"
# All tests passed (276269 assertions in 7 test cases)

just build-wasm   # docker emsdk:3.1.74, confirms src/wasm/CMakeLists.txt registration
```

### DIFF
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -14,6 +14,7 @@ add_library(clifft_core STATIC
     optimizer/peephole.cc
     optimizer/pass_factory.cc
     optimizer/remove_noise_pass.cc
+    optimizer/schedule_dependence.cc
     optimizer/statevector_squeeze_pass.cc
     sampling/plan.cc
     sampling/plan_inspection.cc
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -8,6 +8,7 @@ add_library(clifft_core STATIC
     tableau/tableau.cc
     frontend/frontend.cc
     optimizer/hir_pass_manager.cc
+    optimizer/active_width_analysis.cc
     optimizer/drop_non_unitary_pass.cc
     optimizer/commutation.cc
     optimizer/peephole.cc
--- a/src/clifft/frontend/hir.h
+++ b/src/clifft/frontend/hir.h
@@ -460,6 +460,33 @@ struct HirModule {
     /// invalidated the map for that op.
     std::vector<std::vector<uint32_t>> source_map;
 
+    // Every NOISE channel is presampled: its symbol's value exists before
+    // the action stream, not at the schedule position where the planner
+    // happens to consume it. That is what lets an operation cross a NOISE
+    // op at all -- the crossing only changes which presampled symbols the
+    // planner must fold into the operation's sign, not when noise is
+    // sampled. logical_noise_prefix[i], when non-empty, is the number of
+    // NOISE ops that logically precede ops[i] in the original circuit, and
+    // the planner resolves ops[i]'s noise-dependent sign from that count
+    // instead of from ops[i]'s position among the NOISE ops before it in
+    // `ops`. Noise sites themselves never move, so a logical position is
+    // always expressible as a plain site count.
+    //
+    // Empty means schedule semantics: every operation's logical position
+    // equals its schedule position, i.e. today's behavior with nothing
+    // reordered across noise.
+    //
+    // Contract for passes, mirroring source_map: an entry travels with its
+    // operation, and a pass that deletes an operation drops its entry. A
+    // pass that removes NOISE ops must clear the vector entirely, because
+    // once the sites are gone there is nothing left for a stale entry to
+    // correct for. A pass that absorbs a virtual Clifford gate into later
+    // operation and noise-site masks must not run while an entry disagrees
+    // with its schedule position, because the correction it implies
+    // compares an operation's mask against noise-site masks that the frame
+    // change may have moved into a different basis.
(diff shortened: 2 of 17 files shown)


bachase:
Reviewed and verified independently at `307d32bb`.

- Relation: the `allowed` predicate is exactly `can_swap` plus the noise-transparent waiver for `NOISE`-versus-movable pairs, barriers stay barriers through `can_swap`, and the fixed-op chain keeps every side-table-indexed operation in order. The CSR grouping and `is_linear_extension` are straightforward and the oracle tests check them against `can_swap` pairwise.
- `apply_schedule` materializing the logical prefix only for a noise-transparent relation is the right call, for the reason the header gives: a plain-`can_swap` reorder never needs a correction, and leaving the vector empty keeps peephole usable afterwards.
- The confluence test is the theorem the search in the next PR relies on (the subspace after a set of executed operations is independent of their legal order), and it held on every trial.
- Local Debug build in a separate worktree: `ctest` 960/960, `[schedule_dependence]` 7 cases passing; the two sampling-equivalence sections take about six seconds in Debug, in line with the existing fuzz tests.

Stacked on #459 and #460 as noted; no changes requested.

bachase:
Superseded by the re-cut review series: #460 (PR A, planner logical-position noise signs) and #471 (PR B, analysis, dependence relation, and scheduling pass, consolidated and simplified). Closing without merge; the content lives on in #471.

PR #462, opened 2026-09-02
feat(optimizer): exact active-width schedule search with certificates

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`. Stacked on #461; the diff includes #459, #460, and #461 until they merge.

## Algorithm

`search_width_schedule` replays HIR ops through a `DormantSubspace` exactly as `analyze_active_width` does, over the dependence DAG `ScheduleDependence::build` produces. An op is *ready* once every predecessor has executed; a ready op is *expanding* if executing it would raise the active width `k` (a `T_GATE`/`PHASE_ROTATION` whose axis does not commute with the current subspace `S`, or an `INSTRUMENT` in the Activate branch) and *non-expanding* otherwise (any `MEASURE`, any commuting rotation, and every other op type). *Closure* repeatedly executes the lowest-index ready non-expanding op until none remains; because independent GF(2) updates commute, the resulting subspace and executed-op set are order-independent, so a search state is fully identified by its set of executed ops rather than by how it was reached. The closure theorem (some peak-minimizing schedule executes every ready non-expanding op immediately) means the search only has to branch on which ready expanding op fires next -- an `INSTRUMENT`, being a positional barrier under `ScheduleDependence` (`can_swap` refuses to reorder anything across one), is always the only ready op when it is ready, so it is never actually a branch point, just a forced move that can fail the node if it would exceed the current threshold.

Threshold feasibility for a bound `B` is a depth-first search from the initial closed state: at each node the candidates are the ready expanding ops whose execution would not push `k` above `B`, tried in the order most likely to succeed (lowest width after that candidate's own closure sweep, ties by lower op index); a node with no viable candidate is infeasible and is memoized by its exact executed-op bitset (an `unordered_set<vector<uint64_t>>` with a real hash used only for bucketing, never as an identity) so any other branch reaching the same set of executed ops is pruned for free. The outer loop starts from the incumbent's own peak and threshold-searches downward, sharing one node budget and one failed-node memo across every threshold it tries (a set proven infeasible at a looser bound is also infeasible at any tighter one, since the tighter bound's candidate list is a subset of the looser one's), lowering `upper_bound` to each witness's actual achieved peak (which can undercut the threshold that found it) and raising `lower_bound` to `threshold + 1` on the first exhaustive failure, until the two meet (`optimal()`) or the shared budget runs out.

## Certificate scope

`optimal()` certifies a minimum over the trace class of the exact HIR passed in, under the given `ScheduleDependence` relation, scored by the sampling planner's own structural width model. It says nothing about schedules reachable only through a rewrite that exposes new gate fusion, about amplitude-level stabilizers the structural planner does not track, or about any other circuit that merely samples the same distribution as this one.

## Validation

- **Brute force agreement**: 150 seeded random noisy circuits with at most 9 movable ops (skipping the rare circuit whose linear-extension count still exceeds a 3000-leaf enumeration cap kept low to hold this test to ~8s -- see the code comment for why the movable-op cap alone does not bound enumeration cost), enumerated exhaustively and cross-checked against the search under both relation options: `upper_bound`, `lower_bound`, and `best_order`'s actual peak all agree with the brute-force minimum, and `optimal()` holds. This also exercises the closure theorem directly: the search only ever branches on expanding ops, yet it always lands on the same optimum as trying every legal order.
- **Four-operation regression**: the `R_XX, R_ZY, M_YY, M_YI` HIR from `test_schedule_dependence.cc` drops from incumbent peak 2 to a certified optimal 1, with `apply_schedule(best_order)` reproducing the `0, 1, 1, 1, 0` trace.
- **Never worse than the incumbent**, plus **budget behaviour** (an exhausted budget reports `budget_exhausted`, `upper_bound == incumbent_peak`, `lower_bound` at the order-invariant final width, and never throws -- checked on a purpose-built two-independent-widget circuit, since the four-operation circuit above turns out to resolve within a single search node regardless of budget).
- **coherent_d3_r3 fixture certificates**: after `PeepholeFusionPass` + `StatevectorSqueezePass`, the search certifies the incumbent peak 5 as optimal without noise transparency (71 nodes explored), and certifies 4 with it (64 nodes explored) -- matching the corpus table in #458 exactly, and well inside the 200000-node budget.
- **Sampling equivalence**: applying the noise-transparent `best_order` to a copy of coherent_d3_r3 and of 10 random noisy circuits preserves sampling statistics (20000 shots, 6 sigma) against the original.
- Python: `active_width_trace`'s peak matches `clifft.lower(hir).peak_active_width` on all 8 in-tree fixtures; `search_width_schedule` reproduces the same coherent_d3_r3 certificate (incumbent 5, bounds 4/4) and, with `apply=True`, the lowered program's `peak_active_width` drops to 4 while `clifft.sample` statistics still match the unoptimized program; a separate test confirms `apply=False` leaves the HIR untouched.

A real bug did turn up during development: iterating `SearchFrontier`'s live `std::set` of ready ops while a nested call mutated and restored that same set (to rank candidates) invalidated the outer iterator and segfaulted. That was a memory-safety bug in the implementation, not a disagreement between the search and the brute-force oracle on any actual answer -- fixed by snapshotting the ready set into a plain vector before iterating. Once fixed, the brute-force cross-check never disagreed with the search's claimed bounds.

## Test commands and results

```
# 100% tests passed, 0 tests failed out of 966

./build-debug/tests/clifft_tests "[width_search]" --durations yes
# All tests passed (1009 assertions in 6 test cases); slowest case (brute
# force agreement) 8.1s, every other case under 0.7s

uv venv .venv
# 1389 passed, 7 skipped

just build-wasm   # docker emsdk, confirms src/wasm/CMakeLists.txt registration
```

### DIFF
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -14,6 +14,7 @@ add_library(clifft_core STATIC
     optimizer/peephole.cc
     optimizer/pass_factory.cc
     optimizer/remove_noise_pass.cc
+    optimizer/schedule_dependence.cc
     optimizer/statevector_squeeze_pass.cc
     sampling/plan.cc
     sampling/plan_inspection.cc
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -8,6 +8,7 @@ add_library(clifft_core STATIC
     tableau/tableau.cc
     frontend/frontend.cc
     optimizer/hir_pass_manager.cc
+    optimizer/active_width_analysis.cc
     optimizer/drop_non_unitary_pass.cc
     optimizer/commutation.cc
     optimizer/peephole.cc
--- a/src/clifft/CMakeLists.txt
+++ b/src/clifft/CMakeLists.txt
@@ -15,6 +15,7 @@ add_library(clifft_core STATIC
     optimizer/pass_factory.cc
     optimizer/remove_noise_pass.cc
     optimizer/schedule_dependence.cc
+    optimizer/active_width_search.cc
     optimizer/statevector_squeeze_pass.cc
     sampling/plan.cc
     sampling/plan_inspection.cc
--- a/src/clifft/frontend/hir.h
+++ b/src/clifft/frontend/hir.h
@@ -460,6 +460,33 @@ struct HirModule {
     /// invalidated the map for that op.
     std::vector<std::vector<uint32_t>> source_map;
 
+    // Every NOISE channel is presampled: its symbol's value exists before
+    // the action stream, not at the schedule position where the planner
+    // happens to consume it. That is what lets an operation cross a NOISE
+    // op at all -- the crossing only changes which presampled symbols the
+    // planner must fold into the operation's sign, not when noise is
+    // sampled. logical_noise_prefix[i], when non-empty, is the number of
+    // NOISE ops that logically precede ops[i] in the original circuit, and
+    // the planner resolves ops[i]'s noise-dependent sign from that count
+    // instead of from ops[i]'s position among the NOISE ops before it in
+    // `ops`. Noise sites themselves never move, so a logical position is
+    // always expressible as a plain site count.
+    //
+    // Empty means schedule semantics: every operation's logical position
+    // equals its schedule position, i.e. today's behavior with nothing
+    // reordered across noise.
+    //
+    // Contract for passes, mirroring source_map: an entry travels with its
+    // operation, and a pass that deletes an operation drops its entry. A
+    // pass that removes NOISE ops must clear the vector entirely, because
+    // once the sites are gone there is nothing left for a stale entry to
+    // correct for. A pass that absorbs a virtual Clifford gate into later
+    // operation and noise-site masks must not run while an entry disagrees
+    // with its schedule position, because the correction it implies
+    // compares an operation's mask against noise-site masks that the frame
+    // change may have moved into a different basis.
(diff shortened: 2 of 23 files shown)


bachase:
Reviewed and verified independently at `50466e03`.

- Soundness of the certificate: the memo key is the exact executed-op bitset, a budget-starved subtree returns unknown and is never memoized, and sharing failed nodes across strictly decreasing thresholds is valid because a node's candidate set at a tighter bound is a subset of its set at a looser one. The closure step re-evaluates readiness after every executed op, so a rotation that becomes neutral through a collapse is picked up rather than treated as expanding.
- Barrier handling is right: an `INSTRUMENT` is the only ready op when it becomes ready, so executing it regardless of expansion is forced rather than a choice, and an over-threshold Activate simply fails the node.
- The brute-force oracle over full linear extensions is the strongest check here and agreed on every enumerated circuit under both relations; the coherent d3 r3 certificates (5 without noise transparency, 4 with it) match the prototype numbers in #458 and finish in about 70 nodes.
- Local Debug build in a separate worktree: `ctest` 966/966, `[width_search]` 6 cases in under a second each; Debug Python extension in a worktree-local venv, the new Python tests plus the compile-pass and squeeze suites, 46 passed.

Stacked on #461 as noted; no changes requested.

bachase:
Superseded by the re-cut review series. The exact search and its certificates move out of the library to a research tool under research/active_width_certificates/ on a research branch, per the discussion on #458; the analysis and dependence code it built on lives on in #471. Closing without merge.

PR #463, opened 2026-09-02
feat(optimizer): add state-aware active-width scheduling pass

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`. Stacked on #462; the diff includes #459 through #462 until they merge.

**Follow-up (2026-09-02):** two perf commits landed after the initial review measured the pass on the clifft-paper corpus in Release: `7a7e186c` splits beam expansion into a cheap scoring phase (execute/undo on the parent's own frontier plus a single scratch `DormantSubspace` copy per candidate, no full `BeamState` clone) and a materialize phase that only clones for the `beam_width` survivors, and `fbab2cb0` defaults `exact_node_budget` to 0 since repair never lowered a peak the beam had not already reached on that corpus. Together these took `coherent_d5_r5.stim` at `beam_width=8` from 24.6s to 0.95s in Release with bit-identical results (verified by running the pre-existing test suite unchanged against the scoring-phase commit before touching the budget default). Numbers below are updated for both changes; results (peaks, dense work) are unchanged from the original measurements since both follow-ups are cost-only.

## Algorithm

`ActiveWidthSchedulePass` reduces peak active width, then a dense-work proxy, by beam-searching the closure/readiness trace class `active_width_search.h` already characterizes:

1. **Incumbent.** `analyze_active_width(hir)` on the input order, before touching anything.
2. **Dependence.** `ScheduleDependence::build(hir, {noise_transparent})`.
3. **Beam search with closure, in two phases.** Starting from the initial closure, `score_candidates` ranks every ready expanding op of every beam state cheaply: `execute()`/`undo()` on the state's own shared frontier (never cloned) plus a single scratch `DormantSubspace` copy per candidate, recording the resulting op list, width-after-closure, peak, dense work, and the frontier's post-candidate executed-op bitset. Candidates are deduped by that bitset (keeping the smaller-dense-work one) and ranked by `(peak, width_after_closure, dense_work, first_op_index)` ascending; only the best `beam_width` survivors pay to materialize a full `BeamState` (`materialize_candidate`, one clone of the parent plus a deterministic replay of the already-decided op list -- Debug-only assertions cross-check that replay reaches exactly what scoring predicted). `beam_width = 1` degenerates to the greedy closure scheduler.
4. **Exact repair (optional, off by default).** If `exact_node_budget > 0`, applies the beam's order to a copy and runs `search_width_schedule` on it as a pure improver -- adopts its order only on a strictly lower peak, never reports a certificate. Defaults to 0 (disabled): see Compile-time cost below.
5. **Neutral sinking (optional).** Bubbles each `RotationNeutral` op rightward past independent, non-expanding ops (checked via `ScheduleDependence`, not a fresh commutation scan), stopping before the first dependent or expanding op. A neutral rotation never changes the subspace, so this never changes any other op's classification.
6. **Never worse.** Compares the candidate's `analyze_active_width` against the incumbent lexicographically by `(peak, dense_work)`; applies only if strictly better, otherwise the HIR is byte-for-byte untouched.

Shared machinery: `active_width_search.cc`'s closure/readiness bookkeeping (`SearchFrontier`, `run_closure`, `is_expanding`) moved into a new internal header, `active_width_closure.h/.cc`, reused by both the exact search and this pass instead of a second hand-copied implementation. `active_width_analysis.h/.cc` gained `classify_and_apply` (the per-op primitive `analyze_active_width` now loops over) and `estimate_dense_work` (the planning-proxy sum of `2^width` at each dense action, spelled out in the header).

The pass is registered `default_enabled = false`, record-order-breaking, and instrument-prefix-unstable (like `StatevectorSqueezePass`), and documented to run last in a pipeline: `PeepholeFusionPass` refuses to fuse across a `NOISE` op once a noise-transparent reorder leaves `logical_noise_prefix` disagreeing with schedule position.

## Fixture numbers (peephole + squeeze incumbent -> after this pass)

All measured with the pass's class-default options (`beam_width=8`, `exact_node_budget=0`, `noise_transparent=true`, `sink_neutral_rotations=true`) unless noted. Peaks and dense-work figures are unchanged from the original measurements -- only wall time moved.

| Fixture | Incumbent peak / dense work | Result peak / dense work | Notes |
|---|---:|---:|---|
| coherent_d3_r3 | 5 / 1390 | 4 / 442 | matches the certified-optimal peak from #462's corpus table |
| coherent_d5_r5 | 13 / 2.186e6 | 13 / 1.348e6 (ratio 0.616) | default options measured directly (Release, 0.95s); exact repair never improved on this fixture at any budget tried (0, 500, 2000, 20000), which is why it now defaults off |
| cultivation_d5 | 10 / 41710 | 10 / 33174 (ratio 0.795) | exact repair made no difference at budget 0, 500, 2000, or 20000 |
| surface_d7_r7_p001 | 0 / 0 | 0 / 0 | already optimal; `applied() == false` |

**Fixture where the pass did not reach the fixture-table's "at most half" aspiration:** coherent_d5_r5 reaches dense work ratio 0.616 (beam_width 8) to 0.627 (beam_width 1), not <= 0.5, with default options; widening the beam from 1 to 8 only narrows that gap by about 2 percentage points, and exact repair finds no improvement at all on this fixture at any budget tried, up to and including 20000 (matches #462's own corpus note that a 200000-node search "exhausted budget at 12" here). This looks like a genuine property of the scoring heuristic on this fixture (width-after-closure dominates the ranking before dense work ever breaks a tie) rather than a bug -- brute-force/closure-theorem cross-checks elsewhere in the suite (inherited from #461/#462) still hold, and the pass still cuts dense work by 38% deterministically and never regresses. Flagging for visibility rather than reworking the scoring formula, which the spec fixes.

## Validation

- Four-operation regression (`R_XX, R_ZY, M_YY, M_YI`): reorders to the `0,1,1,1,0` trace, `applied() == true`.
- Fixture peak/dense-work bounds above.
- 100-trial fuzz: `(peak, dense_work)` after the pass is never lexicographically worse than the incumbent, and whenever `applied() == false` the HIR is byte-for-byte unchanged (ops incl. type-specific payload and Pauli mask content, `source_map`, `logical_noise_prefix`).
- Determinism across repeated runs on 20 seeded circuits.
- Sampling equivalence (20000 shots, 6 sigma) against the unoptimized program on `coherent_d3_r3` and 20 seeded random noisy circuits.
- `EXP_VAL`/`INSTRUMENT` barrier regressions: the pass runs without throwing and never moves a fixed op relative to another fixed op; before/after plans have equal `num_exp_vals` and unchanged instrument mode.
- Registry: not in `default_hir_pass_manager()`, resolves via `make_hir_pass`, `is_trajectory_compatible` is false.
- Python: `HirPassManager`-mediated pipeline reaches the same peaks; pass statistics populate on the original Python-owned instance after `HirPassManager.run()` (passes have no `run()` of their own in Python -- only `HirPassManager` does); keyword construction.
- Two-phase scoring bit-identical check: the pre-existing test suite (unchanged) was run against the scoring-phase commit alone before the budget-default commit landed, confirming identical peak, dense-work, and order on every fixture and every seeded random circuit the suite covers; `materialize_candidate` additionally carries Debug-only assertions that its replay agrees with `score_candidates`' prediction, which held (no assertion failures) across every test path in both the C++ and Python suites.

**Performance judgment call:** `coherent_d5_r5` is still the slow fixture even after both follow-ups: 0.95s in Release at `beam_width=8`, versus well under 0.11s for the other three. `ScheduleDependence::build` itself is not the cause here (measured separately at 5.5ms on this fixture); the remaining cost is the beam search's own scoring work on a circuit with a genuinely wide branching factor (tens of simultaneously-ready, mutually independent expanding rotations at many steps), which `beam_width=8` scores in full at every step. In Debug, this fixture is far worse regardless of these follow-ups (still ~29s at `beam_width=8`): `SearchFrontier::execute`/`undo`'s `std::set` insert/erase pair dominates at `-O0` regardless of how little else a candidate clones, so neither follow-up meaningfully changes Debug wall time on this specific fixture. The C++ and Python fixture tests for `coherent_d5_r5` therefore still narrow the beam to `beam_width=1` to stay fast in Debug (verified to reach dense work within 2% of `beam_width=8`'s on this fixture); `cultivation_d5` and `surface_d7_r7_p001` no longer need any override now that exact repair defaults off. `coherent_d3_r3`'s fixture section now passes a nonzero `exact_node_budget` explicitly so that code path stays covered by CI.

## Test commands and results

```
# 100% tests passed, 0 tests failed out of 974; total ctest wall time ~16.8s (parallel)

./build-debug/tests/clifft_tests "[schedule_pass]" --durations yes
# All tests passed (1144 assertions in 8 test cases)
# slowest test case: "Schedule pass reaches the expected peak and dense work on fixture circuits" ~13.1s
# (coherent_d5_r5 narrows the beam to stay fast in Debug, see above); every other test case under 4.1s

uv venv .venv --python 3.13
# 1394 passed, 7 skipped in 77.6s

uv run --group docs --python .venv/bin/python mkdocs build --strict
# Documentation built in 1.88 seconds, no errors

# wasm (docker emsdk:3.1.74), verified before the two follow-up commits (neither touches wasm-relevant code):
# emcmake cmake -B build-wasm -S src/wasm -DCMAKE_BUILD_TYPE=Release -DFETCHCONTENT_QUIET=ON
# cmake --build build-wasm -j4
```

## Compile-time cost (Release, default options)

Measured with a small standalone harness against a Release build of `clifft_core` (not part of the committed test suite -- PR6 owns the real Google Benchmark measurement). All at `beam_width=8`, `exact_node_budget=0` (the class defaults):

- `coherent_d5_r5.stim`: 0.95s (was 24.6s before the two follow-up commits)
- `cultivation_d5.stim`: 0.11s (was 1.35s)
- `surface_d7_r7_p001.stim`: 0.05s (already optimal, closure resolves the whole circuit in one sweep; was 0.10s)
- `coherent_d3_r3.stim`: 0.03s (was 0.036s)

`coherent_d5_r5`'s remaining 0.95s is the beam search's own scoring cost on a fixture with a wide branching factor, not `ScheduleDependence::build` (5.5ms alone) or exact repair (off by default now). Further reduction on this specific fixture would need either a smaller default `beam_width` or a cheaper per-candidate scoring step than a `DormantSubspace` copy plus closure sweep, neither of which this pair of commits attempted -- flagging as a possible target for PR6's benchmark-driven tuning rather than changing here without a broader corpus measurement.

Assisted-by: Claude (Sonnet 5) <noreply@anthropic.com>


bachase:
Reviewed and verified independently at `2d99ec5c`: local Debug `ctest` 974/974, `[schedule_pass]` 8 cases; Debug Python extension in a worktree-local venv, 18 passed across the new test, the search test, and the compile-pass test. The algorithm is right: beam over closed states with confluence-based dedup, exact repair composed back through the beam order, neutral sinking that cannot change any other op's classification, and a strict lexicographic never-worse guard.

Throughput, Release build on this host, `clifft-paper` corpus, best of 3 timed batches after warmup (sampling only; the pass runs once at compile time):

| Circuit | Production peak / plan work / shots/s | With pass (default) | Pass wall time |
|---|---|---|---|
| coherent d3 r3 | 5 / 1247 / 1.09M | 4 / 403 / 2.18M | 27 ms |
| coherent d5 r1 | 12 / 35875 / 83k | 0 / 0 / 3.8M | 2 ms |
| coherent d5 r5 | 13 / 1.99e6 / 2082 | 13 / 1.23e6 / 3385 | 25.4 s |
| distillation | 5 / 155 / 1.15M | 3 / 71 / 1.15M | 9 ms |
| cultivation d5 | 10 / 56618 / 61k | unchanged | 1.36 s |

Two changes requested before this is ready, both about cost rather than correctness:

1. **`expand_state` clones the whole beam state for every candidate before running its closure.** That is the 25 s on d5 r5. Score candidates on the parent with `execute`/`undo` the way `simulate_candidate` does in the exact search, sort the generation, and clone only the `beam_width` survivors by replaying their recorded op lists. On d5 r5 the `beam_width = 8, exact_node_budget = 0` variant already takes 1.04 s and `beam_width = 1` 127 ms, so the target after the change is well under a second at the default beam.
2. **Exact repair never improved a corpus circuit and dominates the remaining time** (12 s of the 25 s on d5 r5, 1.2 s of the 1.36 s on cultivation). Default `exact_node_budget` to 0 and keep it as an opt-in knob; the fixture test can keep exercising it explicitly.

One observation for the follow-up tuning PR rather than this one: the prototype's one-step-lookahead greedy reached a planner dense-work proxy of 7.7e5 on d5 r5 where this beam lands at 1.23e6, so the beam's ranking key is leaving dense work on the table even though both hit peak 13. Worth a scoring experiment once the cost is down.

bachase:
Re-verified at `fbab2cb0`: local Debug `ctest` 974/974, the Debug Python extension's pass, search, and compile-pass tests 18 passed, and my Release harness reproduces the earlier peaks and planner dense-work figures bit-for-bit (d3 r3 4/403, d5 r5 13/1.23e6, distillation 3/71, cultivation unchanged) with the pass now at 20 ms, 0.93 s, 6 ms, and 0.10 s respectively. Sampling throughput matches the table above within this host's timing noise.

One thing the re-run made visible: with `beam_width = 1` the pass keeps d3 r3 at peak 5 (dense work 715), while `beam_width = 8` finds the peak-4 schedule, so the default beam is doing real work there and should stay at 8. The two-phase expansion and the zero repair budget are what I asked for; no further changes on this PR. The remaining d5 r5 cost is the branching factor itself and belongs with the scoring experiment in the next PR.

bachase:
Superseded by the re-cut review series: #460 (PR A, planner logical-position noise signs) and #471 (PR B, analysis, dependence relation, and scheduling pass, consolidated and simplified). Closing without merge; the content lives on in #471.

PR #464, opened 2026-09-02
feat(cuda): add sm_90 sampling backend with cooperative execution tiers

FarLab:
Part of #493.

## Summary

Ports the NVIDIA CUDA sampling backend from the `research/sampling-gpu-bench` study into the tree, laid out file for file like the HIP backend on `main`:

- `src/clifft/sampling/cuda/` with private `device_program.h` descriptors, an `ExecutablePlan` lowering that shares `pauli_preparation` with the CPU and HIP paths, and a retained `Sampler` with the same synchronous, bounded-workspace contract as `hip::Sampler`.
- Three execution tiers selected per program and device: `ThreadPerShot` for peak width <= 4, `BlockShared` when one shot's coefficients fit the opt-in shared memory, `BlockGlobal` otherwise. Every lane of a block replays the scalar control flow redundantly, coefficient sweeps are lane-strided, and probabilities tree-reduce through a fixed 16 KB scratch so tier selection is independent of the launch block size. Seeded rows derive from the global shot index and do not change with batch size or the concurrency cap within one tier.
- A separate `kCudaSamplingExecutorDomain` random-stream domain.
- GPU-free lowering tests in the main suite; a `clifft_cuda_tests` executable that skips without a device (`SKIP_RETURN_CODE 4`, same isolation as HIP); a `cuda-offline-build` CI job compiling `sm_90` device code in the CUDA 13.0 container.
- `docs/development/cuda-backend.md` and CUDA mentions beside the existing HIP ones.

One change outside the backend: nvcc rejects host-only flags such as `-Wall` and `-march=...`, so the root CMake wraps directory-wide compile options in `$<NOT:$<COMPILE_LANGUAGE:CUDA>>`. C++ and HIP translation units are unaffected (verified through `compile_commands.json`).

This is PR 1 of 2. PR 2 adds the `_clifft_cuda` nanobind module and the `clifft.experimental.cuda` facade, mirroring #395.

## Test plan

- [x] `pre-commit run --all-files` clean
- [x] CPU build: full suite green (939 tests, including the 9 new CUDA lowering tests)
- [x] `mkdocs build --strict`
- [x] `sampler.cu` passes clang `-x cuda` host and device syntax passes with `-Wall -Wextra`
- [x] nvcc 13.0.88 build and `clifft_cuda_tests` on an H100 SXM (Nebius via Brev, driver 580.173.02): 26/26 pass, every tier against the CPU oracle
- [x] `cuda-offline-build` CI job green (now also compiles `clifft_cuda_bench`)
- [x] Review follow-up on an H100 PCIe (driver 580 unavailable there, so CUDA 12.8 + g++ 13.3 in `nvidia/cuda:12.8.1-devel-ubuntu24.04`): `clifft_cuda_tests` 21/21 (60812 assertions), including the new shared-memory boundary cases; `clifft_cuda_bench` runs recorded in `docs/development/cuda-backend.md`


src/clifft/sampling/cuda/sampler.cu
```diff
@@ -0,0 +1,1446 @@
...
+            case ActionTag::WriteDetector: {
+                const bool outcome =
+                    evaluate_output_value(program, rows.symbols, rows.records, action);
+                if (lane.is_writer()) {
+                    rows.detectors[action.index0] = static_cast<uint8_t>(outcome);
+                }
+                if ((action.flags & kPostselected) != 0 && outcome) {
+                    discarded = true;
+                }
+                break;
```
bachase, line 596: **Codex review — [P1] Finish detector reads before later record mutations**

Every lane reads the detector's record parity, but there is no barrier before the next action can modify that record. For this circuit, with the detector postselected and a cooperative tier forced, the shot should survive:

```text
M 0
DETECTOR rec[-1]
READOUT_NOISE(1) rec[-1]
```

Lane 0 can advance to the readout flip while a later warp is still reading the detector. The later warp then discards the shot, so lanes take different paths and reach different barriers. A host scheduling probe of the actual lowered interpreter produced `lane 0 discarded=0; lane 32 discarded=1`; this was not a GPU run.

Order all detector reads before subsequent mutations and add a regression using multiple warps. The same sequence can occur after promotions in an automatically selected cooperative tier.

src/clifft/sampling/cuda/sampler.cu
```diff
@@ -0,0 +1,1446 @@
...
+            probability_one += norm(compact_nondiagonal(action, real, imag, packed, true));
+        }
+    }
+    lane.reduce_pair(probability_zero, probability_one);
+
+    const double total = probability_zero + probability_one;
+    const double epsilon = clifft::kMeasurementDustEpsilon * total;
+    bool branch;
+    if constexpr (Replay) {
+        const bool correction = evaluate(program, symbols, action.expression);
```
bachase, line 340: **Codex review — [P1] Stabilize the replay correction before writing the branch symbol**

Lowering stores `typed.outcome` here, which the planner constructs as the correction XOR the new branch symbol. Every lane evaluates it assuming that symbol is still zero, but lane 0 overwrites the symbol at line 353 without a barrier between the reads and write. A slower warp can read the updated value and select the opposite branch. For a deterministic forced-one measurement, some lanes return unreachable while others continue into collapse barriers.

A host scheduling probe of this function produced `lane 0 reachable=1; lane 32 reachable=0`; this was not a GPU run. Synchronize the correction reads before publication, or exclude the branch symbol from the replay correction during host lowering. Add deterministic forced-one replay coverage using multiple warps.

.github/workflows/ci.yml
```diff
@@ -141,6 +141,30 @@ jobs:
       - name: Run GPU-free HIP Python boundary tests
         run: uv run --no-sync pytest tests/python/test_experimental_hip.py -v
 
+  cuda-offline-build:
```
bachase, line 144: **Codex review — [P2] Add the CUDA job to the required CI gate**

Please add `cuda-offline-build` to `ci-gate.needs`, alongside `hip-offline-build`. It is currently missing. Main's branch protection requires only `check-title` and `ci-gate`, with no additional repository rulesets, so CUDA compilation or the optional extension can fail while the required gate still passes.

tests/test_cuda_sampler.cc
```diff
@@ -0,0 +1,672 @@
...
+                         Catch::Matchers::WithinAbs(expected.exp_vals[index], tolerance));
+        }
+    }
+}
+
+TEST_CASE("CUDA execution tiers agree with the CPU on a wide program") {
+    const SamplingPlan plan = plan_from(kWideCircuit);
+    const CudaExecutablePlan executable(plan);
+    const CpuExecutablePlan cpu_executable(plan);
+    REQUIRE(executable.peak_active_width() == 6);
```
bachase, line 422: **Codex review — Exercise the actual cooperative tier boundaries**

The device tests stop at width six; the cultivation width-ten test only lowers the circuit. Forcing `BlockGlobal` on this small state does not cover automatic shared/global selection, large coefficient slabs, or memory-derived concurrency limits.

Add device cases on both sides of the shared-memory boundary for FP64 and FP32, including a naturally selected `BlockGlobal` case. Exercise cooperative noise/postselection and observe the evolved state after collapse. Record the hardware run commands and results: the CUDA CI job skips all fifteen hardware cases, so its green status cannot establish this coverage.

docs/development/cuda-backend.md
```diff
@@ -0,0 +1,203 @@
...
+| `BlockShared` | one thread block | opt-in dynamic shared memory | shot fits the device's shared-memory budget |
+| `BlockGlobal` | one thread block | global memory, one slab per resident block | otherwise |
+
+In the cooperative tiers every thread of a block evaluates the scalar control
+flow (random draws, branch selection, expressions) redundantly from identical
+inputs, so branch decisions need no broadcast storage. Coefficient sweeps are
+strided across the block, measurement probabilities are tree-reduced through a
+fixed shared-memory scratch, and byte outputs are written by lane 0. The
+per-shot RNG derives from the global shot index, so the tier, block size, and
+concurrency cap cannot change seeded rows within one tier. Different tiers sum
```
bachase: **Codex review — Limit the reproducibility guarantee to what the implementation establishes**

Keeping the global shot index fixed preserves RNG stream selection, but changing `block_size` changes the lane-strided sums and reduction tree even within one cooperative tier. That can change floating-point expectation values and measurement thresholds. Please remove the unconditional block-size guarantee, or implement and test a block-size-independent reduction order. State reproducibility for a fixed tier, precision, and block size, while separately documenting the supported batch-size and concurrency-cap invariance.

bachase (commented):
**Codex review**

Additional changes requested alongside the inline comments:

- **Include CUDA in the formatting hook, then format `sampler.cu`.** `.pre-commit-config.yaml` overrides clang-format's supported types with `types_or: [c++, c]`. I verified that pre-commit tags this file as `cuda`, without `c++`, so both local pre-commit and CI skip it. Change the filter to `types_or: [c++, c, cuda]` and rerun the hook. The current green lint result does not check this file's formatting.
- **Add reproducible CUDA performance evidence for an existing workload.** Run a cultivation or QEC fixture against the CPU reference and report correctness, construction/upload time, retained sampling time, memory use, and the hardware/toolchain/commands. Include enough evidence to explain the width-four cutoff and concurrency default. The existing performance-canary results measure CPU behavior and do not validate these CUDA choices.
- **Tighten the kernel comments.** Remove the opening claim that conformance results transfer between interpreters; shared action semantics do not establish the correctness of a new parallel execution model. Also correct the claim that six coordinates keep every lane busy when the tests launch 256-thread blocks.
- **Align both stack commits with the contribution guide.** Keep `Assisted-by`, but remove the AI `Co-Authored-By` trailers as required by `docs/development/contributing.md`.

src/clifft/sampling/cuda/sampler.cu
FarLab: Fixed in 95ae798: WriteDetector now ends with a block barrier, so no lane can start a later record mutation while another warp still reads the detector. Regression case "CUDA postselected detectors read records before a later flip" puts this exact sequence on hardware in every tier, both precisions, block sizes 64/256/1024. It passes on an H100 PCIe with nvcc 12.8. The 1024-lane variant also exposed a separate problem on that toolchain (the thread-per-shot kernel compiled to 80 registers, so a 1024-thread launch was refused); f44cb88 adds `__launch_bounds__(kMaxBlockSize)` to both kernels so the documented block-size range is a compile-time promise.

src/clifft/sampling/cuda/sampler.cu
FarLab: Fixed in 95ae798: in both MeasureActivePauli and MeasureDormantRandom every lane's correction read is barriered before lane 0 publishes the branch symbol. Regression case "CUDA replay resolves a deterministic branch identically in every tier" forces a numerically deterministic active measurement on a width-7 state (several warps) 32 times per tier and precision, asserting the forced-zero path is unreachable and the forced-one path reaches the collapse. Passes on hardware.

.github/workflows/ci.yml
FarLab: Done in 95ae798: cuda-offline-build is in ci-gate.needs. bb9d441 also has that job compile the new clifft_cuda_bench target so it cannot rot.

tests/test_cuda_sampler.cc
FarLab: Added in 64ec703. The boundary depends on the device's opt-in limit, so two new cases locate it at runtime with selected_tier() over a width-k magic-state family and run the width just below and at it with automatic selection, giving a BlockShared case and a naturally selected BlockGlobal case for FP64 and FP32. In each: forced replay pins the path log-probability and the post-collapse EXP_VALs against the CPU executor (the family measures qubit 0, rotates, then takes EXP_VALs, so the evolved state after collapse is observed exactly), sampled marginals and post-collapse EXP_VAL means are compared statistically, and the BlockGlobal workspace is checked to hold one full slab per resident shot. The noisy variant adds a Pauli channel, a postselected detector, and an observable, keeps survivor rows, and checks every retained row for completeness and alignment plus pass rate and observable rate against the CPU.

Hardware run (H100 PCIe 80 GB, driver 570.195.03, CUDA 12.8 + g++ 13.3 in nvidia/cuda:12.8.1-devel-ubuntu24.04): `./build-cuda/tests/clifft_cuda_tests -d yes` → All tests passed (60812 assertions in 21 test cases); the two boundary cases take about 1.6 s each. The full command list is in docs/development/cuda-backend.md.

Not covered on hardware: the memory-derived concurrency bound. It only binds when 32 slabs per multiprocessor approach free device memory (width 20 and up on an 80 GB card), which means a unit test would have to allocate most of the device; the tier-selection case asserts the bounds the derivation must respect instead.

docs/development/cuda-backend.md
FarLab: Reworded in 95ae798: rows are reproducible for a fixed tier, precision, and block size; batch-size and concurrency-cap invariance are stated separately and tested ("CUDA cooperative concurrency cap preserves seeded rows", "CUDA retained sampler preserves seeded rows across bounded batches"). No block-size-independent reduction is claimed.

FarLab:
Review items outside the inline threads:

- **Formatting hook:** done in 95ae798 (`types_or: [c++, c, cuda]`, `sampler.cu` reformatted).
- **Kernel comments:** done in 95ae798 (no conformance transfer claimed; the lane-occupancy remark is gone).
- **Performance evidence:** bb9d441 adds a "Measured performance" section to `docs/development/cuda-backend.md` with the commands, and 64ec703 adds the `clifft_cuda_bench` tool that produced it (every CUDA row is checked against the CPU run's marginals, pass rate, and observable rates; largest |z| was 2.4). H100 PCIe, CUDA 12.8, against the 28-core CPU sampler on the same box:
  - QV-10 (width 10, 100k shots): auto tier 570k shots/s vs 348k on 28 cores (1.6x), 24k on one core.
  - Coherent QEC d5/r5 (width 13, 20k shots): 98k vs 29.5k (3.3x); FP32 171k (5.8x). Shared tier is 2x the global tier.
  - Cultivation d5 postselected (100k shots): the backend **loses**. Most shots fail an early detector, so per-shot overhead dominates; auto tier 257k vs 4.14M on 28 cores (16x slower), forced ThreadPerShot 1.26M (still 3x slower). The doc says so and points discard-heavy programs at `thread_per_shot`; a discard-aware tier policy is future work.
  - Width-four cutoff: on the synthetic width-k family ThreadPerShot beats BlockShared through width 6 (2.7x at 5, 1.6x at 6) and loses from 7. The cutoff is conservative; the doc explains why it stays for now and names the constant.
  - Concurrency default: the shared tier saturates between 16 and 32 resident blocks per multiprocessor at width 10 (1% left beyond the default) and is insensitive to the cap at width 13, where one block fills the multiprocessor.
  - Construction is 1 to 4 ms after the first sampler (which pays about 130 ms of context creation); workspace bytes are in the table.
- **New fix from the hardware run:** f44cb88 adds `__launch_bounds__(kMaxBlockSize)` to both kernels. nvcc 12.8 gave the thread-per-shot kernel 80 registers, so the documented 1024-thread block could not launch; all twelve instantiations now compile to 64 registers (the three thread-per-shot ones with a 48 to 56 byte stack frame).
- **Trailers:** `Assisted-by` is on every commit. The `Co-Authored-By` trailer on the base commit still has to go; that is a reword of the first commit, which I will do in the final rebase before merge.

Hardware: `clifft_cuda_tests` 21/21 on this branch's head; `test_experimental_cuda.py` 32/32 on the stacked #466.

src/clifft/sampling/cuda/sampler.cu
```diff
@@ -0,0 +1,1463 @@
...
+                    exp_vals.data(), log_probabilities.data(), reachable.data(), survived.data());
+                break;
+            }
+            case ExecutionTier::BlockShared: {
+                auto* kernel = detail::interpret_shots_block<Coefficient, Replay, true>;
+                check_cuda(cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize,
+                                                static_cast<int>(dynamic_shared_bytes)),
+                           "shared memory opt-in");
+                const uint32_t grid = std::min(concurrency, shots);
+                kernel<<<grid, block_size, dynamic_shared_bytes>>>(
```
bachase, line 1134: **Codex review — [P2] Avoid changing a shared kernel limit on every launch**

Separate samplers can race here. For example, sampler A sets the shared-memory limit to 192 KiB, sampler B lowers it to 96 KiB, then A launches requesting 192 KiB and fails. The setting belongs to the kernel function, while `BusyGuard` protects only one sampler instance. This affects the documented one-sampler-per-concurrent-caller usage, including Python calls that release the GIL.

Configure a consistent maximum once per device and kernel specialization, with thread-safe initialization, and add a concurrent test using different program widths. This is a static finding supported by [CUDA's documented function-wide limit](https://docs.nvidia.com/cuda/archive/12.8.0/cuda-runtime-api/structcudaFuncAttributes.html), not a hardware reproduction.

docs/development/cuda-backend.md
```diff
@@ -0,0 +1,341 @@
...
+| Exact-probability and state-vector queries | Not supported |
+| Asynchronous or multi-GPU execution | Not supported |
+| Python facade | Not yet; C++ only |
+
+Unsupported programs are rejected during lowering; there is no automatic CPU
+fallback.
+
+## Execution tiers
+
+Where the HIP backend assigns every shot to one thread, the CUDA backend
```
bachase: HIP backend wil change over time. Aside from briefly mentioning the same contract above, this documentation should solely focus on the HIP implementation and not compare to the CUDA one.

docs/development/cuda-backend.md
```diff
@@ -0,0 +1,341 @@
...
+selects one of three tiers per program and device:
+
+| Tier | Shot ownership | Coefficient residence | Automatic selection |
```
bachase: These column names are a bit weird to me. Perhaps better ones are simply "Tier | Shot | Coefficients | Selection"

docs/development/cuda-backend.md
```diff
@@ -0,0 +1,341 @@
...
+  resident per launch, which bounds `BlockGlobal` slab memory. Zero derives the
+  cap from the multiprocessor count and free device memory.
+- `block_size` must be a power of two between 1 and 1024. In the cooperative
+  tiers it is the number of lanes that share one shot.
+- `allocated_device_bytes()` exposes the retained workspace size.
+
+CPU, HIP, and CUDA use separate random-stream domains, so compare
+deterministic branches directly and stochastic results statistically.
+
+## Measured performance
```
bachase: This is useful information, but I don't think it belongs in our docs. The docs should largely be information about developing the CUDA backend; so its architecture and higher-level design, and any scope concerns. 

PErformance information is helpful, but this reads too much like a one-off study. Perhaps leave this in the github comments of the PR? Or alternatively, you can propose another place to store them if its relevant to capture.

benchmarks/cuda_sampling_bench.cc
```diff
@@ -0,0 +1,443 @@
+// CPU-versus-CUDA evidence for the experimental CUDA sampling backend.
+//
+// Each workload is compiled through the production pipeline (parse -> trace
+// -> default HIR passes -> plan_sampling; --postselect marks every detector
+// as postselected so survivor sampling runs). The CPU sampler is timed at the
+// requested thread budgets, then every CUDA execution tier is timed twice:
+// construction (program upload plus workspace allocation) and retained
+// sampling of the same shot count after one warm-up call.
+//
+// Correctness is checked, not assumed: per-record marginals (and, for
```
bachase: Checking for correctness here is nice, but it also feels like its mixing purposes. I'm open to alternatives, but shouldn't we have correctness a separately tested and validated setup, and focus this on pure benchmarking? If so, would this be able to use the google benchmarking approach?

I'm open to pushback on this, but just raising

src/clifft/sampling/cuda/sampler.cu
bachase: This one remains open too

benchmarks/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+Before each sampling measurement, a small untimed warmup checks output shapes
+and survivor-count consistency. Statistical conformance belongs in
+`tests/test_cuda_sampler.cc`; benchmark timings are not correctness evidence.
+Use Google Benchmark's repetition, filtering, and JSON output controls for
+performance comparisons.
+
+## Historical measurements
+
+The [September 2026 H100 report](reports/cuda-h100-2026-09.md) preserves the
```
bachase: Is this file checked in? I don't think we need to, so I'd drop this section.

docs/development/cuda-backend.md
```diff
@@ -148,148 +152,49 @@ const auto rows = sampler.sample(100000, uint64_t{42}, /*block_size=*/256);
...
-| 7296 | 574 k | 98.1 k |
-
-At width 10 a shot uses 40 KB of shared memory, so several blocks share a
-multiprocessor and throughput keeps improving until roughly 16 to 32
-resident blocks per multiprocessor, with 1% left beyond the default. At
-width 13 a shot uses 212 KB, one block fills the multiprocessor, and the
-cap is irrelevant above 114. The default therefore saturates the shared
-tier, and for the global tier it bounds slab memory at 32 slabs per
-multiprocessor unless free memory bounds it first.
+The [historical H100 report](https://github.com/unitaryfoundation/clifft/blob/main/benchmarks/reports/cuda-h100-2026-09.md)
```
bachase: As per my other comment, drop this report and its link from teh repo. You can still refer to benchmark instructions

PR #466, opened 2026-09-02
feat(cuda): add experimental Python facade for the CUDA backend

FarLab:
Closes #493.

## Summary

PR 2 of 2, stacked on #464. Exposes the CUDA sampling backend to Python the same way #395 exposed HIP:

- `_clifft_cuda` nanobind module (`src/python/cuda_bindings.cc`), built only under `CLIFFT_ENABLE_CUDA=ON` and kept out of ordinary wheels, with its own version script.
- `clifft.experimental.cuda` facade: `is_built`, `is_available`, `backend_info`, `compile`, `lower`, `Program`, `Sampler`, `ReplayResult`, mirroring `clifft.experimental.hip`. CUDA-specific additions: `selected_tier(program, precision)`, `Sampler(tier=..., max_concurrent_shots=...)`, and `Sampler.tier` reporting the resolved execution tier.
- `clifft.experimental` lazily loads both backends; `import clifft` still loads neither.
- `tests/python/test_experimental_cuda.py` + `utils_cuda.py`: GPU-free boundary tests plus hardware tests for tier selection, forced replay in every tier, cross-tier agreement on a width-6 program, the concurrency cap, and joint distributions against the CPU.
- The `cuda-offline-build` CI job now also builds the editable package with nvcc and runs the GPU-free Python tests.
- Docs: Python workflow in `docs/development/cuda-backend.md`, API entries in `docs/reference/python-api.md`.

## Note for reviewers

`tests/python/utils_hip.py::assert_forced_record_probabilities` and the HIP replay test use a circuit with a hidden record (`R 0`) plus detectors and observables, which the CPU `record_probabilities` oracle now rejects. That test only runs on AMD hardware, so it has rotted silently; the CUDA copy uses a measurement-only circuit and guards against hidden records. The HIP test deserves the same fix in a separate PR.

## Test plan

- [x] `pre-commit run --all-files` clean
- [x] CPU-only editable install: `test_experimental_cuda.py`, `test_experimental_hip.py`, `test_sampling_integration.py` pass (GPU cases skip)
- [x] `cuda_bindings.cc` compiles against nanobind headers with `-Wall -Wextra`
- [x] `mkdocs build --strict`
- [x] nvcc 13.0.88-built extension on an H100 SXM (Nebius via Brev): `test_experimental_cuda.py` 16/16 pass, including every tier on forced replay and the width-6 program
- [x] `cuda-offline-build` CI job green (builds the editable package with nvcc and runs the GPU-free Python tests)
- [x] Review follow-up on an H100 PCIe (CUDA 12.8): `test_experimental_cuda.py` 32/32, including survivor sampling in every tier and complete-row distributions in the cooperative tiers


tests/python/test_experimental_cuda.py
```diff
@@ -0,0 +1,180 @@
...
+    assert "CLIFFT_ENABLE_CUDA=ON" in cuda.backend_info()
+    with pytest.raises(RuntimeError, match="CLIFFT_ENABLE_CUDA"):
+        cuda.compile("M 0")
+
+
+def test_cuda_facade_rejects_unknown_tier_names() -> None:
+    if not cuda.is_built():
+        pytest.skip("requires the CUDA extension")
+    program = cuda.compile("H 0\nT 0\nM 0")
+    require_cuda_device()
```
bachase, line 63: **Codex review — Run invalid-tier validation without a GPU**

Please remove this `require_cuda_device()` call. Once the extension is built, `cuda.compile()` can lower without a device, and `_tier_value()` raises for the invalid name before the native sampler is constructed. The device guard unnecessarily skips this boundary test in CUDA CI. Invalid-precision validation can be covered the same way.

tests/python/test_experimental_cuda.py
```diff
@@ -0,0 +1,180 @@
...
+H 0
+CX 0 1
+PAULI_CHANNEL_1(0.1, 0.2, 0.05) 1
+M 0 1
+DETECTOR rec[-1] rec[-2]
+OBSERVABLE_INCLUDE(0) rec[-1]
+"""
+    shots = 20_000
+    cpu = clifft.sample(clifft.compile(circuit), shots, seed=41)
+    gpu = cuda.Sampler(cuda.compile(circuit), precision=precision).sample(shots, seed=42)
```
bachase, line 176: **Codex review — Check noisy joint distributions in the cooperative tiers**

This circuit selects `thread_per_shot`, so the joint-distribution check never exercises cooperative noise execution. Please parameterize the test over the explicit tiers as well as precision. The width-six tests compare measurement marginals, which cannot detect incorrect correlations; retain a complete-row distribution check for the cooperative paths too.

bachase (commented):
**Codex review**

Please add a Python test for `Sampler.sample_survivors()` with detector postselection and both values of `keep_records`. Check aggregate counts and the returned measurement/detector/observable/expectation array shapes and row alignment. The new suite does not call this method, so the C++ tests do not cover its conversion into Python `SampleResult` objects.

The backend synchronization, CI gate, formatting-hook, and execution-range feedback is posted on #464; please carry those fixes through this stack.

tests/python/test_experimental_cuda.py
FarLab: Done in d12bb20: the device guard is gone, and the case now also checks that an invalid precision is rejected by both `Sampler` and `selected_tier`. All three rejections happen in the facade before a native sampler or device query exists, so the case runs in cuda-offline-build.

tests/python/test_experimental_cuda.py
FarLab: Done in d12bb20: the joint-distribution test is parameterized over the three explicit tiers, both precisions, and two circuits (the original narrow one, and a width-6 noisy circuit with a detector and an observable, 8 output bits), keeping the complete-row comparison for every combination. The new survivor test also compares retained survivor rows as complete rows per tier. 32/32 on an H100 PCIe.

FarLab:
- **`Sampler.sample_survivors()` coverage:** d12bb20 adds `test_cuda_python_survivor_sampling_counts_and_rows`, parameterized over the three explicit tiers and both `keep_records` values, on a postselected width-6 circuit. It checks `total_shots`, `passed_shots`, `discards`, `logical_errors`, `observable_ones`, the array shapes in both modes, and row alignment of retained rows against themselves: the postselected column is zero, the observable equals the parity of the two records it reads, and a trailing `EXP_VAL Z0` equals `1 - 2*m0` row by row, which a compaction or download that shifted columns would break. Pass rate and observable rate are compared with the CPU under a two-sample tolerance, and retained rows as complete-row distributions.
- Rebased onto the current head of #464 (which carries the backend fixes, the boundary tests, the bench, and the `__launch_bounds__` fix found on this hardware run).
- Hardware: `uv run pytest tests/python/test_experimental_cuda.py -v` → 32 passed on an H100 PCIe (CUDA 12.8, driver 570.195.03).

bachase (commented):
Will re-review after the base parent PR goes through, but looks good so far.

PR #468, opened 2026-09-02
feat(optimizer): tune scheduling-pass ranking, add benchmarks, enable by default

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`. Stacked on #463; the diff includes #459 through #463 until they merge.

## Ranking-key experiment (kept: K1)

`ActiveWidthSchedulePass`'s beam search ranks tied candidates (same peak, same width-after-closure) by a third key before falling back to `first_op`. The current key ties on `dense_work_so_far`. I implemented four candidate keys as an internal, environment-variable-selected enumeration in `active_width_schedule_pass.cc` (not a public option), measured each on `coherent_d3_r3.stim`, `coherent_d5_r5.stim`, `cultivation_d5.stim` (from `tests/fixtures`) and `distillation.stim`, `coherent_d5_r1.stim` (from `clifft-paper/qec_bench/circuits`), all after `PeepholeFusionPass` then `StatevectorSqueezePass`, default pass options (`beam_width=8`, `exact_node_budget=0`), Release build, then deleted the enumeration and hard-coded the winner:

| Circuit | Incumbent peak / work | K0 current (dense_work) | K1 prototype (−swept_count) | K2 (−swept_count, dense_work) | K3 (cost-to-go) |
|---|---:|---:|---:|---:|---:|
| coherent_d3_r3 | 5 / 1390 | 4 / 442 | 4 / 420 | 4 / 420 | 5 / 638 |
| coherent_d5_r5 | 13 / 2,186,274 | 13 / 1,347,594 | 13 / 870,182 | 13 / 870,180 | 13 / 2,186,274 (not applied) |
| cultivation_d5 | 10 / 41,710 | 10 / 33,174 | 10 / 24,486 | 10 / 26,550 | 10 / 33,174 |
| distillation | 5 / 186 | 3 / 86 | 3 / 86 | 3 / 86 | 4 / 106 |
| coherent_d5_r1 | 12 / 39,970 | 0 / 0 | 0 / 0 | 0 / 0 | 0 / 0 |
| **Sum** | 47 / 2,269,530 | 30 / 1,381,296 | **30 / 895,174** | 30 / 897,236 | 32 / 2,223,292 |

K3 (a myopic `dense_work_so_far + 2^width_after_closure` cost-to-go estimate) loses on summed peak alone (32 vs 30) and is eliminated. K0, K1, and K2 tie on summed peak (30); K1 has the lowest summed dense work, beating K2 by about 0.2% and K0 by 35%. K1 is also the simplest of the three surviving keys (four ordered components vs. K2's five), so no tie-break judgment call was needed. K1 replaces K0 as the hard-coded key: `(peak, width_after_closure, −swept_count, first_op)`, where `swept_count` is the number of operations the candidate's closure executes, including the first op. On `coherent_d5_r5` this is a 1.55x reduction in dense work versus the previous key, matching the "another 1.5x" the issue's own scoring-gap note predicted.

Kept unchanged: determinism (same fixture and fuzz tests pass unmodified), the never-worse guard, and sampling-equivalence tests. Two test-only numbers that were downstream of the old key's dense-work output changed and are now updated with their measured values: `coherent_d5_r5`'s `beam_width=1` vs `beam_width=8` dense-work comparison (now 38.8% vs 39.8% of incumbent, both comfortably under half, versus the old key's ~62%) and a source comment describing the ranking key's rationale.

## Benchmarks (Part 2) and default-pipeline effect (Part 3)

Added `schedule_pass_coherent_d5_r5` and `schedule_pass_cultivation_d5` to `benchmarks/clifft_benchmarks.cc`, timing `ActiveWidthSchedulePass::run` alone (default options) on an HIR already taken through `PeepholeFusionPass` and `StatevectorSqueezePass`; each iteration copies the prepared HIR fresh (`benchmark::DoNotOptimize`) so it always times a first-time schedule. Both names are registered in `tools/ci/benchmark_canary.py`'s `DISPLAY_NAMES`.

`ActiveWidthSchedulePass` is now `default_enabled = true` in `pass_registry.h`, after `StatevectorSqueezePass`. Full benchmark binary (`-DCLIFFT_BUILD_BENCHMARKS=ON`, Release, `--benchmark_min_time=0.5s --benchmark_repetitions=3`), medians, run once before and once after that flip, same host, same session:

| Benchmark | Before | After | Change |
|---|---:|---:|---:|
| `squeeze_parallel_t_8192` | 0.029 ms | 0.029 ms | unaffected (sanity check) |
| `compile_plan_cultivation_d5` (compile only) | 13.00 ms | 110.55 ms | 8.5x slower to compile |
| `schedule_pass_coherent_d5_r5` (new) | 735.60 ms | 739.04 ms | unaffected (doesn't use the default manager, sanity check) |
| `schedule_pass_cultivation_d5` (new) | 98.36 ms | 98.31 ms | unaffected (sanity check) |
| `sample_qv10_100_shots` | 4.28 ms | 3.89 ms | 1.10x faster |
| `sample_cultivation_d5_1000_shots` | 15.34 ms | 11.57 ms | 1.33x faster |
| `sample_coherent_d5_r5_100_shots` | 48.17 ms | 20.50 ms | 2.35x faster |
| `sample_surface_d7_r7_10000_shots` | 26.93 ms | 26.86 ms | unaffected (Clifford-only, width 0) |
| `sample_surface_d5_r5_high_noise_10000_shots` | 29.89 ms | 29.99 ms | unaffected (Clifford-only, width 0) |
| `sample_surface_d11_r11_1000_shots` | 10.97 ms | 11.51 ms | unaffected within noise (Clifford-only, width 0) |
| `sample_exp_val_20q_200_probes_100000_shots` | 151.19 ms | 156.52 ms | unaffected within noise (Clifford-only, width 0) |

These are single-host, single-session figures (3 repetitions per benchmark, but one host and one run of the whole binary each side); treat the ratios as directional, not as a performance guarantee.

As a cross-check against the issue's own corpus table (measured under the old key, K0), I also re-measured `coherent_d3_r3` and `coherent_d5_r1` from the `clifft-paper` corpus directly with `tools/profile/profile_sample` (100,000 shots, same host, this PR's final code): `coherent_d3_r3` improves 1.87x (peak 5 -> 4; issue's K0 measurement: 2.0x) and `coherent_d5_r1` improves 47.9x (peak 12 -> 0; issue's K0 measurement: 45x) -- consistent with, and for `coherent_d5_r5` measurably better than, the issue's original estimate, since K1 finds more dense-work reduction than K0 did.

**The tradeoff, plainly:** compile time on `coherent_d5_r5` goes from about 5 ms to about 0.6-0.9 s across the measurements above, and on `cultivation_d5` from about 12-14 ms to about 0.1 s (`tools/profile/profile_compile` total-stage medians on this host: `coherent_d5_r5` 4.75 ms -> 640.5 ms, `cultivation_d5` 13.6 ms -> 102.5 ms). In exchange, sampling throughput improves roughly 1.9x-2.4x on the coherent circuits measured and about 48x on `coherent_d5_r1`, with no measurable change on the Clifford-only fixtures the pass cannot improve. This commit (`feat(optimizer)!: enable ActiveWidthSchedulePass by default`) is kept separate from the ranking and benchmark commits specifically so it can be dropped if the maintainer prefers to keep the pass opt-in; nothing in the other three commits depends on it.

## Docs (Part 4)

New page `docs/theory/active-width.md`: the structural width model (the dormant subspace, its two update rules, width as its codimension), why pivot choice doesn't affect the width trace, confluence and the order-invariant final width, the closure theorem in plain words, noise transparency's soundness argument, `search_width_schedule`'s certificate scope and its explicit non-claims, what `ActiveWidthSchedulePass` does with the never-worse guarantee, and a measured-effect section reproducing the issue's own corpus table, labeled single-host at commit `2d99ec5c`. Linked from `docs/theory/overview.md`'s HIR Optimization subsection and from a new "Compile-time scheduling" subsection in `docs/guide/cpu-execution.md` (the current name of the page formerly at `guide/performance.md`; see judgment calls below). Added to `mkdocs.yml`'s theory nav. `docs/reference/passes.md` is generated from the registry and `docs/macros.py`; confirmed the pass renders with its summary, detail, and an "Enabled" default marker. Also filled a pre-existing gap: `docs/reference/python-api.md` never listed `clifft.ActiveWidthSchedulePass` in its HIR Passes section even after it was exposed to Python, so every other pass had a reference entry and this one didn't; added it.

## Test commands and results

```
cmake -B build-debug -DCMAKE_BUILD_TYPE=Debug -G Ninja && cmake --build build-debug -j8
ctest --test-dir build-debug -j8 --output-on-failure
# 100% tests passed, 0 tests failed out of 974; ~104s wall time (parallel)

cmake -B build-release -DCMAKE_BUILD_TYPE=Release -DCLIFFT_BUILD_BENCHMARKS=ON -G Ninja
cmake --build build-release --target clifft_benchmarks -j8
./build-release/benchmarks/clifft_benchmarks --benchmark_min_time=0.5s --benchmark_repetitions=3
# see the before/after table above

uv venv .venv
# 1394 passed, 7 skipped in 199.81s

uv run --group docs --python .venv/bin/python mkdocs build --strict
# Documentation built in 1.80 seconds, exit code 0

uv run --python .venv/bin/python pytest --codeblocks docs/ README.md -v
# 39 passed, 5 skipped in 3.43s

```

## Judgment calls

- `docs/guide/performance.md` no longer exists; `mkdocs.yml` redirects it to `guide/cpu-execution.md#benchmark-before-overriding`, so I added the "Compile-time scheduling" subsection there instead, as a new top-level section (the rest of the page is scoped to per-shot thread/batch tuning, not compile-time cost, so I introduced it as a distinct concern rather than nesting it under the existing power-user tuning subsections).
- Two test helpers (`coherent_d3_r3_production()` in `tests/test_active_width_search.cc`, `_production_hir()` in `tests/python/test_active_width_search.py`) built their "production" incumbent via `default_hir_pass_manager()` specifically to feed `search_width_schedule`/`clifft.search_width_schedule` as a peephole+squeeze baseline. Once the default manager also schedules, that baseline was already at the exact search's own optimum, collapsing the tests to a no-op. Rather than delete the tests, I gave each an explicit peephole+squeeze-only construction (a second helper on the Python side, since one differential test in the same file legitimately wants the true default pipeline and still uses the original helper).
- `tests/python/test_benchmark_canary.py`'s fork-report fixture already derived its workload count from `DISPLAY_NAMES`; only its assertion text hardcoded "9 of 9". I made that count dynamic (`len(benchmark_canary.DISPLAY_NAMES)`) rather than hand-updating it to "11 of 11", so a future benchmark addition doesn't retrip the same test.
- `build-debug/` and `build-release/` -- the exact directory names this task's own build instructions specify -- aren't covered by `.gitignore` (which only lists `/build/`, `build-coverage/`, `build-profile/`, etc.). I left them untracked and out of every commit rather than fix `.gitignore`, since it's unrelated to this PR's scope, but it's a real gap worth a separate fix; `#463`'s own PR body used the same directory names.

Assisted-by: Claude (Sonnet 5) <noreply@anthropic.com>


bachase:
Reviewed and verified independently at `5c3a67de`.

- Local Debug `ctest` 974/974 and the full Python suite (1394 passed) at `d169751b`; the final commit touches only the two docs files, and the strict docs build is clean at `5c3a67de`.
- Ranking key: keeping the closure-sweep tie-break is the right call; on my Release harness against the `clifft-paper` corpus it takes coherent d5 r5 from 2099 to 4938 shots per second (plan work 1.99e6 to 7.8e5) at 600 ms of pass time, coherent d3 r3 from 1.08M to 2.03M, distillation from 1.12M to 1.18M, and cultivation d5 from 58k to 60k, with coherent d5 r1 unchanged at width 0. The greedy beam now also finds the d3 r3 peak-4 schedule, so the old beam-width sensitivity is gone.
- The theory page's corrected confluence identity, exchange-argument closure proof, noise-transparency mechanism, beam description, and the cultivation wording that distinguishes d3's certificate from d5's exhausted budget all match the code and the measurements.
- The default flip stays a separable commit (`af1b28ca`) with the compile-time tradeoff stated; that decision is the maintainer's.

No further changes requested. This completes the six-PR stack for #458.

bachase:
Superseded by the re-cut review series: the closure-sweep ranking and the pass benchmarks live on in #471 (PR B), the corrected theory page and guide text in #472 (PR C), and the default-on flip is deferred to a separate PR once CI time with the pass in the default pipeline is measured. Closing without merge.

PR #471, opened 2026-09-02
feat(optimizer): structural active-width analysis and scheduling pass

bachase:
Prototype work on the `feature/active-width-scheduling` branch toward #458; review can be less strict than for `main`. Stacked on #460 (planner logical-position noise signs). Consolidates and simplifies #459, #461, #463, and the tuning from #468; the exact search from #462 moves out of the product.

## Objective and algorithm

`DormantSubspace`/`analyze_active_width` recompute the sampling planner's active-width trace directly from HIR-frame Pauli masks with GF(2) linear algebra, without building a coordinate frame. `ScheduleDependence` is a conservative dependence DAG over HIR op indices: only `T_GATE`, `PHASE_ROTATION`, and `MEASURE` ever move, built from `can_swap` plus an optional relaxation that lets a reorder cross a `NOISE` op (sound because the planner resolves that op's noise-dependent sign from its logical position, carried through `HirModule::logical_noise_prefix`, from PR A/#460). The closure theorem these rest on: some peak-minimizing schedule executes every ready non-expanding op as soon as it is ready, so a scheduler only has to choose which ready *expanding* rotation fires next; the theorem now lives as a short invariants comment (ready, expanding, closure, confluence) at the top of `active_width_schedule_pass.cc`'s anonymous namespace, pointing to a forthcoming `docs/theory/active-width.md` for the full argument (see "Closure machinery folded into the pass" below).

`ActiveWidthSchedulePass` is a beam search over that closure/readiness machinery: at each step it scores every ready expanding op's own closure sweep (peak, then width-after-closure, then sweep size, then op index) across all beam members, keeps the `beam_width` survivors, and reports whichever completed beam member scores best on `(peak, dense_work)`. It sinks width-neutral rotations next to the following expansion for executable-plan fusion, and only ever replaces the HIR when its candidate is strictly better than the incumbent by `analyze_active_width`/`estimate_dense_work` -- otherwise the HIR is left byte-for-byte untouched. It is registered `default_enabled = false` (compile time is not yet bounded to a small multiple of the rest of the pipeline) and record-order breaking, positioned after `StatevectorSqueezePass`.

## Leaner frontier

`SearchFrontier`'s ready set moved from `std::set<uint32_t>` (returning a fresh `std::vector<uint32_t>` by value from every `execute()` call) to a flat `ready_flag_` array plus a `lowest_ready_hint_` -- a safe lower bound on the smallest ready index, advanced only when its own witness is removed. `find_ready_non_expanding` scans ascending from the hint, skipping not-ready indices with a cheap flag check, so `is_expanding()` (a `PauliString` allocation plus a GF(2) commutation scan) only ever runs on an actual ready op, in the same lowest-index-first order and with the same early exit a sorted container gave for free. `execute()`/`undo()` now append to and pop from a caller-provided `newly_ready_log` instead of allocating their own vector per call.

Verification: I extracted the pre-rewrite `active_width_closure`/`active_width_schedule_pass`/`active_width_search` sources from this branch's merge-base commit into a side-by-side standalone build (same headers otherwise) and compared `result_peak()`/`result_dense_work()` before and after on `coherent_d3_r3`, `coherent_d5_r5`, `cultivation_d5`, `distillation` (from `~/clifft-paper/qec_bench/circuits/`, not a repo fixture), and `surface_d7_r7_p001`, at both `beam_width=1` and the default `beam_width=8`. A first version of the rewrite (an unordered ready list with a running-minimum scan, no early exit) reproduced identical numbers but regressed Debug wall time (~32 s vs ~21 s at beam 8 on `coherent_d5_r5`) and, separately, changed `coherent_d3_r3` and `cultivation_d5`'s dense work by a small amount at beam 8: `score_candidates`' ready-op iteration order feeds the beam ranking sort, whose key does not fully order every candidate pair, so a different input order could lead an otherwise-tied sort to a different, equally-ranked winner. Making `ready_ops_snapshot` scan ascending (the final design above) fixed both: performance recovered to parity and every fixture is bit-identical to the pre-rewrite implementation.

`coherent_d5_r5` pass wall time before/after, on top of an explicit Peephole+Squeeze pipeline (matches the fixture used in `schedule_pass_coherent_d5_r5`):

- Debug, `beam_width=8`: ~21.0 s before -> ~21.3 s after (parity within noise).
- Release, `beam_width=8`, same compiler flags as the real Release build (`-O3 -march=native -mtune=native -ffast-math`): ~644-661 ms before (standalone harness) -> 634 ms after (official `schedule_pass_coherent_d5_r5` benchmark, median of 3 reps, CV 0.31%).

`is_expanding()`'s own cost (unchanged by this PR) dominates this fixture's wall time, so the frontier rewrite's main benefit here is fewer heap allocations rather than a large wall-clock win; `SearchFrontier` is otherwise unaffected by anything downstream.

## Test hygiene

Every test in this PR that needs a "production" HIR builds `PeepholeFusionPass` then `StatevectorSqueezePass` explicitly through `HirPassManager`, never `default_hir_pass_manager()`. `test_active_width_analysis.cc` previously called `default_hir_pass_manager()` inside its fixture-differential loop and its 400-trial random-circuit test; fixed here so this suite's runtime and behavior track only the passes it actually exercises.

## Review feedback applied

A second pass over this PR, in commit order:

**Beam dedup respects the objective.** `run_beam_search`'s dedup step kept only the lower-`dense_work` candidate among duplicates that converged on the same executed-op bitset, ignoring that the peak reached along the way is path-dependent even though the final subspace and width are confluent. That could discard a strictly better duplicate: lower peak but higher dense work. Dedup now ranks by `(peak, dense_work, first_op)` ascending. Found via a seeded random search over small circuits run against an instrumented copy of the dedup step (not committed) that flagged every peak/dense-work inversion; the frozen regression circuit (8 ops, 4 qubits, in `test_active_width_schedule_pass.cc`) reproduces it exactly: the old rule finds no improvement at all (result peak stuck at the incumbent's 4), the fix reaches the brute-force optimum (peak 3), computed by exhaustively enumerating the dependence relation's linear extensions.

Fixture numbers, before -> after (default options unless noted; `coherent_d5_r5` at `beam_width=8` to match this table's convention, though the in-tree test still narrows it to `beam_width=1` for Debug wall time):

| Fixture | Before (beam 8) | After (beam 8) | Ratio (after) |
|---|---:|---:|---:|
| coherent_d3_r3 | 4 / 420 | 4 / 420 | 0.302 (unchanged) |
| coherent_d5_r5 | 13 / 870,182 | 13 / 870,180 | 0.398 (unchanged to 3 places) |
| cultivation_d5 | 10 / 24,486 | **10 / 23,542** | 0.565 (was 0.587) |
| distillation (clifft-paper corpus) | 3 / 86 | 3 / 86 | 0.462 (unchanged) |
| surface_d7_r7_p001 | 0 / 0 (early exit) | 0 / 0 (early exit) | n/a |

Only `cultivation_d5` moves, and only downward (dense work 24,486 -> 23,542, about 3.9% lower); every other fixture is unchanged or within floating-point noise. Confirmed by temporarily reverting to the old rule and rerunning the exact same C++ fixture test.

**Beam dedup keeps the Pareto front, not a single lexicographic winner.** The `(peak, dense_work, first_op)` rule above still threw away information within one executed-op set. Peak-so-far and dense-work-so-far are both path properties -- peak is a running max over a path-dependent sequence of transitions, and dense_work is the pass's own secondary objective -- so neither dominates the other until the rest of the schedule is known. A candidate reached with lower peak but higher dense work can lose to one reached with higher peak but lower dense work if the shared suffix later forces both up to that same higher peak anyway: at that point only the lower-dense_work duplicate was worth keeping, and the lexicographic rule cannot see that coming since it decides before the suffix ever runs. Dedup now keeps every candidate in a set that no other candidate of the same set dominates (peak no worse and dense work no worse, at least one strictly better), collapsing candidates tied on both to one canonical survivor by comparing the complete prospective order -- the parent's already-executed order followed by this step's ops -- lexicographically, so the choice does not depend on the order candidates were generated in (`first_op` alone is not a total tie-break across candidates from different parents, who can share it). The front stays tiny in practice (one candidate per distinct peak), so `append_pareto_front` is a plain O(k^2) scan over each group, not a new data structure; the beam_width cut afterward is unchanged.

Found by extending the existing 8-op regression circuit's random-search seed with two more hand-picked ops, iterated until the final answer actually changed, then confirmed against an instrumented copy of the dedup step (not committed) that showed the mechanism directly. One executed-op set is reached by two candidates, `(peak 3, dense_work 54)` and `(peak 4, dense_work 52)` -- neither dominates the other. The old rule keeps only the lower-peak `(3, 54)` duplicate and discards `(4, 52)` outright. Two steps later the shared suffix forces every surviving path to peak 4 anyway: an executed set further on is reached by three candidates, `(4, 70)` and `(4, 94)` descending from the kept `(3, 54)`, and `(4, 68)` descending from the discarded `(4, 52)` -- strictly the best of the three, and gone for good under the old rule. So the old rule's beam search reports `(4, 70)` as final; the new rule reaches the brute-force optimum, `(4, 68)`. A new 12-op regression circuit in `test_active_width_schedule_pass.cc` asserts the pass's `(peak, dense_work)` matches the brute-force optimum exactly (the 8-op regression above only certifies the peak).

Re-ran the fixture table above with this change, same sources and default options: identical on all five circuits (`coherent_d3_r3`, `coherent_d5_r5`, `cultivation_d5`, `distillation`, `surface_d7_r7_p001`), confirmed by temporarily reverting to the `(peak, dense_work, first_op)` rule and rerunning the same fixture dump. The Pareto generalization is a correctness fix for a case this particular fixture corpus does not happen to exercise, not a further improvement on it.

**Schedule dependence is bound to its source HIR, exactly.** `apply_schedule()` accepted any `ScheduleDependence` with a matching operation count, so a relation built from one HIR's commutation structure could be silently replayed against a structurally different HIR that happens to share an op count. The first fix for this recorded a fingerprint over each op's inline fields plus its payload *indices* -- `detector_idx`, `observable_target_list_idx`, `readout_noise_idx` -- rather than the side-table content those indices point to. That left a hole: two HIRs with the same op-type sequence but different `detector_targets`/`observable_targets`/`readout_noise[...].meas_idx` content can carry identical indices (each module's own single detector is index 0, say), so the fingerprints collided even though `can_swap` -- which dereferences those indices into the side tables -- would disagree between the two. Confirmed reproduction: source `M 0; M 1; DETECTOR rec[-2]`, target `M 0; M 1; DETECTOR rec[-1]`, order `{0, 2, 1}`; the relation built from the source accepted the order, `apply_schedule` accepted the target, and `plan_sampling` on the reordered target threw "operation 1 reads record 1 before assignment".

Fingerprinting now lives next to `can_swap` itself, in `commutation.h`/`.cc`, as `CommutationFingerprint`/`commutation_fingerprint`, with a header comment stating the contract in one sentence: it covers every `HirModule` field `can_swap` reads, so a change to `can_swap`'s inputs is a change to this function in the same file. It hashes the *dereferenced* side-table content (detector/observable target list length and elements; the readout entry's `meas_idx`; every noise channel's mask words, regardless of probability, since `can_swap` checks all of them) rather than the raw index, and drops `flags()`/dagger, `alpha()`, and `logical_noise_prefix` entirely, since `can_swap` never reads any of them -- a T gate's dagger direction, a rotation's angle, or a materialized `logical_noise_prefix` cannot change which reorderings are legal, so the old fingerprint's sensitivity to them caught nothing `can_swap` cared about. `ScheduleDependence` stores a `CommutationFingerprint`; `apply_schedule` compares against it and throws `std::invalid_argument` on mismatch. New tests in `test_schedule_dependence.cc`: the reproduction above (now throws, target left byte-for-byte unchanged), a mutation test over one circuit exercising every op type `can_swap` follows through a side table or inline mask (flip one input per category, requiring `apply_schedule` to throw and `commutation_fingerprint` to differ, while the unmutated copy is accepted and its fingerprint matches), and a check that materializing `logical_noise_prefix` does not change the fingerprint. `ScheduleDependence`, `ScheduleDependenceOptions`, and `apply_schedule` remain in `namespace clifft::detail`: pass internals, not part of the Python-facing API.

**Structural early exit.** The early-exit test inferred the skip by timing `run()` against a standalone `ScheduleDependence::build` call and asserting the pass was faster -- a real but indirect signal. Added `ActiveWidthSchedulePass::built_dependence()`, a read-only statistic set exactly when `run()` builds the relation; the test now asserts `built_dependence() == false` directly, plus a new positive-case test on `coherent_d3_r3` asserting `built_dependence() == true`. `<chrono>` dropped from the test file.

**Narrower exception catch.** `test_active_width_analysis.cc`'s `require_trace_matches_plan_if_plannable` caught `std::exception` to skip HIR that legitimately overflows the planner's dense active-width limit. Narrowed to `std::overflow_error`, the exact type the limit throws; re-ran the full `active_width` suite (fixture differential, 400-trial random circuits, instrument-mode differential) and no fixture surfaced any other exception type.

**Zero beam width is rejected.** `beam_width == 0` silently fell back to the identity order via a dead branch in `run_beam_search`. The `ActiveWidthSchedulePass` constructor now throws `std::invalid_argument`, the fallback is gone (replaced by an `assert` documenting the invariant), and the Python binding validates the same way so construction raises `ValueError`. New C++ and Python tests.

**Closure machinery folded into the pass.** `active_width_closure.h/.cc` existed to share readiness/closure bookkeeping with a research tool outside this repo; that tool never landed here and nothing else in-tree used the header. `is_expanding`, `UndoStep`, `SearchFrontier`, `undo_all`, `find_ready_non_expanding`, and `run_closure` move into the anonymous namespace of `active_width_schedule_pass.cc`; the unused `apply_op` wrapper is dropped; both files (and their CMake registrations in `src/clifft` and `src/wasm`) are deleted. `classify_and_apply`, `is_expanding_effect`, `pauli_body`, and `dense_work_contribution` stay in `active_width_analysis.h`, which the pass already drew them from. The long closure-theorem sketch collapses to the short invariants comment mentioned above.

**End-to-end compile benchmarks.** Two additions to `clifft_benchmarks`: `compile_schedule_plan_coherent_d5_r5` runs `PeepholeFusionPass`+`StatevectorSqueezePass`+`ActiveWidthSchedulePass`+`plan_sampling` fresh each iteration, protecting the full compile path a production pipeline with scheduling enabled would pay (not just the pass's own `run()` cost `schedule_pass_coherent_d5_r5` isolates). `schedule_dependence_build_surface_d7` deliberately bypasses the pass's early exit (`surface_d7_r7` is pure Clifford, so the pass itself never builds the relation on it) to time `detail::ScheduleDependence::build`'s O(N^2) `can_swap` scan alone, on the corpus's largest movable-op count. Both registered in `tools/ci/benchmark_canary.py`'s `DISPLAY_NAMES`. Release run (single run, this host, `-O3 -march=native -mtune=native -ffast-math`):

- `compile_schedule_plan_coherent_d5_r5`: 523 ms (1 iteration).
- `schedule_dependence_build_surface_d7`: 50.7 ms (10 iterations).

Scaling note, measured ad hoc with the same `detail::ScheduleDependence::build` call outside the benchmark harness: on `surface_d7_r7_p001.stim` post-Peephole+Squeeze (3925 ops, 482 movable) the build takes ~50.7 ms; on `tests/fixtures/surface_d11_r11_p001.stim` post-Peephole+Squeeze (15057 ops, 1682 movable) it takes ~1062 ms (best of 5 reps). Op count grows ~3.8x and movable count ~3.5x between the two fixtures, but build time grows ~21x -- consistent with (worse than either dimension alone, better than the documented O(N^2) worst case over total ops) a scan whose cost tracks the product of movable count and total ops, matching `build()`'s own "one `can_swap` call per pair with at least one movable endpoint" documentation.

Rebased onto `aw/logical-noise-signs` before pushing: the parent branch had gained two commits (`test(planner): narrow logical-noise-prefix inertness catch to overflow_error`, `test(planner): add DEPOLARIZE2 and PAULI_CHANNEL_1 sign-correction coverage`) since this branch's merge-base. One conflict, in `tests/test_logical_noise_prefix.cc`'s include block (this branch's helper-extraction commit had dropped `sampler.h`; the parent's later commit added `symplectic.h` for `anti_commute` in its new DEPOLARIZE2 test): resolved by keeping `symplectic.h` (still needed, used directly) and dropping `sampler.h` (unused after the extraction). Rebuilt and re-ran `[logical_noise_prefix]` after resolving: 20/20 test cases pass.

## Fixture expectations (C++, default options unless noted)

`coherent_d3_r3` reaches peak 4; `coherent_d5_r5` keeps peak 13 with dense work at most 0.45x the incumbent (narrowed to `beam_width=1` in this one C++ test section only, to keep it under Debug's ~15 s budget -- see the in-code comment); `cultivation_d5` keeps peak 10 and does not increase dense work (now decreases it further, see above); `surface_d7_r7_p001` is untouched via the early exit.

## Build, test, and docs

Debug: `cmake -B build-debug -DCMAKE_BUILD_TYPE=Debug -G Ninja && cmake --build build-debug -j8 && ctest --test-dir build-debug --output-on-failure`
- **978/978 tests passed**, total ctest wall time 106-108 s across repeated runs (most recent: 108.28 s; the parent branch's rebase added 3 tests to the prior 972).
- Standalone test file durations (`clifft_tests "[tag]"`): `test_active_width_analysis.cc` 4.9 s, `test_schedule_dependence.cc` 8.0 s, `test_active_width_schedule_pass.cc` 12.4 s. `test_logical_noise_prefix.cc` (the parent branch's file, checked after the rebase) passes all 20 cases.

Release benchmarks: `cmake -B build-release -DCMAKE_BUILD_TYPE=Release -DCLIFFT_BUILD_BENCHMARKS=ON -G Ninja && cmake --build build-release --target clifft_benchmarks -j8`
- `schedule_pass_coherent_d5_r5`: 522 ms (1 iteration, this run).
- `schedule_pass_cultivation_d5`: 149 ms (5 iterations, this run).
- `compile_schedule_plan_coherent_d5_r5` / `schedule_dependence_build_surface_d7`: see "Review feedback applied" above.

Python (Debug extension): `uv venv .venv && SKBUILD_CMAKE_BUILD_TYPE=Debug uv pip install -e . --no-cache --reinstall-package clifft --python .venv/bin/python && uv run --python .venv/bin/python pytest tests/python -x -q`
- **1392 passed, 7 skipped**, ~76 s (was 1391; +1 for the `beam_width=0` `ValueError` test).

Docs: `uv run --group docs --python .venv/bin/python mkdocs build --strict`
- Clean, exit 0. `docs/macros.py`'s registry entry says the pass is off by default and must run last, in three sentences; `docs/reference/python-api.md` gets `active_width_trace` and `ActiveWidthSchedulePass`. No other docs touched (theory page, guide subsection, default-flip, and the `leakage-and-loss.md` sentence are still PR C/D; this PR's code comments reference `docs/theory/active-width.md` as a forward pointer to that forthcoming page).

Lint: `uv run --frozen --only-group dev pre-commit run --all-files --show-diff-on-failure` -- all hooks pass. Six of the seven new commits passed clean on the first run; the closure-folding commit needed one clang-format line-wrap (a signature that got longer once `detail::` was added), clean on the second run.

## Not in this PR

`active_width_search.h/.cc`, its tests, the `clifft.search_width_schedule` Python binding and its tests, and `ActiveWidthScheduleOptions::exact_node_budget`/`maybe_exact_repair` are deleted outright, not deprecated. Comments that referenced the search file were rewritten to stand alone (now in `active_width_schedule_pass.cc`'s invariants comment, since `active_width_closure.h` itself is deleted -- see "Review feedback applied"). `tests/python/test_squeeze_benchmark_integration.py` and `tests/python/test_compile_passes.py` are untouched (they encode the default pipeline, unchanged here).

## Defensive tests

Four test-only commits on top of the reviewed head:

- `test: enrich the shared noisy-circuit generator` -- the generator in `tests/sampling_equivalence_helpers.h` now emits absorbed Cliffords (H, S, CX, CZ), Z_ERROR, DEPOLARIZE2, PAULI_CHANNEL_1, noisy measurements, classical feedback on earlier records, and an observable, with noise probabilities up to 1.0 so a wrong crossing sign cannot hide inside the binomial tolerance. `realize_noise` moved into the header and realizes every kind; the prefix test's local multi-channel generator is gone.
- `test: check exact per-realization equivalence of reordered plans` -- `check_exact_equivalent` replays both plans through `Executor::replay_shot` for every visible record under each presampled noise realization (all of them when there are at most 16, otherwise the all-none realization plus random draws) and requires identical reachability, joint log-probability within 1e-9, and identical detector and observable outputs, plus a per-realization sum-to-one check on the record enumeration. This is the per-realization claim behind noise transparency, checked with no statistics. Three consumers: pass output versus original (13 applied and 10 noise-crossing reorders across the checked trials), random linear extensions under transparency (35 crossing), and the prefix test's noise-crossing walk (23 checked). Teeth: clearing `logical_noise_prefix` on the scheduled HIR fails the pass consumer at its first crossing trial on record reachability. Circuits with readout noise are excluded because `replay_shot` reports any `ExecuteReadoutNoise` action unreachable; the statistical checks still cover them.
- `test(optimizer): schedule a noisy transition circuit through the pass` -- a LEVEL_TRANSITION on its own qubit next to the four-operation regression's Pauli bodies with a Z_ERROR between the rotations. Asserts the pass applied a crossing located after the instrument, the plan still builds (the planner's symbol prepass now has an instrument symbol ahead of the crossed site), the instrument mode is unchanged, and non-trapped shots agree statistically.
- `test(python): add the schedule pass to the coherent d3 semantic oracle` -- the three-way oracle becomes four-way with a scheduled variant at peak 4.

Debug wall time per tag: `[logical_noise_prefix]` 18.3 s to 19.4 s, `[schedule_dependence]` 8.1 s to 7.9 s, `[schedule_pass]` 12.4 s to 14.5 s. Verified at `481e8624`: Debug `ctest` 983/983, Debug Python extension 1392 passed and 7 skipped, strict docs build clean, pre-commit clean.

Two follow-up commits from the review of the tests above:

- `test: draw noise realizations from the library RNG` -- the shared helpers now take `clifft::Xoshiro256PlusPlus` instead of `std::mt19937`. The old `next_unit` applied the sampler's 64-bit uniform recipe to a 32-bit engine, which caps every roll near 2^-32 and below every generated noise probability, so `realize_noise` fired every channel and the legality oracle only ever saw all-fire realizations. `realize_noise` now returns fired and dropped counts and the legality oracle requires both to be positive (82 fired, 134 dropped at the current seed).
- `test(planner): require the exact walk check to cross noise` -- the walk-based exact test now counts `crossed_noise` (18 of 19 checked) so a later generator change cannot silently turn it into a non-crossing equivalence test.

All fixed-seed thresholds cleared with the new engine without retuning (pass exact test 10 applied, 9 crossed; dependence exact test 23 crossed). Verified at `05484639`: Debug `ctest` 983/983, pre-commit clean.

## Pass cost

Four commits after the defensive tests (the first three landed as five and were squashed into three logical commits for review; the fourth came from a later profile on a CCZ circuit). An external run of this pass on 15-to-1 magic-state distillation circuits (Stim exports: 14k ops at d=3, 70k ops and 539 qubits at d=5) and on 80k-op random Clifford+T circuits found that it never lowers width there and costs 4.8 s, 127 s, and 356 s respectively (Release, single core, pass alone). Profiling with perf (invoked as `/usr/lib/linux-tools-6.8.0-106/perf`, since the `/usr/bin/perf` wrapper refuses this kernel) and gprof attributed the cost to bookkeeping in three places:

- `perf(optimizer): keep the dormant subspace in reduced echelon form` (1e27a776). 84% of a 6.7 s structural trace at d=5 was the echelon-cache rebuild, invalidated by nearly every measurement and rebuilt from scratch by the next `contains()`. `DormantSubspace` now keeps one basis permanently in reduced row echelon form with O(dim x words) incremental updates (largest-pivot elimination on intersect, pivot-bit clearing on insert) and no rebuild; the row accessors live in the class body so the per-row loops inline them. Trace at d=5: 6.7 s -> 0.31 s. New tests assert the RREF invariants through the public API after every random update (confirmed to fail when either invariant is broken) and cross-check against an independent generator-list reference with brute-force membership.
- `perf(optimizer): skip implied pairs in the schedule dependence build` (1608be6b). The build tested every pair with a movable endpoint and stored every conflicting pair: 298M edges and 90 s at d=5, 1.06 billion edges and 293 s on the random circuit. Support-based pruning cannot help (initial-frame masks average 59 of 187 qubits), so the build now keeps a ring of per-op ancestor bitsets and skips every pair already implied by a recorded chain, scanning nearest-first so one link absorbs a whole ancestor set. Same transitive closure (a closure oracle test compares full ancestor sets against the pairwise relation, with and without ring eviction; a 40-measurement alternating-basis chain keeps exactly N-1 edges), 2,600x fewer edges, 90 s -> 0.55 s. The adjacency lemma that `sink_neutral_rotations` relies on is documented at both ends; `ScheduleDependenceOptions::ancestor_cache_bytes` (64 MiB) bounds memory. The "Scaling note" earlier in this body is superseded.
- `perf(optimizer): stop closure sweeps from rescanning stale state` (85746c21). Two scans inside the closure sweep repeated work that could not change between steps. The ready-scan hint only advanced one step when its own witness executed, so each closure step re-walked O(n) stale flags (82% of the remaining 7.2 s on the random circuit); `lowest_ready()` now stores the exact first ready index back. And every step re-ran `is_expanding` on every ready rotation below the first non-expanding op (82% of the coherent d5 r5 pass); a rotation found expanding stays expanding while the subspace only grows or holds, and inside a sweep only a `MeasureDormantRandom` step shrinks it, so `SearchFrontier` keeps a generation-stamped memo that `run_closure` resets on entry and after each shrink, with a Debug assert re-checking every hit.
- `perf(optimizer): scan closure candidates a word at a time` (f59dd40f). Found on a direct-T CCZ d5 fixture (100k HIR ops: 62k noise sites, 16k measurements, 7k detectors, 28 EXP_VAL barriers, 8 T gates), where the pass took 14 s and perf put 66% of it in the flag walk inside `run_closure`: an early T that is ready but not yet chosen sits at the lowest ready index, so every closure step walked from it across thousands of not-ready flags to the next chain op, O(n) per step and O(n^2) per sweep. Readiness and the expanding memo are now bitsets scanned a word at a time from a candidate hint, so a step costs the words skipped rather than the ops. 14.1 s -> 5.0 s on that circuit; other circuits unchanged within noise.

None of the four changes a decision: outputs are bit-identical. Pass wall time, Release, single core, pass alone on the prepared HIR:

| Circuit | Ops | Before | After |
|---|---:|---:|---:|
| 15-to-1 d=3 | 14,464 | 4.8 s | 0.15 s |
| 15-to-1 d=5 | 69,853 | 127 s | 1.8 s |
| random N_L=5 | 80,875 | 356 s | 1.0 s |
| random N_L=7 | 80,581 | (not re-run) | 1.6 s |
| coherent d5 r5 (corpus) | | 555 ms | 165 ms |
| cultivation d5 (corpus) | | 192 ms | 14 ms |
| direct-T CCZ d5 (8 T, 100k ops) | 100,297 | 14.1 s | 5.0 s |

Outputs are unchanged: the reordered HIR (op sequence, masks, record indices) hashes identically before and after on 28 circuits (every repo fixture, the clifft-paper corpus, and the distillation exports), re-checked after every step including the final tree. Verified at f59dd40f: Debug ctest 987/987 (was 983; four new tests), Debug Python extension 1392 passed / 7 skipped, pre-commit clean; each of the first three commits builds on its own. The CCZ output hash is also identical before and after the fourth commit. Benchmarks: `schedule_pass_coherent_d5_r5` 522 -> 169 ms, `schedule_pass_cultivation_d5` 149 -> 15 ms, `compile_schedule_plan_coherent_d5_r5` 523 -> 172 ms, `schedule_dependence_build_surface_d7` 50.7 -> 1.6 ms.

What remains on the CCZ circuit, per sweep counters on a scratch build: 204 closure sweeps executing 3.26M ops, about 32 full replays of the measurement algebra. From the third beam step on, each of the 8 beam members is the last blocker of the fixed chain once per step and sweeps the whole tail; that is the beam search working as designed (the exact search certifies its result, peak 5, as the trace-class minimum over 182 states), so cutting it further would mean a search-design change (for example bounding candidate sweeps by the running peak, which needs a dedup-aware argument), not another bookkeeping fix. On that circuit the production pipeline itself now costs as much as the pass: `StatevectorSqueezePass` spends about 4 s in `can_swap` against the 62k noise sites through its adjacent-swap walk, which is outside this PR. What remains elsewhere, per perf on the final build: on measurement-heavy circuits the cost is the per-measurement subspace algebra (`intersect`, `reduce_into_scratch`, already AVX-512 vectorized per `perf annotate`) times the handful of replays the beam search needs, plus one `can_swap` per untested pair in the build; on rotation-heavy circuits it is the remaining `is_expanding` calls from candidate scoring and post-shrink re-tests, and the ready-flag walk itself. Nothing left is a bookkeeping accident of the kind fixed above.

Why the pass finds nothing on those circuits: an exact search over closed states (Dijkstra on executed-op sets, exact by the closure theorem; not committed) shows the breadth-first ordering of the 15-to-1 circuit has exactly one ready expanding rotation at every step (16 states in total) with minimum peak 6, and the depth-first ordering's minimum is 5, so the width-5 ordering is a source-level rewrite outside the width-6 ordering's trace class. The same search reproduces the certified peaks for coherent d3 r3 (4), cultivation d3 (4), and cultivation d5 (10, 25.9k states in 2.3 s).

## Search budget (default-on groundwork)

`feat(optimizer): bound the schedule search with a swept-op budget` (d6ab028c). The beam's cost is circuit-shape dependent: unbounded, beam width 8 sweeps up to 140x the op count on small coherent circuits and 35x on the 100k-op CCZ circuit, while greedy (width 1) always finishes in about 2x and reaches the same peak on most circuits. `ActiveWidthScheduleOptions::search_budget` (a `std::optional<double>`, `None` in Python for no bound) bounds the search in swept ops per HIR op, counted deterministically (initial closure, every scoring sweep including discarded candidates, every survivor replay) so plans stay reproducible across machines. Once over budget, the remaining lower-ranked parents of the current generation are dropped unscored, the next cut keeps one survivor, and the search finishes greedily. `swept_ops()` reports the count.

The default of 8 was vetted on 17 circuits (the clifft-paper corpus, both 15-to-1 d=3 orderings, 15-to-1 d=5, random N_L=3/5/7, the CCZ circuit, qv10, target_qec) at budgets 0, 2, 4, 8, 16, 32 and unbounded, against the exact trace-class minima certified earlier: every budget of 2 or more reaches the unbounded search's peak on every circuit; budget 0 (greedy) loses only distillation (4 vs 3); 8 is the smallest budget that also keeps every dense-work gain (qv10 needs it); and on coherent d5 r5 the narrowed search is better than unbounded (dense work 847,904 vs 870,180). Cost at the default, swept ops per HIR op: at most 10 everywhere (CCZ 35 -> 10, 5.2 s -> 2.3 s; coherent d5 r5 139 -> 22; coherent d3 r3 89 -> 14), unchanged on circuits that never cross it. With the budget left unbounded all 29 output hashes match the pre-budget baseline exactly; at the default only coherent d5 r5 changes. The first push used `infinity` as the unbounded value and failed `release-cpp-smoke`: under the pinned clang 22 with the project's `-ffast-math`, the comparison against an infinite budget folded to true and the unbounded run narrowed immediately (swept 1985, the greedy count, versus 3686 at the default), which the new narrowing test caught. Reproduced locally with the same compiler via `.github/scripts/install_ubuntu_clang.sh`; the budget is now an empty optional when unbounded, validated at the bit level (`detail::is_finite_non_negative`) since `std::isnan`/`std::isinf` are equally unreliable under that flag, and non-finite values are rejected in both C++ and Python. Verified at d6ab028c: Debug ctest 992/992 (five new tests: negative and non-finite budgets rejected, budget 0 byte-identical to beam width 1, the default narrows on coherent d3 r3 without losing its peak, swept-op statistics), pinned-clang Release x86-64-v2 ctest 986/986, Debug Python extension 1396 passed / 7 skipped (four new tests), pre-commit clean.

A static-priority scheduler was spiked first as the cheaper alternative and rejected: earliest-anticommuting-consumer and lowest-index rules give identical schedules and miss the certified optimum on coherent d3 r3 (5 vs 4), distillation (4 vs 3), and CCZ (7 vs 5); the closure-based lookahead is what finds the wins. Not committed.

## Review follow-ups on the budget

Four commits answering an external review of the stack.

- `fix(optimizer): bound candidate scoring once the search budget is exhausted` (3432b463). The reviewer showed that narrowing the beam to one parent does not bound the search: that parent still scores every ready expanding rotation each step, so cost is quadratic in the number of simultaneously ready rotations. Reproduced with independent `H q; T q; H q; M q` blocks: 256 blocks (512 HIR ops) swept 135 ops per HIR op at the default. The budget now has two graduated responses on one number: at half the budget the beam narrows to one parent and one survivor (the rule vetted earlier), and at the full budget each step also scores only the lowest-index ready expanding op. The default is 16, so the parent threshold stays at the vetted 8 traces. A first attempt that stopped scoring candidates at the same threshold as the parents lost coherent d3 r3's peak-4 schedule and broke the budget-0 identity with beam width 1; the graduated rule keeps the greedy tail's candidate scoring, which is cheap on narrow circuits and is what finds those schedules. New regression on the block family: swept ops at most 20x the op count and at most 3x growth per doubling (measured 9x the op count and a 2.0 ratio; unbounded is 514x at 256 blocks).
- `docs(optimizer): state the scheduling pass's acceptance rule and cost bound accurately` (75eaac1a). The registry text claimed the pass never worsens either metric; the rule is lexicographic and dense work can rise when peak falls (the reviewer's eight-op case: peak 4 to 3, dense work 52 to 54). The text now states the exact rule, the budget, and the run-last requirement in plain prose.
- `refactor(optimizer): drop unused scheduling helpers and reuse the robust finiteness check` (db4852d0). Removes `detail::pauli_body`, `SearchFrontier::known_expanding()`, the single-use closure-absorb wrapper, and the `completed` vector plus `pick_best_completed` (a running best with the same comparison, same encounter order). The budget's finiteness check now uses `is_finite_robust` and `opaque_binary64_bits` from `util/numeric.h`, whose comments record that even a direct `bit_cast` exponent test has been folded under fast-math on Apple Clang. The Python binding no longer duplicates the constructor's validation; nanobind translates `std::invalid_argument` to `ValueError`, and the existing tests confirm it.
- `chore(optimizer): keep scheduling comments free of development history` (8f69b3c3). Comment-only.

Re-vetted at the default of 16 over the same 17 circuits plus the block family: every peak matches the unbounded search; qv10's dense-work gain is kept; coherent d5 r5 lands at 856k (unbounded 870k); CCZ stays at 10 op-counts, 2.2 s. Unbounded reproduces all 29 baseline output hashes; the default differs only on coherent d5 r5. Verified at 8f69b3c3: Debug ctest 993/993, pinned-clang Release x86-64-v2 ctest 987/987, Debug Python extension 1396 passed / 7 skipped, pre-commit clean.

Not taken from the review: removing the commutation fingerprint. It catches HIR mutation between building the relation and applying a schedule, which binding the two calls to one object would not, and it costs one hash pass.


bachase:
Reviewed and verified independently at `e84eddd1`.

- Structure: the early exit runs the cheap analysis first and returns before the quadratic dependence build when the incumbent peak is zero or nothing can expand; the flat frontier keeps the lowest-index-first closure order the theorem needs; no reference to the exact search or the repair hook remains; the closure header now carries the persistence and measurement arguments in the form the theory page uses.
- Local Debug build in a separate worktree: `ctest` 969/969 in 15 s wall; Debug Python extension in a worktree-local venv, 1391 passed, 7 skipped; strict docs build clean.
- Release harness on the `clifft-paper` corpus, production pipeline versus production plus the pass at default options: coherent d3 r3 peak 5 to 4, plan work 1247 to 381, 1.15M to 2.19M shots per s (12 ms of pass time); coherent d5 r5 peak 13, plan work 1.99e6 to 7.8e5, 2089 to 5017 shots per s (572 ms); distillation peak 5 to 3, 1.19M to 1.28M (6 ms); cultivation d5 peak 10, plan work 5.66e4 to 5.47e4, unchanged throughput within noise (152 ms). Bit-identical to the numbers on the closed stack, as the PR body states.
- The mask-view overloads are the right shape: the `PauliString` entry points are one-line forwarders, and no scheduler path allocates per op any more.

Stacked on #460 as noted; no further changes requested. Default stays off in this PR; the flip is a separate decision once CI time with the pass in the default pipeline is measured.

bachase:
CI is green at `cdbfee9d`, including the Windows Debug job. The only failure was the early-exit test's absolute 50 ms wall-clock bound, which came from my spec and which an unoptimized MSVC build cannot meet; the test now compares the pass against the dependence build it skips on the same host (about 150x slower there), so it is host-independent. #472 is rebased onto this head.

bachase:
Follow-ups in progress from the external review, all accepted:

- Beam dedup by `(peak, dense_work, first_op)` instead of dense work alone, with a regression where two branches converge on one executed set with different peaks.
- `ScheduleDependence` bound to a fingerprint of its source HIR; `apply_schedule` rejects a foreign same-sized HIR. Both types move to `clifft::detail`.
- Early-exit test checks a `built_dependence()` statistic instead of timing.
- Fixture-skip helper catches `std::overflow_error` only.
- `beam_width == 0` rejected in C++ and Python; identity fallback removed.
- Closure machinery folded into the pass; the theorem lives on the docs page, the code keeps short invariants.
- End-to-end `schedule + plan_sampling` benchmark and a dependence-build scaling benchmark.

One review item is declined on evidence: deferring noise transparency. Release measurements on the `clifft-paper` corpus with the pass at default options, transparency off versus on: coherent d3 r3 1.22M versus 2.26M shots per s, coherent d5 r1 111k versus 4.18M, coherent d5 r5 2307 versus 5109, distillation 1.22M versus 1.28M, cultivation d5 unchanged either way. Without transparency the pass is worth five to fifteen percent; with it, two to fifty times, and the exact search certifies the plain-`can_swap` production peaks as already optimal on every circuit it can finish. The planner's correction scans cost about a millisecond on the noisiest corpus circuit (`plan_sampling` 6.2 ms versus 4.7 ms on cultivation d5), which the new end-to-end benchmark will track.

bachase:
Re-verified at `82325d02` after the review follow-ups: local Debug `ctest` 975/975; Debug Python extension in a worktree-local venv, 1392 passed, 7 skipped; strict docs build clean; Release harness on the `clifft-paper` corpus unchanged from the previous head (coherent d3 r3 peak 4 / plan work 381, coherent d5 r5 13 / 7.8e5, distillation 3 / 71, cultivation d5 10 / 5.47e4) with the same throughput within noise.

On the individual findings:

- The dedup fix was substantive: on the frozen regression circuit the old rule found no improvement at all where the brute-force optimum is one unit lower, and the fixture cultivation d5 picked up a few percent of dense work from the same change.
- The source-HIR fingerprint closes the foreign-relation hole without touching the pass's normal path, and both dependence types are now `detail`.
- The early-exit test is structural (`built_dependence()`), the fixture-skip helper catches `std::overflow_error` only and surfaced nothing new, zero beam width throws on both sides, and the closure code is private to the pass with the theorem left to the docs page.
- The new benchmarks put numbers on the two cost questions: schedule plus plan on coherent d5 r5 is 523 ms end to end, and the dependence build scales from 51 ms on surface d7 to about 1.06 s on surface d11, which is the quadratic term and the reason the early exit matters; a qubit-support prefilter remains the listed follow-up.

No further changes requested.

bachase:
Confirmed the fresh review's fingerprint hole at `82325d02` with the detector counterexample: the relation from `M 0; M 1; DETECTOR rec[-2]` accepts `{0, 2, 1}`, the fingerprint accepts the `rec[-1]` variant, and `plan_sampling` then throws. Root cause: the fingerprint was derived from the HIR data model rather than from what `can_swap` reads, so side-table contents behind detector, observable, and readout indices were never hashed. Fix in progress: a `commutation_fingerprint` defined beside `can_swap` in `commutation.cc` covering exactly its inputs, with the reproduction and a per-input-category mutation test as regressions.

bachase:
Re-verified at `2d598506`: local Debug `ctest` 978/978; Debug Python extension in a worktree-local venv, 1392 passed, 7 skipped; strict docs build clean. My standalone build of the review's counterexample now stops at `apply_schedule` with "target HIR's fingerprint does not match the HIR the dependence relation was built from" instead of reaching the planner.

The fix is the right shape: the fingerprint is defined in `commutation.cc` beside `can_swap`, hashes the dereferenced side-table contents (target lists with their lengths, the readout entry's measurement index, every noise channel's mask words) rather than the indices, and drops the fields `can_swap` never reads. The regression covers the reproduction plus one mutation per input category, and the fingerprint is proven independent of `logical_noise_prefix`. No further changes requested.

bachase:
Round-three follow-ups in progress, both accepted: beam dedup keeps the Pareto front of (peak, dense work) per executed set before the global beam cut, with a total tie-break on the full prospective order and a regression that checks the secondary objective after a later expansion equalizes the peaks; and `CommutationFingerprint` moves into `clifft::detail` beside its only consumer.

bachase:
Re-verified at `b63838fa`: local Debug `ctest` 979/979; Debug Python extension in a worktree-local venv, 1392 passed, 7 skipped; strict docs build clean.

Both round-three items are in the shape asked for. Dedup now keeps the Pareto front of `(peak, dense_work)` per executed set, collapsing exact ties by the full prospective order so the survivor is independent of generation order, and only then applies the global beam cut. The new 12-op regression is the case the review described: one executed set reached at `(3, 54)` and `(4, 52)`, a suffix that forces peak 4 either way, and the old rule finishing at `(4, 70)` where the brute-force optimum is `(4, 68)`. Corpus fixture numbers are unchanged, which is expected since none of those circuits exhibits the tradeoff. `CommutationFingerprint` is in `detail` beside its only consumer. No further changes requested.

src/clifft/optimizer/active_width_schedule_pass.cc
```diff
@@ -0,0 +1,1082 @@
...
+    // has found expanding; the op it returns is the same lowest-index
+    // ready-and-not-yet-classified op an unmemoized linear scan would reach
+    // next. Because known-expanding ops never come back around, there is no
+    // "memo hit" branch left to cross-check here: every op this loop sees is
+    // one is_expanding has not yet classified this sweep, so it always runs
+    // fresh.
+    while (const std::optional<uint32_t> op = frontier.first_candidate()) {
+        if (is_expanding(hir, hir.ops[*op], subspace)) {
+            frontier.note_expanding(*op);
+            continue;
```
bachase-agent, line 364: **[P2] Bound expansion checks as well as executed operations**

Each closure resets the expanding memo and scans the remaining ready rotations, but these classification checks do not increment `swept_ops`. After the budget is exhausted, every single-candidate sweep can therefore reclassify all other ready rotations. The search retains quadratic classification work despite the claimed constant-multiple-of-traces bound.

Reproduction: independent `H q; T q; H q; M q` blocks, prepared with `PeepholeFusionPass` and `StatevectorSqueezePass`, then scheduled with `search_budget=0`. Adding a counter to `is_expanding` gives:

| Blocks | HIR ops | `swept_ops` | `is_expanding` calls |
|---:|---:|---:|---:|
| 128 | 256 | 512 | 8,512 |
| 256 | 512 | 1,024 | 33,408 |
| 512 | 1,024 | 2,048 | 132,352 |
| 1,024 | 2,048 | 4,096 | 526,848 |

At the default budget, 256 and 512 blocks require 1,066,081 and 4,294,881 checks respectively, while `swept_ops` only doubles. The existing regression checks only `swept_ops`, so it misses this work.

Please account for classification probes in the deterministic budget and provide a bounded-cost legal fallback after exhaustion, with a regression on probe counts. The circuit-shape-independent cost guarantee in #472 and `docs/macros.py` also needs to match what is actually bounded.

Verified at `8f69b3c3`; the existing 993 C++ tests pass in Release with assertions enabled.

bachase-agent:
Superseded by #507, which consolidates the final active-width implementation onto main in five ordered review commits. Please continue review at https://github.com/unitaryfoundation/clifft/pull/507; this PR remains available as development history.

PR #53, opened 2026-05-04
feat: runtime-width SVM Pauli frame, drop kMaxInlineQubits ceiling

bachase:
## Summary

Final PR of the issue #45 staged migration. Migrates the SVM Pauli frame (`SchrodingerState::p_x` / `p_z`) from fixed-width `BitMask<128>` to runtime-sized `std::vector<uint64_t>`, removes `CLIFFT_MAX_QUBITS` / `kMaxInlineQubits` / `PauliBitMask` entirely, and brings the docs and Python API up to date. The only remaining hard ceiling is the 65,536-qubit VM-axis bound (uint16_t operands).

## Commits

1. **`test: add boundary-driven width-sweep fuzz harness`** — `tests/test_fuzz.cc`. Sweeps n ∈ {32, 64, 100, 128} and touched qubits ∈ {0, 63, 64, n-1} over deterministic patterns (M, R, X_ERROR, classical feedback, MPP, EXP_VAL, H/M random measurement) and asserts exact outcomes. Lands first to give the SVM frame migration a safety net.
2. **`feat(svm): runtime-width Pauli frame in SchrodingerState`** — replaces `PauliBitMask p_x/p_z` with `std::vector<uint64_t>` sized at construction; threads `program.num_qubits` into the four `sample()` call sites; drops the `n > kMaxInlineQubits` gate in `lower()`. Bit ops in `svm_math.h` move to `std::span<uint64_t>`. The per-op-type rejection tests from PR2 (`REQUIRE_THROWS_AS`) flip into round-trip checks that trace, lower, and sample with deterministic-outcome assertions.
3. **`refactor: remove CLIFFT_MAX_QUBITS, kMaxInlineQubits, PauliBitMask`** — drops the unused config knob, the constant, and the typedef. The dead `anti_commute(PauliBitMask&)` overload goes away. Tests that compared `MaskView == BitMask<N>` keep that path; one test helper moves to `uint64_t`-low-word comparisons.
4. **`docs: drop CLIFFT_MAX_QUBITS, document the 65,536-qubit VM ceiling`** — surface code guide, architecture page, and building docs all stop referencing the compile-time qubit limit. `pyproject.toml` and `CMakeLists.txt` drop the build-system plumbing. `clifft.max_sim_qubits()` now returns 65,536; `tests/python/test_api.py::test_max_sim_qubits` updated.
[...]

PR #473, opened 2026-09-02
docs: add v0.10 performance update

bachase:
## Summary

- complete the v0.10.0 development update with matched v0.10-versus-v0.9 benchmark results
- add a dedicated Performance guide covering near-Clifford throughput, release history, and dense Quantum Volume results
- add concise performance messages to the docs home page and README
- combine Clifft and SymFT absolute throughput and per-workload speedup ratios in one guide figure, replacing the separate rate table
- retain the existing Clifft-only absolute-throughput figure in the README
- keep rendered documentation assets here while benchmark data and reporting code remain in clifft-bench
- document the optional cross-repository benchmark refresh workflow and the distinct README/guide figures in the release guide

## Benchmark framing

- Clifft v0.10 leads SymFT v0.1 on all eight single-core near-Clifford workloads: 1.05x to 87.7x, 1.53x median
- Clifft v0.10 leads v0.9 on all eight workloads: 1.06x to 837x, 3.23x median
- the text distinguishes packed wins from compiler-driven gains, including the 837x active-width result
- the dense QV section reports where Clifft leads and where qsim leads, and documents the asymmetric timing boundaries

Source campaign and reporting: https://github.com/unitaryfoundation/clifft-bench/pull/53

The combined light/dark chart is copied directly from clifft-bench commit 7e1f58311c5639589154b68e99f2a8ff1b9e215a. Merge the benchmark reporting PR before this documentation PR.

## Validation

- `just docs-build`
- `git diff --check`
- all eight combined-chart rate pairs and ratios checked against the source comparison table
- combined PNGs match the benchmark outputs byte-for-byte; README and its existing images unchanged


docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+Clifft is built for fast, exact simulation of near-Clifford circuits. Its
+compiler resolves Clifford coordinates and symbolic dependencies before
+sampling, then executes only the remaining branch values and dense active-state
+operations for each shot.
+
+The measurements below cover three complementary questions: how Clifft
+compares with another near-Clifford CPU simulator, how its throughput has
+changed across releases, and how it behaves when a circuit becomes fully dense.
+
+!!! info "At a glance"
```
bachase: Drop the at a glance callout

README.md
```diff
@@ -67,6 +67,22 @@ An [experimental AMD HIP backend](https://unitaryfoundation.github.io/clifft/sta
 is available through an explicit source build and separate API. It is not part
 of the published wheels or selected automatically.
 
+## Performance
+
+On eight calibrated near-Clifford workloads using one pinned CPU core, Clifft
```
bachase: Rather than a comparison framing, how about something like "millions to XX" shots per second on near-Clifford workloads and then link to performance page. Something like "think stim like performance for near-Clifford simulations"

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+    On the eight calibrated, single-core near-Clifford workloads below, Clifft
+    v0.10 is faster than SymFT v0.1 on every workload. The advantage ranges
+    from **1.05x to 87.7x**, with a **1.53x median** across the workload set.
+
+## Near-Clifford throughput
+
+The recurring [`clifft-bench`](https://github.com/unitaryfoundation/clifft-bench)
+campaign measures attempted shots per second for complete circuits on one
+pinned logical CPU. For each workload and simulator it tests batch capacities
```
bachase: Drop the details on batch -- just say selects best batch size for that workload.

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+of 1, 32, 256, 1024, and 2048 when supported, then measures the fastest
+configuration. Batch capacity 1 is scalar execution.
+
+![Clifft v0.10 throughput relative to SymFT v0.1 across eight near-Clifford workloads](../assets/performance/clifft-vs-symft-light.png#only-light)
```
bachase: I see the filled/non-filled etc but since they overlap so much, is it really telling us anything? I'd just drop that level of detail and have a single dot and then drop the legend.

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+![Clifft v0.10 throughput relative to SymFT v0.1 across eight near-Clifford workloads](../assets/performance/clifft-vs-symft-dark.png#only-dark)
+
+Diamonds indicate that SymFT selected a packed configuration; circles indicate
+scalar SymFT. Filled markers mean Clifft also selected packing. Both tools were
+calibrated independently for every workload.
+
+| Workload | Clifft attempted shots/s | Clifft batch | SymFT batch | Clifft / SymFT |
```
bachase: As above, drop the detail on batch/batch selection in the table; I think the shots/second might still be useful to give the absolute speed. Is that better as a table or a plot?

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+## Performance over time
+
+Clifft's compiler-like structure provides several independent places to make
+simulation faster: circuit optimization, symbolic planning, executable
+preparation, and scalar or packed active-state kernels. That structure does not
+mean every speedup comes from the compiler, but it lets later releases improve
+one stage without moving circuit analysis back into the per-shot execution
+loop.
+
+![Median Clifft throughput by release relative to v0.1](../assets/performance/performance-over-time-light.png#only-light)
```
bachase, line 46: For the packing + compile rcallout, don't render the v0.10.0 because that is already in the x-axis

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+![Median Clifft throughput by release relative to v0.1](../assets/performance/performance-over-time-dark.png#only-dark)
+
+The first broad step arrived in v0.8, when symbolic plans replaced the original
+localized-Pauli virtual machine. Version 0.10 combines another compiler
+improvement with packed sampling: its median throughput is **3.23x v0.9** and
+**7.0x v0.1** across the eight workloads.
+
+Packing was selected for four v0.10 workloads and scalar execution remained
```
bachase: Again drop the details of the packing/scalar vs batch. Those are more implementation details.

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+simulation faster: circuit optimization, symbolic planning, executable
+preparation, and scalar or packed active-state kernels. That structure does not
+mean every speedup comes from the compiler, but it lets later releases improve
+one stage without moving circuit analysis back into the per-shot execution
+loop.
+
+![Median Clifft throughput by release relative to v0.1](../assets/performance/performance-over-time-light.png#only-light)
+![Median Clifft throughput by release relative to v0.1](../assets/performance/performance-over-time-dark.png#only-dark)
+
+The first broad step arrived in v0.8, when symbolic plans replaced the original
```
bachase: LInk to the release blogs for each of these?

docs/guide/performance.md
```diff
@@ -0,0 +1,122 @@
...
+[complete result table](https://github.com/unitaryfoundation/clifft-bench/blob/f02d8496ee9269c9fa25a4cf4bdb982ffd8a28e2/experiments/qv/results/qv-0.10.0rc1-20260902/cases.csv).
+
+## Scope and next steps
+
+The near-Clifford comparison is CPU-only. A future campaign will cover GPU
+execution, including Tsim, under a separate hardware and measurement contract.
+A pure-Clifford comparison with Stim will likewise be added after the relevant
+Clifft execution changes and benchmark cases land.
+
+The v0.10 measurements use the published `v0.10.0rc1` artifact, which contains
```
bachase: Drop this details on v0.10 -- just have this section about future benchmark scope.

README.md
bachase: Actually, would one of the plots from the performance page work well here? Either vs symft? Or a similar one that is absolute TPS in that same visual format?

Or could be the relative performance plot with dual axis?

docs/guide/index.md
```diff
@@ -5,6 +5,10 @@ Clifft, how it is compiled, which scientific result to request, and how that
 work runs on the available hardware. Start with the result you need; execution
 tuning comes last.
 
+See [Performance](performance.md) for calibrated near-Clifford comparisons,
```
bachase: Here and elsewher, I think calibrated is too confusing. I'd drop it in this commentary.

docs/updates/packed-sampling.md
```diff
@@ -0,0 +1,124 @@
...
+efficiently. Packed sampling instead treats several shots as lanes of one
+executor. Boolean state is stored in packed columns, active-state coefficients
+are interleaved across lanes, and each prepared action is applied to the group.
+The trajectories remain mathematically independent, but the executor shares
+more of the work around them.
+
+The approach was inspired by the packed cross-shot path in
+[SymFT](https://arxiv.org/abs/2607.28600), whose results on low-active-width
+circuits helped motivate a similar mode in Clifft. Clifft adapts the idea to
+its prepared sampling plans and result APIs. The dedicated Performance page
```
bachase: Drop the "dedicated performance page" sentence.

docs/updates/packed-sampling.md
```diff
@@ -0,0 +1,124 @@
...
+$k$ complements both packed and threaded execution.
+
+Apple Silicon also gains native NEON kernels for direct and fused rotations and
+diagonal active measurements. Clifft selects them automatically when an
+operation is large enough to benefit and retains scalar fallbacks for smaller
+work.
+
+## Measured against v0.9.0
+
+The matched `clifft-bench` release campaign calibrates each version over the
```
bachase, line 83: This is the one spot where I think calibrate works well.

bachase (commented):
Should the benchmark data and plotting code live in this repo, or in clifft-bench?

README.md
bachase: Reframed the README around absolute throughput and replaced the comparison figure with a Clifft-only attempted-shots-per-second plot. I skipped the dual axis so the README keeps one clear message; the Performance guide carries the comparison.

docs/updates/packed-sampling.md
bachase: Removed that sentence.

docs/guide/performance.md
bachase: Linked the v0.8 and v0.10 release blogs from the corresponding discussion.

docs/guide/performance.md
bachase: Updated the callout to just “packing + compiler”; the release remains on the x-axis only.

docs/updates/packed-sampling.md
bachase: Kept “calibrates” here as the precise description of the release-campaign procedure.

docs/guide/performance.md
bachase: Simplified this to say the campaign selects the best batch size for each workload.

docs/guide/performance.md
bachase: Removed the callout.

docs/guide/performance.md
bachase: Removed the release-candidate details and made this section solely about future benchmark scope.

docs/guide/performance.md
bachase: Removed the packed/scalar selection discussion here. I retained the active-width explanation for the 837x outlier without presenting it as batching.

docs/guide/performance.md
bachase: Agreed. The SymFT comparison now uses one dot style and no legend.

docs/guide/performance.md
bachase: Removed both batch columns. I kept a compact table for exact absolute rates and ratios; the README now uses the absolute-rate plot for the more visual presentation.

docs/guide/index.md
bachase: Removed “calibrated” from the general docs commentary and homepage. It remains only in the release-blog sentence that explains the calibration procedure.

bachase:
Addressed the source-ownership question as well: benchmark data and plotting/reporting code stay in `clifft-bench`; this PR now keeps only the rendered documentation assets. The duplicated CSV snapshots and `tools/plot_docs_performance.py` have been removed.

docs/guide/performance.md
```diff
@@ -0,0 +1,108 @@
...
+size for that workload before collecting the comparison.
+
+![Clifft v0.10 throughput relative to SymFT v0.1 across eight near-Clifford workloads](../assets/performance/clifft-vs-symft-light.png#only-light)
+![Clifft v0.10 throughput relative to SymFT v0.1 across eight near-Clifford workloads](../assets/performance/clifft-vs-symft-dark.png#only-dark)
+
+Clifft is faster than SymFT on all eight workloads. The advantage ranges from
+**1.05x to 87.7x**, with a **1.53x median** across the workload set. Absolute
+Clifft rates are listed alongside the ratios for scale.
+
+| Workload | Clifft attempted shots/s | Clifft / SymFT |
```
bachase: Should this use the plot that is now in the README.md instead?

bachase: Agreed. The Performance guide now uses the same absolute-throughput plot as the README, with the SymFT ratios retained in the table below. Removed the unused relative-comparison image pair. Included in 2656b7d6.

bachase: Updated to the combined-chart design agreed in our follow-up: the Performance guide now shows both Clifft and SymFT attempted shots/s, plus a speedup column, in one figure. This replaces the plot and markdown table. The README retains its existing Clifft-only plot unchanged. The generator and copy instructions are in clifft-bench PR #53; the docs update is signed and pushed as 32801400.

Issue #475, opened 2026-09-03
Shared CPU oracle conformance foundation and feature onboarding


bachase:
## Goal

Establish a reusable, independently checked CPU conformance foundation so compiler and execution changes inherit integration coverage. A new feature must show both that its intended path ran and that the resulting behavior agrees with an independent expectation.

**Status: in progress. #476 is the first foundation PR against main. It does not complete this issue.** The next priority is bounded noisy joint-distribution coverage, followed by adoption in the active-width scheduling series (#460, #471, #472). Deliver small, independently reviewable PRs.

### 1. CPU foundation: #476

- Share circuit expectations, compiler profiles, and CPU sampling configurations. Cover HIR optimization disabled, the actual production-default factory, scalar execution, packed batches, and automatic selection.
- Preserve the existing Qiskit statevector cases and amplitude assertions under both compiler profiles. Add representative independent terminal-record checks and complete small record histograms, including detector and observable mappings.
- Require real peephole/squeeze transformations. Guard the pass inventory so additions require a coverage decision.
- Reject unsupported oracle inputs instead of silently removing physical operations or malformed syntax. Related: #37.
- Check deterministic nonzero outputs at word and partial-batch boundaries, where aggregate statistics can hide a few corrupt rows. Substantiate actual native execution separately from argument forwarding.
- Commit reproducible negative controls, preserve existing regressions, and report comparable baseline/changed runtimes. No noisy-reference or GPU-runtime claim belongs to this increment.

### 2. Bounded noisy joint reference

Implement a small, inspectable independent reference for selected circuits combining non-Clifford rotations, Pauli noise, mid-circuit measurement, reset, and classical feedback. Calculate complete visible-record probabilities, with hidden branches marginalized correctly, and check detector/observable results and a supported postselection subset, including acceptance probability and conditional outputs.

- Derive expectations from original circuit semantics. Do not reuse Clifft HIR, coordinate planning, lowering, or executor decisions as the reference calculation.
- Bound qubit count, record width, and branch growth explicitly. Start with operations needed by designated cases; reject unsupported operations. This is not a general-purpose simulator project.
- Validate the reference itself using analytic examples and Aer/Stim where applicable. Keep independent end-to-end comparisons primary; Clifft-to-Clifft exact replay remains useful supplementary evidence.
- Declare compatible case/configuration combinations. Semantics-changing passes need their own expectations. Record-order restrictions and unsupported execution contracts must be tested explicitly, not hidden by broad skips.
[...]

PR #476, opened 2026-09-03
test: establish shared CPU oracle conformance coverage

bachase:
Compiler and sampler changes can pass focused tests while missing integration regressions: the independent statevector corpus previously disabled HIR optimization, and aggregate sampling checks can miss incorrect correlations or a few unwritten tail rows. This PR runs reusable independent expectations through the production-default and unoptimized CPU configurations, with explicit transformation witnesses and exact output-boundary checks.

This is the first CPU foundation increment of #475. It changes tests and development documentation only.

## Coverage

| Coverage | Implementation and expectation |
|---|---|
| Existing Qiskit corpus | Preserve all 123 circuits/seeds/assertions under each of two compiler profiles: 246 cases. Compare phase-aligned amplitudes against Aer. The default profile calls the real factory with fresh pass instances. |
| Small shared sampling corpus | Five circuits cover fusion, squeezing, Bell correlations, mixed Pauli rotations, and a mirror. Check all terminal-record probabilities against Aer, then complete record histograms and detector/observable mappings: 10 exact checks and 30 sampling configurations. |
| CPU execution configurations | Single-shot, packed capacity 65, and automatic selection, each explicitly using one worker. References are cached by source/worker; compiled sampling programs are reused across modes. |
| Exact output boundaries | A separate analytic circuit has 65 measurement columns with nonzero bits on both sides of a word boundary and distinct detector/observable rows. Check every row at 63, 64, 65, 66, 129, 130, 131, and 2049 shots across both compiler profiles and all three modes: 48 configurations. |
| Actual transformations | Separate witnesses require peephole and squeeze effects. Removing either pass must fail its witness even when the circuit's semantics remain correct. |
| Compatibility and coverage declarations | Every registered pass needs a coverage decision. Noise/nonunitary removal passes are explicitly excluded from original-distribution equivalence. The noiseless translator rejects unsupported physical operations and malformed syntax instead of silently changing the input. |

No existing test cases or semantic assertions were removed. The original Qiskit collection was compared one-for-one with both new profiles after normalizing profile IDs. Consolidation is limited to shared registry reading and reference calculation; focused regressions and other oracle suites retain their purposes.

## Tests of the assertions

Committed negative controls verify that the suite detects:

- Either intended pass missing from the production-default factory.
- A zeroed final measurement, detector, or observable row in a partial batch.
- Wrong correlations with unchanged individual-bit marginals.
- Biased probabilities among legal outcomes, independently of the exact-zero support check.
- An unaccounted registered pass.

Run these controls with:

```bash
uv run --no-sync pytest tests/python/test_compiler_conformance.py -k 'rejects or detects' -q
```

These are targeted controls, not a comprehensive mutation score. The boundary circuit validates output extraction; the non-Clifford corpus supplies active-state semantics checks.

## Native execution evidence

The adapter test checks argument forwarding; native path execution was checked separately. With a Debug extension, GDB breakpoints on `Executor::run_shot()` and the ordinary `BatchExecutor::run_batch(...)` observed the following for the actual mixed-Pauli pytest case at 8193 shots:

| Compiler profile | Single-shot | Packed 65 | Automatic |
|---|---|---|---|
| Production default | Scalar only | Packed only | Packed only |
| Unoptimized | Scalar only | Packed only | Packed only |

All six pytest cases passed while traced. The development testing guide includes the build command, pytest selector, and breakpoint instructions for repeating this audit. Existing native policy/execution tests were also run. This is local execution evidence, not a claim about every automatic-selection branch or platform; no production instrumentation or new public API was introduced.

## Validation

- Release Python suite: **1,648 passed, 7 skipped**, also passing with `CLIFFT_FORCE_ISA=scalar`.
- Changed Python suites with a Debug extension: **385 passed**.
- Existing native packed/selection/thread-layout tests: **20 passed**.
- All **123 original Qiskit collection cases** verified present under **both** profiles.
- Repository-required pre-commit, including mypy: passed.
- GPU execution remains unvalidated by this PR; GPU-dependent tests are among the skips. No new C++ coverage build was performed.

Commands:

```bash
uv run --no-sync pytest tests/python/ -q -n 4
CLIFFT_FORCE_ISA=scalar uv run --no-sync pytest tests/python/ -q -n 4
uv run --no-sync pytest tests/python/test_compiler_conformance.py tests/python/test_qiskit_aer.py tests/python/test_pass_docs.py -q
cmake -S . -B build/native-tests -G Ninja -DCMAKE_BUILD_TYPE=Debug
cmake --build build/native-tests --parallel 8
ctest --test-dir build/native-tests --output-on-failure -R 'Packed|Threaded fixed-row|Sampling thread layouts' --parallel 4
```

### Runtime budget

Three interleaved main/changed full-suite runs on the same VM, Python 3.14.2, GCC 13 Release extension, four pytest workers, locked dependencies. Production sources are identical between revisions; the same extension was used for both suites.

| Median measurement | Main baseline | This PR |
|---|---:|---:|
| Process wall time | 8.902 s | 9.314 s |
| Total child CPU time | 27.309 s | 28.002 s |
| Passing cases | 1,390 | 1,648 |

The local increase is **0.412 s wall time (4.6%)** and **0.693 s CPU time (2.5%)**. Existing CI runs the full Python suite six times per PR, plus two wheel-suite runs on pushes to main. Multiplying the local deltas by the six PR runs gives approximately 2.5 summed runner-seconds and 4.2 CPU-seconds; this is cost accounting, not a prediction for the different CI hosts/builds. No CI jobs or timing assertions were added.

## Scope after this increment

#475 now defines a finite CPU foundation milestone: bounded noisy/mid-circuit joint coverage, actual opt-in scheduling adoption, explicit compatibility/negative controls, and a measured PR corpus plus scheduled CPU campaign. The noisy reference is the next priority. Sampling interfaces and GPU execution consume that foundation within their own feature series; hardware availability and wholesale test migration do not block this PR.

This corpus does not yet supply shared noisy/mid-circuit, postselection, retained-sampler, or GPU conformance. Existing focused tests remain in place. Review expectation provenance, strict translation, bit ordering, actual-path evidence, statistical sensitivity, and every proposed exclusion alongside the green test results.

### DIFF
--- a/docs/development/testing.md
+++ b/docs/development/testing.md
@@ -53,6 +53,66 @@ All procedural generators are centralized in [`utils_fuzzing.py`](https://github
 
 ## External Cross-Validation Oracles
 
+### Shared compiler conformance
+
+`tests/python/utils_conformance.py` defines the compiler profiles used by the
+Qiskit statevector suite and the small end-to-end corpus in
+`test_compiler_conformance.py`. Each profile constructs a fresh pass manager;
+the default profile calls the production factory rather than copying its pass
+list. The unoptimized profile disables HIR passes, not planning or executable
+preparation.
+
+The Qiskit suite keeps its existing circuits, seeds, and amplitude assertions
+and runs them under both profiles. The small corpus also compares every
+terminal measurement record's probability with Qiskit, then samples through
+single-shot, packed-capacity-65, and automatic CPU execution. Sampling checks
+complete record distributions and checks detector and observable bits against
+the returned records. References are computed independently of Clifft and
+cached by source within each worker. Compiled sampling programs are reused
+across execution modes.
+
+Two designated circuits require the default pipeline to reduce peak active
+width. One needs peephole cancellation; the other needs squeezing. Disabling
+either pass must fail its witness even when all output probabilities remain
+correct. The mirror case has a known all-zero outcome, but its elimination
+does not count as evidence that later execution paths ran.
+
+The pass-inventory check requires every registered pass to have a witness or
+an explicit exclusion. The current exclusions remove noise or nonunitary
+operations, changing the expected behavior. Add an explicit compiler profile
+and an exercise witness when covering an opt-in pass. New cases then run
+across compatible shared profiles. A new semantic feature still needs cases
+that actually exercise it.
+
+The noiseless oracle accepts only its supported unitary syntax, comments, and
+`TICK`. It rejects noise, measurements, resets, feedback, other annotations,
+and unsupported syntax instead of silently changing the circuit. This corpus
+does not yet provide a noisy, mid-circuit joint-distribution reference or GPU
+execution coverage. GPU adapters must declare their supported operations and
+precision; absence of a required GPU should fail a hardware CI job rather
+than skip all execution tests.
+
+#### Runtime and review
+
+Keep the PR corpus bounded. Measure before/after timings with the same build,
+worker count, and host conditions, including cost across the CI matrix.
+`pytest --durations=20` helps locate expensive tests; timing assertions do not
+belong in semantic correctness tests. Larger random sweeps can run separately.
+Do not reduce statistical sensitivity merely to meet a time budget.
+
+When consolidating tests, map each old case, seed, oracle, assertion, and
+execution mode to its replacement before removing it. Keep distinct bug
+regressions and structural or resource-lifetime tests. Some overlap is useful;
+coverage of the same lines alone does not establish redundancy.
+
+Review the source of expected results, explicit configuration selection, and
+exercise witnesses alongside coverage reports. Useful negative controls
+include removing a pass from the default factory, corrupting returned records,
+and adding an unaccounted pass to the inventory. Such controls test whether
+assertions can detect defects; line or branch coverage only shows execution.
+
+### Existing reference suites
+
 End-to-end Python tests compare Clifft against independent references whenever
 practical. These tests validate the full symbolic sampling pipeline rather
 than isolated implementation details.
--- a/docs/development/testing.md
+++ b/docs/development/testing.md
@@ -65,11 +65,19 @@ preparation.
 The Qiskit suite keeps its existing circuits, seeds, and amplitude assertions
 and runs them under both profiles. The small corpus also compares every
 terminal measurement record's probability with Qiskit, then samples through
-single-shot, packed-capacity-65, and automatic CPU execution. Sampling checks
+single-shot, packed-capacity-65, and automatic CPU execution with one worker
+so host-dependent thread counts do not change the tested configuration. Sampling checks
 complete record distributions and checks detector and observable bits against
 the returned records. References are computed independently of Clifft and
 cached by source within each worker. Compiled sampling programs are reused
 across execution modes.
+
+A separate analytic case checks every output row exactly at 63, 64, 65, 66,
+129, 130, 131, and 2049 shots. Its 65 measurement columns contain nonzero bits
+on both sides of a word boundary, and its detector and observable rows differ.
+These checks detect unwritten final rows and incorrect output mappings that
+could be too small to affect a statistical histogram. They cover output
+boundaries; the non-Clifford corpus provides the active-state checks.
 
 Two designated circuits require the default pipeline to reduce peak active
 width. One needs peephole cancellation; the other needs squeezing. Disabling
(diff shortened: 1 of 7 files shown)


docs/development/testing.md
```diff
@@ -53,6 +53,115 @@ All procedural generators are centralized in [`utils_fuzzing.py`](https://github
 
 ## External Cross-Validation Oracles
 
+### Shared compiler conformance
```
bachase: This new section is way too much detail. Please pare it down significantly, but do include a (concise!) section on how to properly extend and add new coverage/circuits.

docs/development/testing.md
```diff
@@ -53,6 +53,115 @@ All procedural generators are centralized in [`utils_fuzzing.py`](https://github
...
+does not yet provide a noisy, mid-circuit joint-distribution reference or GPU
+execution coverage. GPU adapters must declare their supported operations and
+precision; absence of a required GPU should fail a hardware CI job rather
+than skip all execution tests.
+
+#### Runtime and review
+
+The adapter's configuration-forwarding test alone does not prove which native
+executor ran. The existing policy tests in `tests/test_batch_executor.cc`
+check automatic selection and explicit capacities. For a direct execution
```
bachase: I don't like this runtime/debugger approach to see if something ran. Do we have some other simpler approach that can be automated/in scope for a CI run? Is this test really that important? 

We could add some optional logger and inspect that, but that might be overkill?

docs/development/testing.md
```diff
@@ -53,6 +53,115 @@ All procedural generators are centralized in [`utils_fuzzing.py`](https://github
...
+`clifft::sampling::BatchExecutor::run_batch(clifft::SeedRoot const&, unsigned int, unsigned int)`.
+Run each compiler profile with `single-shot`, `packed-65`, and `automatic`.
+The single-shot configuration must enter the scalar executor; capacity 65
+must enter the packed executor. The current automatic policy chooses packed
+execution for this case at 8193 shots. Record what actually ran when that
+policy changes. This audit supplies execution evidence; the independent
+oracle assertions supply semantic evidence. It does not validate other
+platforms, GPU execution, or every automatic-selection branch.
+
+Keep the PR corpus bounded. Measure before/after timings with the same build,
```
bachase: Yeah this and the following sections are too much instructions to implementors and should not be in the package documentation. Even many of these are transient "consolidating tests, mapping etc.) again not appropriate for documentation.

tests/python/test_compiler_conformance.py
```diff
@@ -0,0 +1,306 @@
+"""Small independent-oracle corpus shared across compiler and sampling modes."""
```
bachase: IS this mostly exercising the intersection of compiler and sampling modes with the various passes? Please share exactly these combinations come together in this test file.

tests/python/utils_conformance.py
```diff
@@ -0,0 +1,79 @@
...
+
+@lru_cache(maxsize=256)
+def unitary_reference(source: str) -> npt.NDArray[np.complex128]:
+    """Calculate once per source per worker, independently of compiler profiles."""
+    state = qiskit_statevector(stim_to_qiskit_noiseless(source))
+    state.setflags(write=False)
+    return state
+
+
+def assert_joint_distribution(measurements: npt.NDArray[np.uint8], expected: npt.ArrayLike) -> None:
```
bachase, line 55: Do we have any tests that validate this joint distribution approach is correct?

docs/development/testing.md
```diff
@@ -53,6 +53,41 @@ All procedural generators are centralized in [`utils_fuzzing.py`](https://github
...
+measurements. Other feature suites retain their own configurations; noisy,
+mid-circuit, postselection, and GPU coverage are not supplied by this matrix.
+Existing C++ tests in `tests/test_batch_executor.cc` check execution policy
+and directly exercise the packed executor.
+
+```bash
+uv run --no-sync pytest tests/python/test_compiler_conformance.py tests/python/test_qiskit_aer.py -q
+```
+
+#### Extending the corpus
```
bachase: This is still too specific and doesn't fit the style of this page. -- lets drop this section.

docs/development/testing.md
```diff
@@ -53,6 +53,41 @@ All procedural generators are centralized in [`utils_fuzzing.py`](https://github
 
 ## External Cross-Validation Oracles
 
+### Shared compiler conformance
+
+`test_qiskit_aer.py` runs its statevector corpus with HIR optimization disabled
```
bachase: This also still reads a bit too lower level detail. Lets keep in the style of this page where its purpose is for human readers to understand each style of test, but then rely on the code itself for them to see details.

PR #481, opened 2026-09-04
docs: reproduce logical Shor noise sweep on neutral atoms

bachase:
## Summary

- add an advanced tutorial that reconstructs the public Figure 9 logical Shor noise sweep across four neutral-atom schedules
- check in four nominal alpha=1 Clifft circuits with a concise Markdown provenance note, then rescale selected noise terms in memory without Cirq or a separate reproduction repository
- compare stabilizer-compatible rate rebalancing with exact conditional no-jump back-action
- add a 50,000-trajectory-per-model TVD sweep and a 10,000-trajectory-per-model exact-minus-rebalanced acceptance study
- include a controlled CZ rate-asymmetry stress test and keep lower sampling budgets as the quick tutorial defaults
- link the Logical Shor Noise Sweep tutorial from the leakage guide, delayed-loss tutorial, and documentation navigation

## Validation

- 1390 Python tests passed, 6 skipped, 1 xfailed
- high-statistics self-contained figure generation completed from the checked-in circuits
- tutorial matched and exact point commands passed
- strict MkDocs build passed
- full repository pre-commit suite passed


docs/guide/circuits/neutral_atom/manifest.json
bachase, line 1: Where did this format come form?

bachase: Yeah we don't need unit tests or this copmlexity. We can keep this to just have details of where the circuits originally came from, but don't need to go overborad.

bachase: Good catch. The JSON schema was a one-off format I introduced, not an existing project convention. I removed it and the dedicated unit-test module in f4589f0c. The circuits now have a short README.txt that records only the public paper and artifact, archive checksum, Apache-2.0 license, alpha setting, and motion translation.

natestemen: Out of curiosity, what is your workflow here?

bachase: Its inconsistent, but almost all code I generate is via an LLM these days. I then have another LLM do an initial review. Sometimes I have that post its feedback to the PR for the implementing LLM to review, but often I copy and paste it (with some edits) directly between agent conversaionts. Once the LLMs believe it is "ready", I do a manual review, usually in Github. I'll leave comments (like the first one here) and then tell the implementing LLM to review and respond to my feedback. Sometimes I'll give it some direct feedback there as well to be sure it goes in the right direction.  After it fixes, it often responds back to the comments as it did here.

Currently, these all use my github account to interact, hence it looks like I'm talking to myself. I've been meaning to setup a separate account for the LLMs to help distinguish the two. Perhaps this is a good time as any!

I'm also not sure this workflow is ideal if/when we get more contributors. We've been more focused on features and performance for now, but at some point that will shift and I'll want to revisit these workflows at that time.

natestemen: Thanks for the details! Seems like harder with contributors around, but until then 🤷🏻. Seems like it's working well.

When having LLMs review each other, do you just start a new session from another terminal window or are there better ways of doing this?

bachase: I usually have codex review claude and vice versa (each has its own directory). Occasionally create a new session in another worktree with the same agent. 

The workflow has evolved as the agents capabilities do too, so will continue to see what is the best approach.

Issue #487, opened 2026-09-14
Early-aborted shots still pay for whole-program noise events: d=5 end-to-end cultivation 0.25–0.7× slower than 0.1.0 since 0.8.0


bachase:
## Summary

On the clifft-paper end-to-end cultivation circuits (inject+cultivate, then escape into a d=15 surface code; 463 qubits; cultivation-stage detectors postselected), every release since 0.8.0 is slower per attempted shot than 0.1.0, and the cost **rises** with the physical noise rate even though more shots abort inside cultivation. Removing noise from the escape stage only (never reached by 88 to 100% of shots) removes most of the gap, so it looks like every pre-sampled noise event of the whole program is processed (~0.3 µs each) regardless of where the shot aborts. The 42-qubit inject+cultivate-only benchmark is too small to show this and is 2.4× faster on 0.10.0.

## Measurements

EPYC 9554P, one pinned core, `sample_survivors(keep_records=True)`, 20k shots, median of 3, µs per attempted shot. 0.1.0 is a source build with `CLIFFT_MAX_QUBITS=512` (required for these circuits); 0.8.0 to 0.10.0 are PyPI wheels. Packed = `batch_size=2048`, otherwise `batch_size=1`.

| d=5 end-to-end circuit | discard | 0.1.0 | 0.8.0 | 0.9.0 | 0.10.0 | 0.10.0 packed |
|---|---|---|---|---|---|---|
| p=5e-4 | 0.65 | 55.7 | 56.6 | 53.2 | 48.1 | 60.4 |
| p=1e-3 | 0.88 | 32.3 | 52.5 | 51.1 | 46.4 | 48.3 |
| p=2e-3 | 0.986 | 17.0 | 69.2 | 69.3 | 67.9 | 52.0 |
| p=5e-3 | 1.00 | 18.6 | 149 | 151 | 170 | 84.7 |
| p=2e-3, escape-stage noise removed | 0.986 | 16.7 | | | 28.9 | |
| p=5e-3, escape-stage noise removed | 1.00 | 16.9 | | | 44.9 | |

The d=3 end-to-end circuit shows the same shape (p=5e-3: 33 µs on 0.1.0, 78 µs on 0.10.0). For the paper's actual d=5 runs (p=1e-3 and 2e-3) this is 0.70× and 0.25× of 0.1.0 throughput.

## Reproduce

[...]

PR #491, opened 2026-09-15
fix(sampling): defer scalar noise until its first use

bachase:
Early-rejected scalar shots currently sample noise for the whole circuit before executing any postselection checks. Prepare each noise site's first-use deadline and retain one upcoming event, so dispatch samples noise only when needed and rejection stops the remaining work.

Replaces #489, which was closed during the stack update when #488 merged. GitHub rebased the branch onto main but refuses to reopen the old PR after that force-push. The production code is unchanged, verified with git range-diff. The early-rejection canary from #488 is now on main, so this PR directly measures the fix against that benchmark.

The [hosted canary on the original PR](https://github.com/unitaryfoundation/clifft/pull/489#issuecomment-5671213596) improved from 2.91 ms to 952 us per 1,000 attempted shots (67.2% less runtime); all nine other workloads showed no material change.

The implementation reuses the ordinary dispatch loop, with no new runtime mode, opcode, experimental selector, or hot allocation. Non-postselected plans remain eager; packed, fixed-fault, supplied-noise, and instrument-continuation sampling retain their noise policies. Addresses the scalar path in #487.

Validation on the original PR: all 10 native canary workloads; 933 C++ tests; 208 Python tests passed with one platform skip, including Stim and Qiskit Aer references and continuation coverage; all pre-commit hooks and CI passed. The replacement PR also passed its canary, with 64.1% less runtime on early rejection and no material changes on the other nine workloads ([report](https://github.com/unitaryfoundation/clifft/pull/491#issuecomment-5672629060)).


docs/guide/cpu-execution.md
```diff
@@ -138,6 +138,12 @@ capacities can produce different rows. Every supported strategy remains
 statistically equivalent. Keep the complete execution configuration fixed when
 exact seeded replay is required.
 
+Scalar postselected sampling draws quantum noise as its compiled dependencies
```
bachase: We don't need to worry about noise seeds being consistent across releases. 

I'd lean towards dropping this addition as a result.

src/clifft/sampling/executable_plan.h
```diff
@@ -391,6 +391,9 @@ class ExecutablePlan {
     std::vector<PreparedNoiseOutcome> noise_outcomes_;
     std::vector<PreparedNoiseSite> noise_sites_;
     std::vector<double> noise_hazards_;
+    // A site must fire before its first consuming action. Monotone deadlines
+    // let scalar sampling retain one upcoming event in circuit-site order.
+    std::vector<uint32_t> noise_action_deadlines_;
```
bachase, line 396: Layout wise is this a good spot for the vector?

PR #498, opened 2026-09-21
feat(hip): execute wide active states with cooperative tiers

josemonsalve2:
## Summary

The HIP backend rejected every plan with a peak active width above 4. That ceiling belongs to the thread-per-shot kernel, not to the packed action stream: the stream carries masks, pairings and phases that are identical whatever tier runs them. This moves the width limit to where it belongs, alongside the launch policy, and adds two cooperative tiers that give a whole workgroup to one shot.

## Execution tiers

Selection is automatic, from the plan's peak active width `k`, the coefficient element size, and the device's per-workgroup shared-memory budget. Per-shot coefficient storage is `3 * 2^k` elements, so `24 * 2^k` bytes in FP64 and `12 * 2^k` in FP32.

| Tier | Execution | State lives in | Typical reach |
|---|---|---|---|
| `thread-per-shot` | one shot per thread | thread-private | `k <= 4` |
| `cooperative-lds` | one shot per workgroup | on-chip shared memory | `k <= 11` on a 64 KB device, `k <= 12` on 160 KB (one more width in FP32) |
| `cooperative-global` | one shot per workgroup | a per-shot global slab | bounded only by device memory |

On the cooperative tiers the lanes of a workgroup split every dense pass: rotations partition the XOR pairs, promotion partitions the copy, and measurement reduces the branch probability through a workgroup tree reduction. `thread-per-shot` is unchanged and remains the small-width reference implementation. No new public API.

## Correctness notes

Measurement collapse routes the surviving half through the existing scratch buffers instead of compacting in place. Lanes read and write the same array with no ordering between them, so an in-place compaction lets one lane overwrite a coefficient another lane has not yet read. This was caught before the tier was measured, not after.

Rows are not promised to match bit-for-bit **between** tiers: a workgroup sums measurement probabilities in a different order than a single thread, so a near-tie can resolve the other way. Within one tier a fixed seed still reproduces exactly. This matches the existing contract between the CPU's scalar and packed modes and is now documented alongside the tier table.

Batching behaviour is unchanged. A batch larger than the retained workspace is split into several synchronous launches, and because the RNG keys off the global shot index, splitting never changes seeded rows. `cooperative-global` additionally clamps the workspace to what the device can allocate, so a wide plan runs as more batches rather than failing to allocate.

## Tests

Every existing device test predated the ceiling, so nothing in the suite exercised a plan wider than `thread-per-shot`. This adds four cases built from independent T gates, which keep one non-Clifford coordinate live per qubit and so reach a chosen peak active width with no external input:

- expectation values against the CPU executable on `cooperative-lds` (`k = 8`), FP64 and FP32
- expectation values against the CPU executable when the state exceeds any workgroup budget (`k = 14`), FP64 and FP32
- detector rates against the CPU executable at a cooperative width, compared within six standard errors
- repeatability of a fixed seed within each precision at a cooperative width

Each asserts the width it actually reached. An optimizer that later squeezes the active state would otherwise leave these silently running on the narrow tier while still passing.

`select_execution_tier` also gets direct coverage of its boundaries in `tests/test_hip_executable_plan.cc`.

## Test plan

Run on an AMD Instinct MI350X (gfx950), ROCm 7.2.3, built with `-DCLIFFT_ENABLE_HIP=ON -DCMAKE_HIP_ARCHITECTURES=gfx950`.

- [x] C++ tests pass — **955 / 955**, 0 failed
- [x] HIP subset — **28 / 28**, **0 skipped** (`catch_discover_tests` uses `SKIP_RETURN_CODE 4`, so the skip count is the real check); direct run of `clifft_hip_tests` reports **59,130 assertions in 18 test cases**
- [x] Python tests pass — **1,413 passed, 1 skipped, 3 failed**. All three failures reproduce identically on pristine `main` at 8a00dbe3 on the same machine: two are a `record_probabilities` column-count mismatch in `tests/python/utils_hip.py` on the CPU path this PR does not touch, and one is an OpenMP/Aer coexistence timeout (`main` fails three of those cases, this branch one). No regressions.
- [ ] Doc examples — not run
- [x] Formatting — `git-clang-format` against `main` with the pinned clang-format 18.1.8, so only changed lines are touched. Running clang-format whole-file reformats 121 lines of unmodified `sampler.hip`, which I left alone.

## Contributor agreement

- [x] I confirm this contribution is my original work (or I have the right to submit it) and I license it under this project's Apache-2.0 license.
- [x] This contribution was AI-assisted; I have reviewed the generated code and each commit carries an `Assisted-by:` trailer.


src/clifft/sampling/hip/sampler.hip
```diff
@@ -503,9 +503,444 @@ __global__ void interpret_shots(ProgramView program, SeedRoot seed_root, uint64_
...
+            case ActionTag::ApplyReadoutNoise: {
+                if constexpr (Replay) {
+                    reachable = false;
+                } else {
+                    const bool source = evaluate(program, symbols, action.expression);
+                    const double probability = source ? action.value1 : action.value0;
+                    const bool flip = probability >= 1.0 ||
+                                      (probability > 0.0 && rng.next_double() < probability);
+                    symbols[action.index0] = static_cast<uint8_t>(flip);
+                    records[action.index1] ^= static_cast<uint8_t>(flip);
```
bachase-agent: [P1] Apply each readout flip only once per shot

Every lane executes this non-atomic XOR against the same per-shot record byte. A wavefront that reads after another has written can undo the flip; the barrier after the action does not serialize the updates. For example, the following reaches active width 8, where the cooperative block has 128 lanes, and must return measurement one and detector one for every shot:

```text
H 0 1 2 3 4 5 6 7
T 0 1 2 3 4 5 6 7
EXP_VAL X0
M 8
READOUT_NOISE(1) rec[-1]
DETECTOR rec[-1]
```

The production CPU probe returns 16/16 ones for both outputs. In the cooperative kernel, multiple wavefronts can instead apply the flip repeatedly. Keep RNG advancement consistent across lanes, but let only lane 0 publish the flip symbol and update the record, then synchronize. CUDA already uses this [single-writer pattern](https://github.com/unitaryfoundation/clifft/blob/57e15957cacc997c876b376b7cb5e88b35c71ef3/src/clifft/sampling/cuda/sampler.cu#L563). Add cooperative-width readout-noise regressions, including detector and survivor behavior.

src/clifft/sampling/hip/sampler.hip
```diff
@@ -503,9 +503,444 @@ __global__ void interpret_shots(ProgramView program, SeedRoot seed_root, uint64_
...
+    // is consumed identically in all lanes.
+    reduce_two_sums(reduction, probability_zero, probability_one);
+
+    const double total = probability_zero + probability_one;
+    const double epsilon = clifft::kMeasurementDustEpsilon * total;
+    bool branch;
+    if constexpr (Replay) {
+        const bool correction = evaluate(program, symbols, action.expression);
+        branch = (forced_records[action.index1] != 0) != correction;
+        if ((probability_one <= epsilon && branch) || (probability_zero <= epsilon && !branch)) {
```
bachase-agent, line 744: [P1] Finish replay correction reads before publishing the branch symbol

The lowered outcome expression includes this action's own branch symbol. A faster wavefront can publish one at line 726 before a slower wavefront evaluates the correction here. The slower lanes then select the opposite branch, potentially corrupting collapse results or disagreeing about reachability and barriers. Restricting writes to lane 0 alone is insufficient: every lane must finish reading the old symbol value before publication.

A production-plan probe at width 8 confirms the correction has constant zero and one term, the branch symbol itself. Using the eight-qubit H/T prefix in the other comment followed by `T 0`, `T 0`, `T 0`, `H 0`, `M 0` gives a reachable forced-one CPU replay with log probability zero; reading the already-published symbol instead chooses the unreachable raw-zero branch.

Add a workgroup barrier after the branch calculation and before publication, and use one writer for outputs. Apply the same protection to `MeasureDormantRandom` at lines 852-861. CUDA already protects [active replay](https://github.com/unitaryfoundation/clifft/blob/57e15957cacc997c876b376b7cb5e88b35c71ef3/src/clifft/sampling/cuda/sampler.cu#L334) and its dormant replay path. Add cooperative forced-replay cases that check the record, likelihood, and post-collapse state.

bachase-agent (changes requested):
Moving width eligibility out of lowering and selecting a tier during sampler construction is consistent with CUDA. I am requesting changes for the two cooperative-kernel races called out inline: repeated readout flips against the same record byte, and replay correction reads racing with publication of the branch symbol.

The three execution categories are equivalent to CUDA's, so please align their names:

| HIP name in this PR | Consistent C++ name | Text/API spelling |
| --- | --- | --- |
| `ThreadPerShot` | `ThreadPerShot` | `thread_per_shot` |
| `CooperativeLds` | `BlockShared` | `block_shared` |
| `CooperativeGlobal` | `BlockGlobal` | `block_global` |

LDS can remain in the implementation explanation. Device-specific capacity thresholds and lane counts can differ while the category names agree. The CUDA documentation headings would also work here: Tier, Threads per shot, Coefficient storage, Auto selection.

A few related consistency points:

- HIP currently ignores the caller's `block_size` on cooperative tiers and derives 64, 128, or 256 lanes from peak width. CUDA honors a power-of-two block size. Either support that control consistently or explicitly document HIP's automatic sizing; the current HIP guide says the argument controls launch geometry without this qualification.
- Consider exposing the resolved tier on `Sampler` and a device-aware `selected_tier` query, as CUDA does, so tests can assert which kernel they exercised. Forced tier selection can remain a separate scope decision.
- HIP allocates one global slab per retained batch shot, whereas CUDA reuses a bounded pool across shots and separates concurrency from output batch capacity. That scheduling difference does not require different tier names. Independent concurrency control is a useful follow-up, not a prerequisite for this naming cleanup.
- The capability table should state `k <= 30`, subject to available memory. The thread-per-shot coefficient storage is a global slab owned by the shot; describing it that way would make the storage comparison clearer.

The new wide-state tests cover expectation values, detector statistics, and repeatability, but do not exercise cooperative readout noise or forced replay. Please add targeted regressions for both fixes, covering the shared and global tiers, deterministic and stochastic replay branches, and the state after collapse. A broader shared conformance suite can remain follow-up work for this experimental backend.

Validation: a local Release CPU build with `CLIFFT_FORCE_ASSERTS=ON` passed all 10 HIP lowering and tier-selection tests. Production-plan/CPU probes confirmed width-8 examples, deterministic readout results, and replay expressions containing their own branch symbols. No AMD GPU was available locally, so the kernel findings are based on code and permitted interleavings, not a claimed hardware reproduction. Existing `hip-offline-build` CI passes; the current lint failure is clang-format changing `tests/test_hip_sampler.cc` and also needs to be cleared.

Reviewed with Codex assistance.

src/clifft/sampling/hip/sampler.hip
josemonsalve2: Fixed as suggested: every lane still draws the flip so the RNG stream stays identical, a
single lane performs both stores, and the barrier at the end of the action publishes them.

Two things I did differently.

**I applied the discipline to every write site, not just this one.** Rather than patch the
two reported actions, all shared per-shot writes in the block kernel now go through a
`CooperativeLane` with `is_writer()` and `sync()`, mirroring CUDA's `Lane`. That covers
`RecordClassical`, `DefineSymbol`, `WriteDetector`, `WriteObservable`,
`WriteExpectationValue` and noise presampling, which take the writer flag for the same
reason CUDA's `sample_noise` does. Those were benign — lanes agreed on the value and none
was a read-modify-write — but the failure here was a class of bug, and leaving the same
shape in place invites it back.

**The regression test does not currently fail without the fix, and I want to be explicit
about that rather than let the green tick imply otherwise.** I reverted the single-writer
guard, rebuilt, and ran the new test: it passed. The reason appears to be that every
action ends in a barrier, so wavefronts enter the readout action tightly coupled, and the
window between the shared read and the store is a few instructions — on gfx950 all four
wavefronts of a 256-lane block read the old byte before any writes, and the XOR lands
once. So the defect is real (there is no ordering guarantee, and the write is a
non-atomic RMW) but latent under this hardware's scheduling, which matches your note that
the finding came from permitted interleavings rather than a hardware reproduction.

The test is therefore a behavioural lock, not a reproduction: `READOUT_NOISE(1)` is
certain, so every shot must report one, and the test asserts that at a `block_shared`
width and a `block_global` width, at both precisions, asserting the resolved tier so a
future capacity change cannot quietly move it onto the narrow kernel. It would catch a
different block size, a different wave size, or a scheduler that does skew wavefronts.

src/clifft/sampling/hip/sampler.hip
josemonsalve2: Fixed. The barrier goes immediately after the branch calculation and before publication,
exactly as described, and the same protection is applied to `MeasureDormantRandom`.

One thing beyond what you flagged: the record written alongside the symbol is
`evaluate(program, symbols, action.expression)`, which reads the symbol that same statement
publishes. With every lane writing, a lane could evaluate that expression against either
the old or the new symbol depending on timing, so the two stores now come from one lane to
stay self-consistent.

**Two notes on the regression test.**

It does not fail without the fix — I checked by reverting rather than assuming. Same cause
as the readout comment: the preceding `reduce_two_sums` ends in a barrier, so every lane
evaluates the correction within a few instructions of the others and no lane lags into the
published value. Real defect, latent scheduling.

More interesting: while writing it I found that the **unreachable branch cannot be asserted
at FP32**, and it is not a tier problem. Reachability is decided against
`kMeasurementDustEpsilon = 1e-18` applied as a *relative* threshold. FP32 rounding leaves
dust orders of magnitude above that, so a mathematically impossible branch reads as
reachable. I confirmed this is precision-specific rather than kernel-specific by shrinking
the same circuit to width 3 so it runs on `thread_per_shot`, where FP32 disagrees
identically. Filed as #503; the test asserts the unreachable branch at FP64 only, with a
comment explaining why (no issue number in the comment, per AGENTS.md).

The test otherwise checks what you asked for: the record, the log probability against the
CPU executor, and the post-collapse expectation values, at a `block_shared` width and a
`block_global` width.

josemonsalve2:
Thanks — both races were real and I have fixed them, plus everything else raised. Details on the two kernel findings are inline; the summary below covers the rest, and flags one thing I deliberately did differently.

## The two races

Fixed as described, and I extended the discipline rather than patching two sites: every write to shared per-shot state in the block kernel now goes through a `CooperativeLane` with `is_writer()` and `sync()`, mirroring CUDA's `Lane`. Noise presampling takes the writer flag for the same reason CUDA's `sample_noise` does.

**Neither regression test fails without its fix.** I verified that by reverting each fix and rebuilding rather than assuming, and I would rather say so than let a green tick imply a reproduction. Every action ends in a barrier, so wavefronts enter each action tightly coupled and the read-to-publish window is a few instructions; on gfx950 the race does not manifest. The defects are still real — a non-atomic read-modify-write and an unsynchronised read of a location the action publishes — and this matches your note that the findings came from permitted interleavings rather than hardware. The tests are behavioural locks that would catch a different block size, wave size, or a scheduler that does skew.

## Naming and API

Tiers are now `ThreadPerShot` / `BlockShared` / `BlockGlobal`, spelling `thread_per_shot` / `block_shared` / `block_global`. `selected_tier(executable, precision)` reports the tier without uploading, `Sampler::execution_tier()` reports what a sampler resolved, and both route through one function with the Sampler so a query cannot disagree with what runs. The new tests assert the tier they exercise instead of inferring it from a width.

I did **not** add `ExecutionTier::Auto`. Forcing a tier is out of scope per your note, and an `Auto` enumerator with nothing to distinguish it from would be dead API. Easy to add with forced selection.

Docs now use the CUDA guide's headings (Tier, Threads per shot, Coefficient storage, Auto selection), state `k <= 30` subject to device memory, and describe thread-per-shot storage as the global slab it is.

## Block size — where I diverged, with a measurement

You offered two options: support the control consistently, or document the automatic sizing. **I did both, because honouring the request unconditionally turned out to cost 2.2x.**

On the block tiers the block size is also the reduction width, so it must be a power of two between a wavefront and the static reduction capacity (256); anything else is now rejected rather than silently replaced. But making the 256 default authoritative regressed narrow widths badly — at `k = 5` the state has 16 pairs of work, and 256 lanes means 240 idle plus an 8-level barrier reduction instead of 6:

| k | honouring a flat 256 | automatic sizing |
|---|---|---|
| 5 | 0.450x | 0.996x |
| 6 | 0.465x | 0.997x |
| 7 | 0.415x | 1.000x |
| 8 | 0.662x | 1.002x |
| >= 9 | 1.000x | 1.000x |

So `block_size` now defaults to `kAutoBlockSize` (0), which asks the backend to size the launch — 256 for thread-per-shot, derived from the width for the block tiers. Any explicit power of two in range is used exactly as given. This is the one place the HIP spelling differs from CUDA's literal `kDefaultBlockSize = 256`, and the divergence is documented in the guide with the reason.

That makes `block_size = 0` mean "choose for me" instead of being rejected, so the commit is marked `!` with a `BREAKING CHANGE:` footer, alongside the tier renames.

## Two upstream bugs found along the way

- **#503** — FP32 replay reports mathematically unreachable branches as reachable. `kMeasurementDustEpsilon = 1e-18` is applied as a *relative* threshold, which FP32 cannot resolve. Not tier-specific: the same circuit at width 3 on `thread_per_shot` disagrees identically. The new replay test asserts the unreachable branch at FP64 only for this reason.
- **#504** — `test_hip_python_forced_replay_probes_each_branch` fails on a clean `main`: the helper builds its record table from `num_records` while `record_probabilities` validates against `num_measurements`. Verified on both `8a00dbe3` and `57e15957`, where the three files involved are byte-identical.

Neither is caused by this branch. I have not proposed patches since both are contract decisions rather than obvious fixes.

## Validation

Rebased onto `57e15957`. On an MI350X (gfx950), ROCm 7.2.3:

- **C++: 967 / 967 pass.** HIP subset **31 / 31 with 0 skips** (the `SKIP_RETURN_CODE 4` masking makes the skip count the real check); device suite **59,381 assertions in 21 test cases**.
- **Python: 1,674 passed, 32 skipped, 4 failed** — two are #504, two are the qiskit-aer/OpenMP start-up timeout. `main` fails the same tests on this machine, three of the OpenMP cases to this branch's two.
- **Lint: `pre-commit run --all-files` passes**, using the pinned hook rather than a local clang-format. Worth noting for anyone else hitting this: a `pip install clang-format==18.1.8` build reformats 121 lines of unmodified `sampler.hip`, so it is not interchangeable with the mirror the hook pins.
- **Performance: median 1.000x** against the pre-review branch over 87 circuits spanning k = 4..21, measured with a 3-second target per circuit so the comparison sits at sustained clocks rather than in boost. The added barriers and single-writer guards cost nothing measurable.

Assisted-by: Claude Code (Claude Opus 5) <noreply@anthropic.com>

docs/development/hip-backend.md
```diff
@@ -22,16 +22,75 @@ launch controls stay outside the stable API.
...
 | Asynchronous or multi-GPU execution | Not supported |
 
-The current tier uses one GPU thread per shot and targets circuits with small
-active states. Unsupported programs are rejected during lowering; there is no
-automatic CPU fallback.
+Unsupported programs are rejected during lowering; there is no automatic CPU fallback.
+
+### Execution tiers
+
+The backend runs one of three tiers. These are the same categories the CUDA backend uses,
```
bachase: I'd drop this reference to CUDA, as implementations may drift over times. I'd just have these docs standalone as descriptions of the HIP backend.

src/clifft/sampling/hip/sampler.hip
```diff
@@ -834,18 +1410,59 @@ class Sampler::Impl {
...
+                               shot_offset, shots, coefficient_storage, symbols.data(),
+                               records.data(), forced_records.data(), detectors.data(),
+                               observables.data(), exp_vals.data(), log_probabilities.data(),
+                               reachable.data(), survived.data());
+        } else {
+            // One block per shot. The block is also the reduction width, so the caller's
+            // request is honoured within the domain the reduction can support rather than
+            // being silently overridden; validate_launch_block_size rejects the rest.
+            const uint32_t lanes =
+                cooperative_block_size(block_size, program.view.peak_active_width);
```
bachase-agent: [P2] Use cooperative launch geometry for replay

`run_replay()` still calls `launch(..., shots=1, block_size=1, ...)` at lines 1364/1366. The new `cooperative_block_size()` returns any explicit nonzero request unchanged, so this path launches both block tiers with one thread and bypasses the 64-lane minimum enforced for sampling.

A host runtime mock intercepting launches from the compiled backend confirms:

| Peak width | Resolved tier | Sampling threads per block | Replay threads per block |
| --- | --- | --- | --- |
| 8 | `BlockShared` | 128 | 1 |
| 14 | `BlockGlobal` | 256 | 1 |

This serializes all dense passes and reductions during wide replay. It also means the cooperative replay regression cannot exercise cross-lane publication hazards, regardless of GPU scheduling. These are host launch observations, not a hardware execution claim.

Choose automatic cooperative geometry for replay on the block tiers, retaining one thread for thread-per-shot if desired, and cover that launch choice. [CUDA already makes this distinction](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/src/clifft/sampling/cuda/sampler.cu#L1087-L1095).

tests/test_hip_sampler.cc
```diff
@@ -509,3 +583,211 @@ TEST_CASE("HIP sampler matches CPU survivor statistics with noise") {
...
+                    continue;
+                }
+                REQUIRE_THAT(actual.log_probability,
+                             Catch::Matchers::WithinAbs(expected.log_probability, 1e-12));
+                REQUIRE(actual.outputs.measurements ==
+                        std::vector<uint8_t>(forced.begin(), forced.end()));
+                // The post-collapse state is what the expectation value is read from, so
+                // agreeing here is the check that collapse used the same branch.
+                REQUIRE(actual.outputs.exp_vals.size() == cpu.exp_vals().size());
+                for (size_t index = 0; index < actual.outputs.exp_vals.size(); ++index) {
```
bachase-agent: [P2] Make the post-collapse assertions observe actual state

`parallel_t_replay()` ends with `M 0` and contains no `EXP_VAL` instruction. Production-plan/CPU probes at widths 8 and 14 both report `num_exp_vals == 0` and zero post-collapse expectation outputs. Consequently, this size comparison passes with two empty vectors and the loop executes zero iterations; corrupting the collapsed coefficient state would not be detected here.

Add downstream expectation values or later measurements that observe the collapsed state, and assert the expected output count is nonzero. Also include stochastic forced branches: the sole reachable branch in this fixture has log probability zero, so it does not validate the nontrivial probability accumulation path. Run these cases on both cooperative tiers and precisions, with appropriate tolerances.

src/clifft/sampling/hip/sampler.hip
```diff
@@ -769,6 +1305,45 @@ class Sampler::Impl {
...
+        size_t free_bytes = 0;
+        size_t total_bytes = 0;
+        if (hipMemGetInfo(&free_bytes, &total_bytes) != hipSuccess) {
+            free_bytes = 0;
+        }
+        // Headroom for the per-shot record/detector/observable arrays and driver overhead.
+        const uint64_t budget =
+            free_bytes != 0 ? static_cast<uint64_t>(static_cast<double>(free_bytes) * 0.6)
+                            : (uint64_t{1} << 30);
+        uint64_t affordable = budget / per_shot;
```
bachase-agent: [P2] Include all retained per-shot buffers in the memory budget

`per_shot` counts only coefficients, but construction also allocates `max_batch` rows of symbols, records, detectors, observables, FP64 expectation outputs, and survival flags. These can exceed the fixed 40% allowance, so the calculated affordable batch can fail allocation even when a smaller batch would fit.

I reproduced this using the compiled host backend with a HIP runtime mock enforcing a 64 MiB device budget. A valid width-14 plan, built from independent H/T gates followed by 50000 `EXP_VAL X0` instructions, fails default construction: the expectation allocation requests 38400000 bytes with only 25559064 bytes left. The same plan succeeds with `max_batch_shots=32`, retaining 29184017 bytes total. No GPU execution is involved in this allocation probe.

Compute the batch bound using all per-row device storage and the remaining fixed allocations, leaving a separate reserve for runtime overhead. Add an allocation-budget regression for an output-heavy plan; coefficient-only plans do not catch this failure.

bachase-agent (dismissed):
Reviewed `d3c4988`. The earlier readout and replay publication fixes are present, and the overall tiering approach is consistent with CUDA. I am requesting changes for the three implementation and validation findings inline. I did not identify another clear numerical error in the cooperative arithmetic.

**Changes requested**

1. Cooperative replay still launches with one thread because `run_replay()` passes a literal block size of one. This serializes wide replay and prevents the replay race regression from exercising cross-lane behavior.
2. The replay regression does not actually check post-collapse state: its circuit has no `EXP_VAL`, so the expectation comparison is empty. Its only reachable branch is deterministic, leaving nontrivial replay log probabilities untested.
3. The automatic batch clamp counts coefficient storage alone. The fixed 40% allowance can be exceeded by the other per-shot buffers, causing avoidable allocation failure even when a smaller retained batch fits.

**API and quality improvements, separate from those findings**

- Please consider adding a user-supplied tier override with `ExecutionTier::Auto` as the default, and `tier="auto"` in Python. [CUDA already defaults to Auto](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/src/clifft/sampling/cuda/sampler.h#L69-L72). Forward it through `SamplingOptions`, the constructor, convenience functions, and bindings; validate forced shared-memory fit including reduction scratch. The automatic width-four cutoff should not become a hard restriction on a forced thread-per-shot tier without checking the kernel's actual limits. Overrides would also let conformance tests exercise specific kernels independently of device capacity.
- Expose tier inspection through Python as well. This PR adds C++ `selected_tier()` and `Sampler::execution_tier()`, but [the HIP bindings](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/src/python/hip_bindings.cc#L63-L112) and facade still lack CUDA's `selected_tier()` and `Sampler.tier` equivalents.
- Propagate device-query failures instead of silently selecting a tier from an assumed zero LDS budget. On a host with no AMD GPU, HIP `selected_tier()` currently returns `block_global` for a wide plan. CUDA's `query_device()` checks and reports runtime errors. The same principle applies to the guessed memory budget after `hipMemGetInfo` fails.
- HIP's automatic block sizing and AMD-specific legal block sizes are reasonable differences from CUDA; exact matching is not required. Independent control of coefficient concurrency versus retained output batch size, as CUDA provides, can remain follow-up work.
- Update stale implementation comments: `sampler.hip:531-532` still says identical concurrent per-shot writes are benign, despite the new single-writer discipline, and `device_program.h:46-48` says block size is fixed despite supporting 64/128/256 lanes.
- Broaden cooperative conformance coverage to stochastic forced replay, non-diagonal collapse followed by observable state checks, survivor handling, and batching invariance. Assert the specific intended resolved tier for each precision instead of only `tier != ThreadPerShot`; that condition alone cannot guarantee coverage of both shared and global storage. CUDA's [per-tier replay tests](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/tests/test_cuda_sampler.cc#L650-L703) are a useful comparison.

**Validation**

A fresh Release build with `CLIFFT_FORCE_ASSERTS=ON` compiled the `gfx942` HIP backend, `clifft_tests`, and `clifft_hip_tests`. The focused HIP run had **12 passed, 19 skipped, 0 failed**; the skips require an AMD GPU, which was unavailable locally. CPU production-plan probes confirmed the empty post-collapse assertions. A separate host HIP runtime mock intercepted launches from the compiled backend and reproduced the allocation failure described inline. These probes do not execute GPU kernels. Current applicable CI checks pass.

Reviewed with Codex assistance.

bachase:
Also worth a rebase now that the two bugfixes landed?

bachase-agent:
A few supplementary testing and documentation points to accompany [the latest review](https://github.com/unitaryfoundation/clifft/pull/498#pullrequestreview-5295021778):

1. **Correct the remaining documentation.** The [later launch-controls description](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/docs/development/hip-backend.md#L201-L204) still says `block_size` must be between 1 and 1024, omitting automatic sizing and the cooperative restrictions. The kernel-development guide still describes the cooperative path as [future work](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/docs/development/hip-kernel-development.md#L118-L123) and tells contributors to update only [one interpreter switch](https://github.com/unitaryfoundation/clifft/blob/d3c4988edb1444ae7f3179b908987731ea02bb02/docs/development/hip-kernel-development.md#L94-L103). Update those passages to match the final implementation.

2. **Strengthen boundary and measurement coverage.** Find the shared/global transition with `selected_tier()` for each precision, then exercise widths immediately below and above it and assert the resolved tiers. Add a biased measurement outcome alongside the current 50/50 detector case so the test can distinguish the intended probabilities from an incorrectly uniform result.

3. **Build on the shared replay cases introduced by #511.** The reusable width-3 or width-4 circuit is suitable for a separate test-only PR against main. For this PR, exercise the same scenario at cooperative widths, checking active and dormant measurements, nontrivial branch probabilities, and post-collapse outputs, with explicit tier assertions. A validated scenario is an H/T prefix across the active qubits, followed by `CX 0 1`, `H 0`, `M 0`, `EXP_VAL X1*X2`, `H 40`, and `M 40`. CPU probes at widths 8 and 14 give four reachable branches with probabilities 3/8, 3/8, 1/8, 1/8 and expectation values 2/3 or 0 depending on the first outcome. These wider tests still depend on fixing replay's launch geometry; passing the small-width case does not validate cooperative execution.

4. **Track interpreter duplication as an optional follow-up.** Consider sharing action semantics between the thread and block kernels, as CUDA does, to reduce the risk of one implementation drifting. This refactor is not a merge requirement. For now, document that new actions must be implemented in both switches and keep conformance tests covering both paths.

Reviewed with Codex assistance.

josemonsalve2:
Rebased onto `9d49d27` (picking up #509, #511 and #512) and addressed all three findings plus the improvements. One of them corrected something I had told you with more confidence than it deserved, so that first.

## The replay launch geometry — and what it revealed about my own claim

You were right, and it is my regression. Before I made `block_size` honourable it was ignored on the block tiers entirely, so `run_replay`'s literal `1` only ever reached thread-per-shot. Once the caller's value was honoured, cooperative replay dropped to one lane. `run_replay` also bypasses `validate_launch_block_size`, so nothing rejected it. Replay now asks the backend to size the launch on the block tiers; at `k = 12` that resolves to 256 lanes rather than 1.

**I then re-ran the negative control, because your finding implied my replay regression had never exercised cross-lane behaviour — and it still does not fail.** Reverting the read-before-publish barrier with a full block of lanes: all 26 tests pass. The reason looks structural rather than lucky: `reduce_two_sums` ends in `__syncthreads()`, so every lane leaves the reduction together and evaluates the correction within a few instructions of lane 0's write. There is no window to lag into.

So the position is unchanged but now better founded: both defects are real — a non-atomic read-modify-write and an unsynchronised read of a location the action publishes — and neither reproduces on gfx950. The tests are behavioural locks, not reproductions, and I would rather keep saying so than let three green runs imply otherwise.

## The other two findings

**Batch clamp.** Every retained buffer scales with the batch, so counting only coefficients and leaving a fixed allowance for the rest refused batches that would have fit whenever the rest exceeded it. All per-shot buffers are counted now; the remaining headroom covers driver allocation and fragmentation only.

**Post-collapse state.** Correct — the old replay circuit had no `EXP_VAL`, so the expectation comparison was empty, and its only reachable branch was deterministic. Replaced per your point 3: #512's biased active-and-dormant scenario widened to straddle the tier boundary, with the dormant measurement on qubit 40 so it stays dormant however wide the prefix grows. Branch probabilities are 3/8, 3/8, 1/8, 1/8, and both expectation values are read after a collapse, so a wrong branch shows up in the state and not only in the record.

## Tier override and inspection

`ExecutionTier::Auto` added and forced values threaded through `SamplingOptions`, the `Sampler` constructor and `replay_shot`. A forced tier is validated against the device including reduction scratch. Per your note, the width-four cutoff is **not** a restriction on a forced thread-per-shot tier: its coefficients live in a global slab like `BlockGlobal`'s, so only `BlockShared` can fail, and only on capacity.

Python gains the same surface — `hip.Sampler(program, tier="block_global")`, `Sampler.tier`, and module-level `hip.selected_tier(program, precision)`.

**Device-query failures now propagate.** You are right that returning `block_global` from an assumed zero budget on a host with no GPU is worse than an error. Both the property query and `hipMemGetInfo` report failure instead.

## Coverage

Tests now **discover** the shared/global boundary with `selected_tier()` per precision and assert the resolved tier on each side, rather than asserting `!= ThreadPerShot`. I had written a hard-coded width while doing this and caught it against exactly your point — it is gone. Added: forced-tier acceptance and rejection at the boundary, the widened biased replay at both tiers, survivor sampling, and batching invariance (a 3,000-shot workspace against a 20,000-shot request must reproduce every row, since the RNG keys off the global shot index). The biased first measurement is 3:1, so a uniform result fails.

One assumption I had been making silently and now assert: `plan_from` in the C++ tests does not run the squeeze pass that `clifft.compile` does, so peak width equals the prefix there and is one lower through Python. The tests `REQUIRE` the width they got.

## Documentation

All four passages you cited are corrected: the launch-controls list, the "future work" description of the block path, the instruction to update one interpreter switch — now explicitly both, with the single-writer requirement — and the two stale code comments.

## Left as follow-up

- Independent concurrency vs retained batch control, per your note that it can remain follow-up.
- Sharing action semantics between the two interpreter switches. Not attempted here; the guide now states the obligation to implement in both, which is the documentation half of your point 4.
- Adding the widened cases to the shared `kGpuReplayCases` corpus. That table drives CUDA too, and I cannot run CUDA, so I did not want this PR to add test obligations for a backend I cannot verify. Happy to if you would rather have them there.

## Validation

MI350X (gfx950), ROCm 7.2.3, rebased onto `9d49d27`:

- **C++ 974 / 974**, HIP subset **37 / 37 with 0 skips**, device suite **64,524 assertions in 26 test cases**
- **Python 1,773 passed, 68 skipped, 0 failed** — #506 and #511 cleared the failures I was carrying
- **Pre-commit passes** on all hooks
- **Performance median 1.001x** over 87 circuits spanning k = 4..21, against the branch as you first reviewed it

Assisted-by: Claude Code (Claude Opus 5) <noreply@anthropic.com>

bachase (approved):

